PR #18, opened 2022-04-24
Add a lint-checker script

Harry-Hopkinson:
In subjunction with #17


Brendonovich:
This should probs be part of CI, having a separate workflow kinda defeats the point of linting since it's not stopping anything from happening if linting fails. Would also be handy to have linting as a pre-commit hook.

Harry-Hopkinson:
I will upload a PR addressing these changes now.

PR #30, opened 2022-04-27
added experimental submenu

MarquesCoding:
+ debug toggle for experimental view


package.json
```diff
@@ -50,5 +50,8 @@
     "npm": "pnpm",
     "yarn": "pnpm",
     "node": ">=14.0.0"
+  },
+  "dependencies": {
+    "zustand": "^3.7.2"
```
jamiepine, line 55: avoid adding dependencies to the root of the repo, add packages like so `pnpm interface add zustand`

PR #31, opened 2022-04-27
add: custom scrollbar

sr2echa:
## Changes
`.\packages\interface\src`

➖  **Removed :**
> 1. no-scrollbar
> 2. explorer-scroll

➕  **Added :**
> Universal custom scrollbar (so no need for explorer only scrollbar and a different scrollbar for overview,etc)

### Before:
image

### After:
image

#### After (scrollbar Onhover) [sorry that my cursor is not visible in the screenshot]:
image

### DIFF
--- a/packages/interface/src/style.scss
+++ b/packages/interface/src/style.scss
@@ -15,27 +15,10 @@ html { scroll-behavior: smooth; }
 ::-webkit-scrollbar-thumb {
 	background-color: rgba(127, 127, 127, 0.6);
 	background-clip: padding-box;
-	border: 2px solid transparent;
+	border: 1px solid transparent;
 	border-radius: 5px;
 	transition: 1.0s;
-	&:vertical {
-		&:hover {
-			background-color: rgb(110, 110, 110);
-			transition: 0.3s;
-		}
-		&:active {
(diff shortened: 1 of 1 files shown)


MarquesCoding:
Only issue is that it's effecting other elements within the page (buttons).

jamiepine:
We'll need a global scrollbar, but I'd like to keep the explorer scroll completely custom, as it specifically matches the layout with compensation for the header space. 
 
CleanShot 2022-04-27 at 10 27 16@2x

sr2echa:
@jamiepine so you want the global scrollbar to match the same style as the explorer scrollbar?
or retain the explorer scrollbar and also the global one with different styles?

sr2echa:
@jamiepine now this is how the scrollbar looks (no hover option)
image
should I add hover effect tho?

sr2echa:
@jamiepine like you said, now the default scrollbar (based on the overview page's scrollbar) will get applied by default if a custom scrollbar class is not mentioned. If a class is mentioned it will take attributes from the class.

lmk if i should make any other changest

jamiepine:
the visible scrollbar on the sidebar isn't what we're looking for, thanks for the PR but had to close

PR #46, opened 2022-04-28
New CI

Brendonovich:
Caching still needs work but it's better that we have some CI than none.


IllusionMan1212:
It seems that `libavresample-dev` has been deprecated in favor of `libswresample-dev` so there's probably no need to install it as well.

Brendonovich:
Prisma Client Rust is still rebuilding but idk how to fix it

xypnox:
The commit messages tell a tale altogether by themselves

PR #25, opened 2022-04-26
readme: add prerequisite linux packages

IllusionMan1212:
Useful for first time building instead of hunting the missing packages one by one.

Taken from the ci workflow file.

PR #47, opened 2022-04-28
added contributing guide + setup script

oscartbeaumont:


oscartbeaumont:
This will PR will replace the need for #45 and #25. Those PR's add installation documentation but this PR adds a script to automate the setup process of a local development environment.

README.md
```diff
@@ -67,34 +67,45 @@ The first implementation of a VDFS can be found in this UC Berkeley [paper](http
...
+Many of us have multiple cloud accounts, drives that aren’t backed up and data at risk of loss. We depend on cloud services like Google Photos and iCloud, but are locked in with limited capacity and almost zero interoperability between services and operating systems. Photo albums shouldn’t be suck in a device ecosystem, or harvested for advertising data. They should be OS agnostic, permanent and personally owned. Data we create is our legacy, that will long outlive us—open source technology is the only way to ensure we retain absolute control over the data that defines our lives, at unlimited scale.
```
mmattbtw, line 70:
```suggestion
Many of us have multiple cloud accounts, drives that aren’t backed up and data at risk of loss. We depend on cloud services like Google Photos and iCloud, but are locked in with limited capacity and almost zero interoperability between services and operating systems. Photo albums shouldn’t be stuck in a device ecosystem, or harvested for advertising data. They should be OS agnostic, permanent and personally owned. Data we create is our legacy, that will long outlive us—open source technology is the only way to ensure we retain absolute control over the data that defines our lives, at unlimited scale.
```

should probably keep this as "stuck"

.github/scripts/setup-system.sh
```diff
@@ -0,0 +1,66 @@
+#!/bin/bash
```
ForsakenHarmony, line 1: Just found this repo and was looking around randomly, just thought I'd add a random comment here

I think this might be better in the root directory (or anything outside `.github/` as `.github/` is meant just for files directly related to GitHub (e.g. workflows, pr templates, dependabot))

Maybe `.scripts/`?

PR #55, opened 2022-04-29
Accessibility improvements

ryanhaticus:
This pull request introduces one vast improvement for Spacedrive's landing pages in regards to accessibility.

Typically, those with screen readers and other assistive devices take advantage of the tab functionality to navigate through a webpage. As the site stands now, users are required to tab through the entire navigation bar before they can reach the main content.

This pull request introduces a "Skip to content" button which is placed before the navigation bar and is initially hidden. The functionality will appear when the button enters a focused state (as a result of hitting tab).

Skip to content button

By design, users with these devices can skip to the main content of the page instead of having to strenuously navigate the header each time they visit a page.


jamiepine:
really awesome pr, will merge soon

apps/landing/src/main.tsx
```diff
@@ -8,11 +8,21 @@ import { Footer } from './components/Footer';
...
       <div className="dark:bg-black dark:text-white ">
+        <a
+          tabIndex={1}
+          className="duration-200 -translate-y-16 focus:translate-y-0 fixed ml-8 mt-3 left-0 z-50"
+          href="#content"
+        >
+          <Button className="cursor-pointer " variant="gray">
+            Skip to content
+          </Button>
+        </a>
```
maxichrome, line 25: Does this create a nested structure like so?:

```html
<a href="#content">
  <button>
    Skip to content
  </button>
</a>
```

we should make sure that the Button component is able to just be an anchor tag on its own (maybe via `type="link"` and an `href` attribute) to avoid creating extra structure that may actually be _more_ disorienting to SR users

maxichrome: (this is not a criticism necessarily -- I'm unfamiliar with the sd-ui button component, just throwing this comment out there. if we had easily accessible test deploys I would just look at that instead of making this comment)

maxichrome: looking into this now: it does. going to work on fixing this and then will PR it into @ryanhaticus' fork

maxichrome: @ryanhaticus Opened a PR into your PR to resolve this: https://github.com/ryanhaticus/spacedrive/pull/1

maxichrome:
My changes from ryanhaticus#1 got merged; approved CI runs, will try to get this merged today if possible :D

maxichrome (commented):
this should fix the TypeScript CI run

maxichrome:
> this should fix the TypeScript CI run

it did not. that separate LinkButton component idea is pretty appealing now @ryanhaticus :rofl:

maxichrome:
Dedicating some time to this today :v:

ryanhaticus:
@maxichrome Feel free to push when you have this back up and running.

maxichrome:
I'm just gonna merge despite the TypeScript build step failing once the core builds succeed because the TS and JS steps are failing on the [latest `main` commits](https://github.com/spacedriveapp/spacedrive/commits/main?before=8ea2dcc7270a6dfdf0be8176059bb06eeb3ce31e+35&branch=main&qualified_name=refs%2Fheads%2Fmain)

maxichrome:
Thanks for the contrib @ryanhaticus and sorry it took some extra time to get everything taken care of!

PR #76, opened 2022-04-30
Support Docker

midrissi:
Reasons of this PR:
---

- Enhance the developer experience, by building a single reusable Docker image containing the system dependencies so that developers can focus on developing features instead of wasting time debugging the right version of system dependencies
- Make it easier for new contributors to get up and running just by having very few dependencies: **Docker**
- Make it easier for new contributors to start developing a package. For example, in order to start developing the **landing** package, follow these steps:

```bash
# Build the docker image
docker build -t spacedrive .

# Run a new container
docker run -v `pwd`/apps/landing:/app/apps/landing -it spacedrive /bin/bash

# Run the dev command:
pnpm landing dev --host 0.0.0.0
```

- Once the docker image is published (on [dockerhub](https://hub.docker.com/) or alternative), it will be very easy for everyone to try [spacedrive](https://www.spacedrive.app/) on their own machines.

### DIFF
--- a/.github/scripts/setup-system.sh
+++ b/.github/scripts/setup-system.sh
@@ -15,24 +15,29 @@ if [ $? -eq 1 ]; then
 fi
 
 if [[ "$OSTYPE" == "linux-gnu"* ]]; then
+        if (( $EUID != 0 )); then
+                echo "Please run as root"
+                exit
+        fi
+
         if which apt-get &> /dev/null; then
                 echo "Detected 'apt' based distro!"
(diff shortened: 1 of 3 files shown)


jamiepine:
Looks really great, will test shortly

jamiepine:
Spoke to @oscartbeaumont, looks like this would only work for landing page development, as Tauri (the desktop app) needs to run outside docker, but I haven't tested so I can't confirm.

oscartbeaumont:
As Jamie said I don't think we should merge this as it will cause more confusion than it will help due to Docker not being designed to run GUI applications. Docker would also not support Windows (Docker on Windows runs in Linux VM). I am working on a setup script for Windows and Linux which should make the development experience much better. An early version of that script is available in the repo and documented in the contributing guide.

PR #80, opened 2022-04-30
fix rounding on windows

mmattbtw:
this should remove the rounding seen on this image: https://cdn.discordapp.com/attachments/960383223421075507/970081518716325918/success.PNG

this is a draft because i'm not on a windows machine at the moment and can't test it: so if someone could do that for me that would be appreciated :) 

edit: current progress on Windows:

### DIFF
--- a/packages/interface/src/App.tsx
+++ b/packages/interface/src/App.tsx
@@ -63,7 +63,7 @@ function AppLayout() {
       setHasWindowBorder(false);
     }
     else {
-      setHasWindowBorder(false);
+      setIsWindowRounded(false);
     }
   }, []);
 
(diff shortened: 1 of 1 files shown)


jamiepine:
just fyi your changes currently would break macos rounding, I know its only a draft tho

mmattbtw:
fixed the rounding on macos, just need someone to test on windows

maxichrome:
I believe this has been fixed in recent changes. Please reopen if this is still broken on Windows

https://github.com/spacedriveapp/spacedrive/blob/b21df0e57f5dced4fc3ea87037cd22e2f4fe12b0/packages/interface/src/App.tsx#L65-L67

Poopooracoocoo:
Hi. I honestly don't know much about Spacedrive but I'm a UX person. I just wanted to ask if windows are rounded on Windows 7 and 11

oscartbeaumont:
My understanding is only macOS has rounded corners at the moment.

We will need to add a system for getting the Windows version because to my understanding Windows 11 needs rounded corners while Windows 10 doesn't.

Windows 7 is EOL so I don't see us supporting it.

PR #83, opened 2022-05-01
make directories not have extensions

### LINKED ISSUE #78: Folders containing `.` are having a type associated with it
**Describe the bug**
Any folder that contains a `.` in it, will have a type associated with it: this could cause a folder to have some random type at the end and the actual folder name could be cut off.

**To Reproduce**
1. Make a new folder: ex: `spacedrive.app`
2. Add it to Spacedrive
3. Observe that `app` is the type.

**Expected behavior**
I would expect it to ignore adding a type since it's a folder, and folders don't usually have types.

**Screenshots**
image
image

**Desktop (please complete the following information):**
 - OS: macOS
 - Version: Commit c4723ac I believe

mmattbtw:
fixes #78

### DIFF
--- a/core/src/file/indexer/scan.rs
+++ b/core/src/file/indexer/scan.rs
@@ -179,12 +179,12 @@ fn prepare_values(
 
   // if 'file_path' is a directory, set extension to an empty string to avoid periods in folder names
   // - being interpreted as file extensions
-  if !file_path.is_dir() {
-    extension = extract_name(file_path.extension());
-    name = extract_name(file_path.file_stem());
-  } else {
+  if file_path.is_dir() {
     extension = "".to_string();
     name = extract_name(file_path.file_name());
+  } else {
+    extension = extract_name(file_path.extension());
+    name = extract_name(file_path.file_stem());
   }
 
   let materialized_path = match file_path.to_str() {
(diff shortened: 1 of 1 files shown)


core/src/file/indexer/scan.rs
```diff
@@ -172,8 +172,20 @@ fn prepare_values(
...
+
+  // if 'file_path' is a directory, set extension to an empty string to avoid periods in folder names
+  // - being interpreted as file extensions
+  if !file_path.is_dir() {
+    extension = extract_name(file_path.extension());
+    name = extract_name(file_path.file_stem());
+  } else {
+    extension = "".to_string();
+    name = extract_name(file_path.file_name());
+  }
```
maxichrome, line 188: since this is an if-else, should we avoid inverting the `is_dir` call so we don't have to double-negative when reading the code?

```suggestion
  if file_path.is_dir() {
    extension = "".to_string();
    name = extract_name(file_path.file_name());
  } else {
    extension = extract_name(file_path.extension());
    name = extract_name(file_path.file_stem());
  }
```

maxichrome, line 188: can we move the comments inside of the braces, or is it preferred to have it outside? I think it would be more readable to put it inside, but it's a minor thing

maxichrome (commented):
Not an official review or anything but just a few thoughts ^_^

core/src/file/indexer/scan.rs
jamiepine: lol

PR #86, opened 2022-05-01
Added ffmpeg dependencies for fedora

kor-pixel:
Added ffmpeg dependencies for fedora as requested.
Fedora still requires nodejs to be installed seperately.


oscartbeaumont:
Thanks!

PR #89, opened 2022-05-01
Add execute permissions to setup-system script

maxichrome:
This updates the `setup-system.sh` script to have the permissions as denoted in **CONTRIBUTING.md**.
Execute perms propagate in git (usually, <abbr title="in my experience">imx</abbr>) so this theoretically should remove the need for the `chmod` command found on [line 41 of the contributing guide][l41].

[l41]: https://github.com/spacedriveapp/spacedrive/blob/c4723ac178f50f5229bfec899c9f3bd43d2d0d10/CONTRIBUTING.md#L41


maxichrome (commented):
lol @ this showing the change as "empty file" instead of something like 'no changes' — the file is NOT EMPTY but just the file permissions were updated

PR #90, opened 2022-05-01
Reformat markdown files (prettier)

maxichrome:
Small ting: used prettier to format markdown files because they do be a little outta style and vscode do be reformatting the whole thing every time I edit 😵‍💫.


maxichrome:
not really necessary at the moment so I'll just drop this off — we should probably enforce some formatting rules at some point though

PR #112, opened 2022-05-04
Small changes to package.json

Zariaa27:
Small change to package.json, please, start versioning.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "spacedrive",
-  "version": "0.0.0",
-  "private": true,
+  "version": "0.0.0", // please, start versioning.
+  "private": false,
   "scripts": {
     "prep": "pnpm db:gen",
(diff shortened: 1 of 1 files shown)


oscartbeaumont:
I don't think this change is useful so I am going to close it.

We are going to keep `private` set to true as Spacedrive and its packages will not be published to npm in the near future.

We will start including versions on Spacedrive when we are ready to launch the alpha version as doing it before adds no value. I would not be surprised if we use GitHub Releases/GitHub Actions to do that instead of the package.json.

Brendonovich:
Versioning the root `package.json` is not required as this is a monorepo. Additionally, no other packages need to be versioned as they will not be published on npm.

Zariaa27:
My comments weren't passive agressive at all, at least I didn't want them to come off as that, sorry if they did.

PR #115, opened 2022-05-05
Renamed .todo to todo.md

abhnva:
I don't know if I'm supposed to do this though


Brendonovich:
This is intentional for a VSCode extension that Jamie uses, it's not just a markdown file.

PR #149, opened 2022-05-12
Add a type-check script

Harry-Hopkinson:
Add a script to check if TS code is valid using the same method CI uses. Running `pnpm typecheck` in the root executes this check.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,6 @@
     "node": ">=14.0.0"
   },
   "dependencies": {
-    "zustand": "^3.7.2",
-    "typescript": "4.6.4"
+    "zustand": "^3.7.2"
   }
 }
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -35,6 +36,7 @@
     "node": ">=14.0.0"
   },
   "dependencies": {
-    "zustand": "^3.7.2"
+    "zustand": "^3.7.2",
+    "typescript": "4.6.4"
```
maxichrome, line 40: remove this, the script you added calls the existing `tsc` binary in each workspace

package.json
```diff
@@ -19,7 +19,8 @@
     "core": "pnpm --filter @sd/core -- ",
     "docs": "pnpm --filter @sd/docs -- ",
     "client": "pnpm --filter @sd/client -- ",
-    "server": "pnpm --filter @sd/server -- "
+    "server": "pnpm --filter @sd/server -- ",
+    "typecheck": "pnpm -r exec tsc"
```
maxichrome, line 23: Thanks for this, could prove very useful!

maxichrome (changes requested):

PR #184, opened 2022-05-24
update licence field to be GPL-3.0-only

oscartbeaumont:
Solves issue brought forward in discussion #182.


maxichrome (approved):
Good looking out — keeps everything in line :)

Brendonovich (approved):
👍

PR #195, opened 2022-05-29
Only countup on first load.

xPolar:
No issue, was talked about in Discord VC.

This PR will make it so the stats don't count up every single time, only on the first load. It also makes the StatItem hidden if the amount is 0.


jamiepine:
Great work!!

PR #196, opened 2022-05-29
Store isExperimental in local storage.

xPolar:
No issue, discussed in Discord VC.

Store isExperimental in local storage so the debug option in the experimental menu stays consistent on reload.


jamiepine (dismissed):

packages/interface/src/components/device/Stores.tsx
```diff
@@ -1,5 +1,19 @@
...
+const setLocalStorage = (key: string, value: any) =>
+	window.localStorage.setItem(key, JSON.stringify(value));
+
+type NodeState = {
+	isExperimental: boolean;
+	setIsExperimental: (experimental: boolean) => void;
+};
+
+export const useNodeStore = create<NodeState>((set) => ({
+	isExperimental: getLocalStorage('isExperimental') || false,
```
maxichrome, line 13: does this query local storage every time the value is accessed or only if setIsExperimental has been called?

maxichrome (approved):
Sweet, having this persisted will be super nice; just one comment to consider re: refetching from localStorage.

packages/interface/src/components/device/Stores.tsx
xPolar: wdym

maxichrome: as in, does this call `getLocalStorage()` every time you access `useNodeStore().isExperimental`, or only when `create()` is called initially and when the setter is run?

xPolar: It does it every time.

maxichrome: Should be fine but we could revisit later for efficiency and all that good stuff

maxichrome (approved):
pog

PR #199, opened 2022-05-29
fix: windows - window borders

KodingDev:


mmattbtw (approved):
it work on le mac still
image

maxichrome (approved):
Looks good! appProps shouldn't change much (if at all) so I don't think recomputation will be too expensive atm

Brendonovich (approved):

PR #200, opened 2022-05-29
fix: volumes dropdown

KodingDev:


maxichrome (approved):
I'm happy to merge this for now and we can rework the transition later on. Thanks for your contributions @KodingDev :)

PR #204, opened 2022-05-30
Oscar Changes - Update Deps and Core cleanup

oscartbeaumont:
Changes:
 - [update dependencies across workspace](https://github.com/spacedriveapp/spacedrive/commit/562efe3c6b8b0d9dd9ad7f62150ffdc60a20cbfd) - both Rust and Node deps were updated to the latest
 - [remove unused code + cleanup Rust imports](https://github.com/spacedriveapp/spacedrive/commit/6e0c944d2e48a7a5fd8597fea2a33e2b16b0181f) - removal of code in the core that isn't being used.
 - [remove anyhow](https://github.com/spacedriveapp/spacedrive/commit/a6551e83806f09d23e6debf3fe0362ebd08384ab) - this has been replaced by `thiserror` which was already being used throughout the codebase.


core/src/encode/metadata.rs
```diff
@@ -46,91 +44,91 @@ pub struct AudioStream {
...
-				media_item.steams.push(stream_item);
-			}
-			println!("{:#?}", media_item);
-		}
-
-		Err(error) => println!("error: {}", error),
-	}
-	Ok(())
-}
+// fn extract(iter: &mut Iter, key: &str) -> Option<String> {
```
Brendonovich, line 47: I assume there's a reason for all this to be comment it out, but why is that?

core/src/encode/thumb.rs
```diff
@@ -126,7 +128,7 @@ pub fn generate_thumbnail(file_path: &PathBuf, output_path: &PathBuf) -> Result<
 		imageops::FilterType::Triangle,
 	));
 	// Create the WebP encoder for the above image
-	let encoder: Encoder = Encoder::from_image(&img).map_err(|_| anyhow::anyhow!("jeff"))?;
+	let encoder: Encoder = Encoder::from_image(&img).unwrap();
```
Brendonovich, line 131: There's still a large amount of unwraps being used. I'm guessing you aren't trying to introduce new error types but cleanup existing ones?

core/src/file/cas/checksum.rs
```diff
@@ -1,34 +1,34 @@
...
 static SAMPLE_SIZE: u64 = 10000;
 
-fn read_at(file: &File, offset: u64, size: u64) -> Result<Vec<u8>> {
+fn read_at(file: &File, offset: u64, size: u64) -> Result<Vec<u8>, io::Error> {
 	let mut buf = vec![0u8; size as usize];
 
 	#[cfg(target_family = "unix")]
-	file.read_exact_at(&mut buf, offset)?;
+	{
+		use std::os::unix::prelude::FileExt;
```
Brendonovich, line 18: For consistency you could include the whole prelude
```suggestion
		use std::os::unix::prelude::*;
```

core/src/node/state.rs
```diff
@@ -39,15 +38,15 @@ lazy_static! {
 	static ref CONFIG: RwLock<Option<NodeState>> = RwLock::new(None);
```
Brendonovich, line 38: This just shouldn't be stored like this, though I'm guessing that's out of scope of this PR?

core/src/util/db.rs
```diff
@@ -1,18 +1,30 @@
...
 static MIGRATIONS_DIR: Dir = include_dir!("$CARGO_MANIFEST_DIR/prisma/migrations");
 
-pub fn sha256_digest<R: Read>(mut reader: R) -> Result<Digest> {
+#[derive(Error, Debug)]
+pub enum DatabaseError {
+	#[error("Unable to initialize the Prisma client")]
+	ClientError(#[from] prisma::NewClientError),
+}
+
+pub async fn create_connection(path: &str) -> Result<PrismaClient, DatabaseError> {
```
Brendonovich, line 20: Is this function necessary? It just wraps one function call and I assume is only called in one place.

Brendonovich (changes requested):
More changes are definitely needed but if this is part of a larger multi-PR effort then lgtm

core/src/encode/thumb.rs
oscartbeaumont: I intend to go back and cleanup the errors more in the future, as with the rest of the codebase. The codebase is completely covered in panics or error being ignore `.map(|| ())` so I don't think this makes a major difference now.

core/src/node/state.rs
oscartbeaumont: Already working on that today. I am merging in changes from the core refactor one by one. I just wanna get my changes in master as quickly as possible so we don't get messy merge conflicts.

core/src/encode/thumb.rs
maxichrome: > I intend to go back and cleanup the errors more in the future, as with the rest of the codebase. 

Should we make sure this is tasked out?

> The codebase is completely covered in panics or error being ignore `.map(|| ())` so I don't think this makes a major difference now.

Does `.unwrap()` ignore errors or cause them to panic (which I assume would be expected behavior based on your comment)?

core/src/util/db.rs
oscartbeaumont: I have already removed it with some local changes I am working on now. Will push later today. It's one Jamie had around. I think we should have a `async fn load_and_migrate(db_url: &str) -> Result<PrismaClient, MigrationError>` in that file instead of the two functions.

core/src/encode/metadata.rs
oscartbeaumont: It's unused and was throwing warnings from the Rust compiler due to that. It can all be uncommented when it's ready to be used.

core/src/encode/thumb.rs
Brendonovich: > Does `.unwrap()` ignore errors or cause them to panic (which I assume would be expected behavior based on your comment)?

`.unwrap()` forces a panic if the result is an error, which is why it's better to bubble errors upwards to a central error handling mechanism.

oscartbeaumont: I added error handling is ENG-91 part of the core refactor ENG-49.

Unwrap causes a crash which for now would be expected instead of `.map(|| ())` which ignores the error and continues execution.

Brendonovich (approved):

PR #216, opened 2022-06-02
Allow adding locations on the web.

xPolar:
Closes ENG-108 on linear.


Brendonovich:
Adding locations on web will require more than just adding a file browser, since what it actually should do is add a location from the node that the interface is connected to, not from the device that is displaying the interface. This will require fetching the files that are available on the server, displaying them in a custom file picker in the interface and performing the same operation as the desktop.

PR #220, opened 2022-06-02
fix: disappearing icons

benja:
**Problem**
The folder icon uses a gradient which requires an ID so that the path element can use it as a reference when using `fill`. However, we use these SVGs in a lot of places, meaning we have duplicate instances of the same gradient definition. This causes problems in browsers where icons might not render properly from time to time. 

**Solution**
To solve this, I first attempted playing around with the [Shadow DOM](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_shadow_DOM) to keep using SVGs directly as React components, but with the benefit of residing inside a fully encapsulated environment where styles and IDs don't leak outside. However, I wasn't very successful with this so I ended up sourcing the SVGs as images to avoid the id collisions.

The icons seem to behave the same, so I don't have any immediate concerns with this solution. Feel free to let me know if anyone has any objections.

_[SVGO](https://react-svgr.com/docs/options/#svgo) should automatically prefix ids to help avoid collisions like these, but that didn't seem to do work, so I moved away from trying to transform the SVG through svgr to create unique IDs per instance. However, this solution is still open if we want to keep using svgs over image elements, but would require a bit more hacking._ 

Image showcasing folder icons with gradients

### DIFF
--- a/apps/landing/src/pages/roadmap.tsx
+++ b/apps/landing/src/pages/roadmap.tsx
@@ -2,7 +2,7 @@ import React from 'react';
 import { Helmet } from 'react-helmet';
 import { ReactComponent as Content } from '~/docs/product/roadmap.md';
 
-import folderSvg from '../../../../packages/interface/src/assets/svg/folder.svg';
+import { Folder } from '../../../../packages/interface/src/components/icons/Folder';
 import Markdown from '../components/Markdown';
 
 function Page() {
@@ -13,7 +13,7 @@ function Page() {
 				<meta name="description" content="What can Spacedrive do?" />
 			</Helmet>
 			<div className="w-24 mb-10">
-				<img src={folderSvg} alt="Folder icon" />
+				<Folder />
 			</div>
 			<Content />
 		</Markdown>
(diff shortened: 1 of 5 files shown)


apps/landing/src/pages/roadmap.tsx
```diff
@@ -13,7 +13,7 @@ function Page() {
 				<meta name="description" content="What can Spacedrive do?" />
 			</Helmet>
 			<div className="w-24 mb-10">
-				<Folder className="" />
+				<img src={folderSvg} alt="Folder icon" />
```
Brendonovich, line 16: It would be nice to still have a `Folder` icon component rather than using raw images everywhere, since that way the alt text would be standardised.

Brendonovich (changes requested):

apps/landing/src/pages/roadmap.tsx
benja: Went right above my head to componentize this `img` element being used everywhere, but that makes a lot more sense to do! I've pushed a commit with those changes now.

Brendonovich (approved):

PR #227, opened 2022-06-07
Remove jobs tab from landing.

xPolar:
Removes jobs tab from landing as it is currently 404ing.


apps/landing/src/components/NavBar.tsx
```diff
@@ -64,7 +64,7 @@ export default function NavBar() {
 					<NavLink link="https://spacedrive.hashnode.dev">Blog</NavLink>
 					{/* <NavLink link="/change-log">Changelog</NavLink>
           <NavLink link="/privacy">Privacy</NavLink> */}
-					<NavLink link="/jobs">Jobs</NavLink>
+					{/* <NavLink link="/jobs">Jobs</NavLink> */}
```
maxichrome, line 67:
```suggestion

```

maxichrome (changes requested):
On the back of what @benja commented, here's a quick suggestion so I can apply the patch and merge that version:

maxichrome (approved):

PR #229, opened 2022-06-08
New team page

benja:
Implemented new team page.


apps/landing/src/pages/team.tsx
```diff
@@ -1,8 +1,182 @@
...
+	},
+	{
+		name: 'Vijay Sharma',
+		role: 'CEO, Belong',
+		investmentRound: 'Seed',
+		image: 'investors/sharma.jpg'
+	},
+	{
+		name: 'Naveen R',
+		role: 'NocoDB',
```
benja, line 175: As suggested in #231.

```suggestion
		role: 'Founder, NocoDB',
```

apps/landing/src/pages/team.tsx
```diff
@@ -1,8 +1,183 @@
 import React from 'react';
 import { Helmet } from 'react-helmet';
-import { ReactComponent as Content } from '~/docs/product/credits.md';
 
+import { ReactComponent as ArrowRight } from '../../../../packages/interface/src/assets/svg/arrow-right.svg';
```
Brendonovich, line 4: Can this be imported from `@sd/interface`?

docs/product/credits.md
```diff
@@ -1,16 +1,16 @@
-## Spacedrive
+# Spacedrive
```
Brendonovich, line 2: Is this file necessary anymore? It used to be the source for the team page, but now it's not used on the site.

LICENSE
```diff
@@ -1,4 +1,4 @@
-Copyright (c) 2021-present Jamie Pine.
```
Brendonovich, line 1: Is this PR overriding the `spacedrive-incorp` repo that I believe had a lot of the same changes to legal copy? That fork hasn't been updated for 22 days, so I'm not surprised if that is the case.

apps/landing/src/pages/team.tsx
maxichrome: idk if `@sd/interface` aliases to `src/` as its root, but if so we'd have to omit that from the path as well

```suggestion
import { ReactComponent as ArrowRight } from '@sd/interface/src/assets/svg/arrow-right.svg';
```

docs/product/credits.md
maxichrome: from slack

> that markdown file is intended to be embedded in the about tab of the actual app, so maybe we keep it

I interpret this as 'keep for now unless decided otherwise later on'

LICENSE
maxichrome: haven't touched that at all, yeah this is taking its place. i've not taken it down atm

benja:
> Should the column be a bit wider, and have 3 larger images per row instead of 2 on desktop? image

That is an option. It's just using the same markdown file interpreter for its content right now iirc, but we can definitely change update this. @jamiepine @maxichrome thoughts? :)

PR #231, opened 2022-06-09
VC Role updated

tomasaguiar:
Naveen's Role updated 

On the team's page, the investor "Naveen R" doesn't have a role in their company, but just the name of the company itself, so I visited "https://nocodb.com/team" and found that Naveen's role was Founder. 

Because it isn't an "issue" I didn't open one. 

Naveen's Correct Role

### DIFF
--- a/apps/landing/src/pages/team.tsx
+++ b/apps/landing/src/pages/team.tsx
@@ -172,7 +172,7 @@ const investors: Array<TeamMemberProps> = [
 	},
 	{
 		name: 'Naveen R',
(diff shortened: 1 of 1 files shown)


benja:
Hello! Great catch. I've gone ahead and added your suggested change as a suggestion to the existing PR (#229). Please use this feature next time you suggest a minor change like this to an open PR, thank you! 🙌🏼

tomasaguiar:
Hey! Glad to help. Thank you for the advice.

PR #235, opened 2022-06-11
Fix contributing link.

xPolar:
Closed ENG-131.


maxichrome (approved):
Looks good to me. If @jamiepine still uses that `.todo` feature we should revert its removal and instead move it — otherwise I'm happy to have this applied

PR #245, opened 2022-06-14
Static Generate Landing

oscartbeaumont:


jamiepine:
Incredible, it loads so fast. Great work.

jamiepine:
It's merged with main, but it won't deploy so don't worry if you need to do more for prod deployment.

PR #259, opened 2022-06-19
Install dependencies to avoid clippy errors in CI

kkaosninja:
To avoid this error => https://github.com/spacedriveapp/spacedrive/runs/6951523959?check_suite_focus=true

Closes #(issue)

### DIFF
--- a/.github/workflows/clippy.yml
+++ b/.github/workflows/clippy.yml
@@ -15,6 +15,9 @@ jobs:
       
       - run: rustup component add clippy
       
+      - name: Install Dependencies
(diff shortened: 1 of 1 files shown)


oscartbeaumont:
I am going to close this as the changes would not fix the problem and don't align with the rest of our CI setup (using the `./.github/scripts/setup-system.sh`). I will merge a fix later today once I get it working on my fork.

To check Spacedrive the Clippy workflow would need FFmpeg and other deps, along with the Prisma code to be generated.   This PR only adds a few of the required dependencies.

Thanks for taking the time to open a PR!

PR #263, opened 2022-06-20
(fix) landing: Windows performance

jamiepine:
Rasterized the blur effect into PNGs (created by exporting bloom recreation in the Figma). This has dramatically improved performance, but the bubbles still impact it. I'm leaving them in for now cause sexy.


benja:
There is a weird jump (no animation) the first time you load the landing page and go to the team page, but seems to only happen the first time you load the page (without cache).

Go here — https://spacedrive-landing-git-eng-148-landing-lag-on-0a5b17-spacedrive.vercel.app/team and hard refresh, you might be able to see it too.

benja (approved):

benja:
I'm going to merge this for the increased performance in the meantime, but we should definitely look into why that is happening

PR #286, opened 2022-06-22
Updates to ReadMe

jenniferdewan:
1st change - This sentence should be combined because its format is like for X kinds of people Y program provides Z service. While I may have not explained it too well, these two sentences would flow better together. 

2nd change - The sentence makes more sense this way.

Closes #(issue)


jamiepine (approved):

voletro:
Typo in the change, VDFS not VSFS, and also should be 
"a VDFS is not restricted to a single machine." instead of 
"VSFS is not restricted to a single machine."

jenniferdewan:
oh yeah sorry about that!

PR #304, opened 2022-06-24
Update paths to show correctly on windows

waydealphax:
Updated this file so that file paths are correct on windows.

Added an if else statement to check if the system is Win32 and if so change the full path from x/x to x\\x to conform with windows file paths


packages/interface/src/components/file/Inspector.tsx
```diff
@@ -35,20 +35,25 @@ export const Inspector = (props: {
 	location?: LocationResource;
 	selectedFile?: FilePath;
 }) => {
-	const file_path = props.selectedFile;
-	let full_path = `${props.location?.path}/${file_path?.materialized_path}`;
+	
+	const file_path = props.selectedFile,
+	file_id = props.selectedFile?.file?.id || -1;
+	if (window.navigator.platform === 'Win32') {
```
maxichrome, line 41: somewhere in the AppProps we take in a platform prop -- let's use that context instead here

packages/interface/src/components/file/Inspector.tsx
```diff
@@ -35,20 +35,25 @@ export const Inspector = (props: {
...
+		var full_path = `${props.location?.path}\\${file_path?.materialized_path}`;
+	} else {
+		var full_path = `${props.location?.path}/${file_path?.materialized_path}`;
+	}
```
maxichrome, line 45: not a big fan of using `var` declarations anywhere, period. 

```suggestion
	const file_path = props.selectedFile,
	file_id = props.selectedFile?.file?.id || -1;
	
	let full_path: string;
	
	if (window.navigator.platform === 'Win32') {
		full_path = `${props.location?.path}\\${file_path?.materialized_path}`;
	} else {
		full_path = `${props.location?.path}/${file_path?.materialized_path}`;
	}
```

maxichrome, line 45: use an array here instead, maybe:

```suggestion
	const pathSeparator = window.navigator.platform === 'Win32' ? '\\' : '/';

	full_path = [props.location?.path, file_path?.materialized_path].join(pathSeparator);
```

packages/interface/src/components/file/Inspector.tsx
```diff
@@ -63,95 +68,95 @@ export const Inspector = (props: {
...
+					{/* <MetaItem title="Date Last Modified" value={file?.date_modified} />
             <MetaItem title="Date Indexed" value={file?.date_indexed} /> */}
-					</div>
-				)}
-			</div>
-		</Transition>
+				</div>
+			)}
+		</div>
+		// </Transition>
```
maxichrome, line 160: what does this massive diff have to do with path separators?

core/src/sys/locations.rs
```diff
@@ -213,6 +220,29 @@ pub async fn create_location(ctx: &CoreContext, path: &str) -> Result<LocationRe
...
+		.await?;
+
+	ctx.emit(CoreEvent::InvalidateQuery(ClientQuery::SysGetLocations))
+		.await;
+
+	println!("Location {} deleted", location_id);
+
+	Ok(())
+}
+
```
maxichrome, line 245: does this help path separators display properly? I'm always glad to see updated functionality but this seems out-of-scope?

maxichrome (changes requested):
great looking out! — idk if it's just GitHub showing extraneous changes in the diff but there seem to be a lot of unrelated changes in here... 

requested a couple changes with suggestions for changes core to this PR's purpose though.

packages/interface/src/components/file/Inspector.tsx
waydealphax: i was basing off of jamies eng branch because i saw he was working within the same area :p

core/src/sys/locations.rs
waydealphax: refer to prev. comment most like related to the basing off of jamies branch

waydealphax:
@maxichrome looks fine to me now, have a check and lmk

maxichrome:
> looks fine to me now, have a check and lmk

still includes lots of unrelated changes. any way you can base this on `main` / mark it as a draft PR until the related changes are merged in?

oscartbeaumont:
Thanks for the PR but I don't think in its current form we will want to merge it. To fix this issue work will need to be done on the Rust backend to properly understand the operating system the file was created on and return the correct path. The current fix in this PR will break when it comes to hooking the app up to the P2P sync layer because it uses the OS your running the app on (`appProps.platform`) to decide how to display the path instead of the operating system of the computer the file was created on.

PR #307, opened 2022-06-25
Switch license to AGPL

Brendonovich:
Unlike GPL, the AGPL triggers when software is accessed over a network. This would stop Spacedrive and its libraries from being forked, modified and served as a SaaS platform without the modifications being made public. This does not prevent Spacedrive and its libraries from being used inside commercial projects, but it does extend the boundary of the original GPL license.

For all of Spacedrive's code to be relicensed we must get consent from all contributors. If you are a contributor, please comment on this PR indicating that you accept the license change. If you do not accept this change or fail to accept it by 27 June 8:00PM PST then your changes will be removed.

- [x] @jamiepine 
- [x] @Brendonovich  
- [x] @maxichrome 
- [x] @oscartbeaumont 
- [x] @benja 
- [x] @xPolar 
- [x] @LayZeeDK 
- [x] @KodingDev 
- [x] @sr2echa 
- [x] @CreatingBytes 
- [x] @FahimFBA 
- [x] @mmattbtw
- [x] @Harry-Hopkinson
- [x] @XSPGMike 
- [x] @he1d1 
- [x] @IllusionMan1212 
- [x] @ecwyne 
- [x] @krumware 
- [x] @stevharve 
- [x] @mark-strudwick 
- [x] @chroxify 
- [x] @TheUltDev
- [ ] @̶i̶l̶k̶k̶a̶ (Edit: Only commit fixed a broken link in the readme, not a code change. No response, not waiting.)
- [x] @CodePurble 
- [x] @abhnva 
- [x] @cnrad 
- [x] @MarquesCoding 
- [x] @alsonick 
- [x] @CodeWithBryan 
- [ ] @̶f̶i̶v̶e̶s̶t̶o̶n̶e̶s̶ (Edit: Added a single apostrophe to the readme. No response, not waiting.)
- [ ] @̶y̶k̶a̶b̶u̶s̶a̶l̶a̶h̶ (Edit: The only commit by this user was already overwritten. No response, not waiting.)
- [x] @kor-pixel 
- [ ] @̶r̶y̶a̶n̶h̶a̶t̶i̶c̶u̶s̶ (Edit: Can not be found on the contributor list. No response, not waiting.)

No external contributions should be accepted into this repository until this pull request has been merged, which will not happen until all contributors have agreed to the license change.

I accept this license change.

PR #314, opened 2022-06-26
Tool Tips added

berbaroovez:
While using the app, I had no idea what some of the icons were, so I added a tooltip component using radix-ui.


maxichrome:
Now that #307 is merged I'll take a peek at this as soon as I've got a chance :)

xPolar:
Looks like this will close #330, didn't see this PR before making the issue public.

oscartbeaumont (approved):

Issue #77, opened 2022-04-30
Setup scripts


oscartbeaumont:
- [x] Windows
- [ ] Linux - Arch
- [x] Linux - Ubuntu
- [x] Linux - PopOS
- [x] Linux - Debian
- [ ] Linux - Fedora
- [x] Convert CI to use the script

oscartbeaumont:
- [ ] Windows
- [ ] Linux - Arch
- [ ] Linux - Ubuntu
- [ ] Linux - PopOS
- [ ] Linux - Debian
- [ ] Linux - Fedora

kolton-musgrove:
Could I take care of writing the windows setup script?

I think a logical flow of operations would be:

1. Check for rust, cargo, nodejs, and npnm installations
2. run pnpm i
3. run pnpm prep

Is there other functionality you wanted the scripts to have?

oscartbeaumont:
@kolton-musgrove Feel free to do the Windows setup script!

The functionality you listed sounds good. You will have to install the required Tauri dependencies along with `ffmpeg` or its required dependencies so that the Rust code will build.

abhnva:
What about the Linux script?

oscartbeaumont:
What about them?

The alpha version of the Linux script [already exists in the repo](https://github.com/spacedriveapp/spacedrive/blob/main/.github/scripts/setup-system.sh). I just need to get Arch and Fedora working properly, which is hard as I only have a Debian based system (although imma test with Docker). I have had a busy few days but I will push my changes before mid-week.

CodePurble:
Hi @oscartbeaumont, I'd be happy to help out with the Arch Linux install.

voletro:
I am rewriting the windows setup script, given that so many people are having issues.

From past experience, the ffmpeg binaries are need in path, including each library ffmpeg needs and their respective binaries.

I'll keep updated here and will test with multiple machines.

oscartbeaumont:
Awesome! Feel free to message on here or Discord if you got any questions!

Issue #306, opened 2022-06-24
[ENG-189] AppImage fails to build 


rockerBOO:
### Describe the bug

`pnpm desktop build` panics when bundling the AppImage on linux.

```
thread '<unnamed>' panicked at 'couldn't find a square icon to use as AppImage icon', /build/tooling/bundler/src/bundle/linux/appimage.rs
```

### Reproduction

Contributor flow and then: 

`pnpm desktop build` 

### Expected behavior

```
Finished release [optimized] target(s) in 48.40s
Bundling spacedrive_0.1.0_amd64.deb (/mnt/900/builds/spacdrive/target/release/bundle/deb/spacedrive_0.1.0_amd64.deb)
Bundling spacedrive_0.1.0_amd64.AppImage (/mnt/900/builds/spacdrive/target/release/bundle/appimage/spacedrive_0.1.0_amd64.A
Finished 2 bundles at:
    /mnt/900/builds/spacdrive/target/release/bundle/deb/spacedrive_0.1.0_amd64.deb
    /mnt/900/builds/spacdrive/target/release/bundle/appimage/spacedrive_0.1.0_amd64.AppImage

```

### Platform and versions

```Shell
cargo 1.61.0 (a028ae4 2022-04-29)
rustc 1.61.0 (fe5b13d68 2022-05-18)
node: 16.14.2
pnpm: 7.3.0
Linux Monolith 5.18.2-zen1-1-zen #1 ZEN SMP PREEMPT_DYNAMIC Mon, 06 Jun 2022 19:59:01 +0000 x86_64 GNU/Linux
```

### Additional context

https://github.com/tauri-apps/tauri/blob/dev/tooling/bundler/src/bundle/linux/appimage.rs#L68-L72 where the panic is from in Tauri.

It is fixed by doing the following in `tauri.linux.conf.json` (expanding from just `icons/icon.icns`)

```
			"icon": [
				"icons/icon.icns", 
				"icons/32x32.png",
				"icons/128x128.png"
			],
```

I can make this change if needed.

oscartbeaumont:
Thanks for reporting this. It will be fixed as part of PR #321

PR #321, opened 2022-06-30
Rewrite of setup-system.ps1 for Windows development setup

voletro:
This is a full rewrite of setup-system.ps1 that sets up a user's machine for Spacedrive development. 

It currently:

1) Checks for Rust and Cargo
2) Installs pnpm (if not installed)
3) Installs the latest version of Node.js using pnpm
4) Installs LLVM (compiler for ffmpeg-rust)
4) Downloads ffmpeg and set as an environment variable

The CONTRIBUTING.md has also been updated to include info on running the script.

This PR helps close issue #77 by creating a working Windows setup script.


voletro:
This commit should make that happen.

I had to add LLVM to the ci as the script downloads and runs an exe to install it.

I think the error was that I was trying to change the title and background colour of the window, so I also removed that.

oscartbeaumont:
Thanks for making the change. It looks like CI is failing to find `libavutil`. If you have some time could you look into this. The old script was working in CI so maybe it could help as a reference for debugging.

voletro:
Turns out GH Actions uses GITHUB_PATH and GITHUB_ENV rather than PATH and ENV. I can add a check for ci and set the correct environment variables accordingly.

oscartbeaumont:
If you enable GitHub Actions on your fork, they will run and you can check if it is working.

voletro:
Ok, ci now works, and ffmpeg-sys-rust now builds successfully. 

I can make a new pr if you would like, given there are so many commits in this one.

oscartbeaumont:
I can squash the commits when we merge, so don't worry about how many it is.

It looks like CI is now failing in Linux on issue #306. I will give this one a look in the next day or so, so we can get this merged. Thanks for all your help on this PR.

voletro:
All good, love to help.

oscartbeaumont (approved):

oscartbeaumont:
Whoops, completely forgot to squash commits, my bad.

PR #323, opened 2022-07-03
Client tries to reconnect when sockets close

### LINKED ISSUE #292: [ENG-112] Reconnect if websocket connection lost on web
If the websocket connection is lost it will not be reconnected when the server comes back online.

he1d1:
It tries to reconnect after:
1. 1 second
2. 2 seconds
3. 5 seconds
4. 10 seconds (keeps trying this forever)

Each attempted reconnect puts an error in the console, this can be suppressed by adding a `error` event listener to `startWebsocket`.

I didn't suppress it because I think this really should have some way of telling the user that its reconnecting, ideally a toast notification, but for the time being, an error is probably okay. 

Toast notifications wouldn't be too hard to add, the [react-hot-toast](https://react-hot-toast.com/) package is pretty nice, but they aren't too hard to add manually so I didn't want to do that without getting opinions of more people. 

Closes #292

### DIFF
--- a/apps/web/src/App.tsx
+++ b/apps/web/src/App.tsx
@@ -1,35 +1,47 @@
 import { BaseTransport } from '@sd/client';
-import { ClientCommand, ClientQuery, CoreEvent } from '@sd/core';
+import { ClientCommand, ClientQuery } from '@sd/core';
 import SpacedriveInterface from '@sd/interface';
 import React, { useEffect } from 'react';
 
-const timeouts = [1, 2, 5, 10];
-
-const startWebsocket = (timeoutIndex = 0) => {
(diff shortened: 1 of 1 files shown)


Brendonovich (changes requested):
I'd like to see some jitter introduced as well as the backoff. If our server goes offline then all clients will start repolling at the same time, effectively DDOSing ourselves. Adding jitter means that all the requests will be offset somewhat, spreading out the load on our servers.

Also, doesn't `setTimeout` accept a time in milliseconds? If so then your current implementation would be repoll way too fast!

oscartbeaumont:
@Brendonovich issues fixed. Give it a review and merge if your happy.

Brendonovich (approved):

PR #327, opened 2022-07-07
Focus search bar on press Super+K

### LINKED ISSUE #326: [ENG-197] Super+K doesn't select the search bar
### Describe the bug

The Search bar doesn't focus when super+k is pressed in location file browser.

### Reproduction

1. Add a Location on the left hand side menu. (anywhere in the file system it doesn't matter)
2. press super+K

### Expected behavior

When super+k is pressed the search box should be selected so that the user can type in it.

### Platform and versions

```Shell
MacOS 12.3
```

### Stack trace

```Shell
pnpm: 7.5.0
cargo 1.62.0 (a748cf5a3 2022-06-08)
rustc 1.62.0 (a8314ef7d 2022-06-27)
```

### Additional context

_No response_

RockBacon9922:
Implements focus on search bar when super+k is pressed.
Hides the key combo hint when search bar is focused.

Closes #326


packages/interface/src/components/layout/TopBar.tsx
```diff
@@ -67,6 +67,18 @@ export const TopBar: React.FC<TopBarProps> = (props) => {
...
+	const [showSuperK, setShowSuperK] = React.useState(true);
+
+	//create function to focus on search box when cmd+k is pressed
+	const searchRef: any = React.useRef(null);
+	React.useEffect(() => {
+		document.addEventListener('keydown', (e) => {
+			if (e.metaKey && e.key === 'k' && searchRef.current) {
+				searchRef.current.focus()
+			}
+		})
```
maxichrome, line 79: this should be modified to use a keyboard shortcut in the desktop app's `menu.rs` — this code should only activate on the web client

you'll need to use a Tauri event to listen and run `focus()`

packages/interface/src/components/layout/TopBar.tsx
```diff
@@ -67,6 +67,18 @@ export const TopBar: React.FC<TopBarProps> = (props) => {
...
+	},[])
```
maxichrome, line 80: this useEffect should have a cleanup function to remove the event listener on derender

packages/interface/src/components/layout/TopBar.tsx
```diff
@@ -90,13 +102,20 @@ export const TopBar: React.FC<TopBarProps> = (props) => {
...
+								setShowSuperK(true);
+							}}
 							className="w-32 h-[30px] focus:w-52 text-sm p-3 rounded-lg outline-none focus:ring-2  placeholder-gray-400 dark:placeholder-gray-500 bg-[#F6F2F6] border border-gray-50 dark:bg-gray-650 dark:border-gray-550 focus:ring-gray-100 dark:focus:ring-gray-600 transition-all"
 						/>
-						<div className="space-x-1 absolute top-[2px] right-1">
+						{showSuperK && <div className="space-x-1 absolute top-[2px] right-1">
 							<Shortcut chars="⌘K" />
 							{/* <Shortcut chars="S" /> */}
-						</div>
+						</div>}
```
maxichrome, line 118: I believe this should use a class to hide instead of entirely conditionally rendering. could use a tailwind `group` utility -- on `group-focus` then hide this

maxichrome (changes requested):
Thank you for adding this!! I would like to see some of the requested changes made and then I'm happy to merge!

packages/interface/src/components/layout/TopBar.tsx
RockBacon9922: Surely running it through tauri would remove the feature from the web version?

RockBacon9922: @maxichrome I have done some looking around and come across this.
https://tauri.app/v1/api/js/modules/globalshortcut/
I am trying to implement it but I am getting a dependency warning when I try to import:
    import { register, unregister } from '@tauri-apps/api/globalShortcut'
Any ideas?

packages/interface/src/components/layout/TopBar.tsx
RockBacon9922: Resolved used tailwind peer class to control visibility of the element.

packages/interface/src/components/layout/TopBar.tsx
jamiepine: Don't import any Tauri apis in the interface package, modify app props and import into Desktop and bind via prop

RockBacon9922: Unfortunately I still don't understand what you are requesting @jamiepine.

maxichrome: I will work on solving this tonight

Brendonovich:
IMO `Super+K` is the wrong keybind for the search bar. It's usually used to access an app's command pallete, which Spacedrive doesn't have yet. Instead I propose two other options:
- `Super+L`: Common keybind to access the search bar on web browsers
- `/`: Common keybind to access search functions within websites. Works in GitHub, Google, YouTube, Linear and many other sites. I think this is the best one.

Also, I don't think it should be done in Rust. These keybinds are used by many applications, only detecting them when the app is focused is a pretty standard expectation. To go further and hijack them would be irresponsible.

RockBacon9922:
> IMO `Super+K` is the wrong keybind for the search bar. It's usually used to access an app's command pallete, which Spacedrive doesn't have yet. Instead I propose two other options:
> 
> * `Super+L`: Common keybind to access the search bar on web browsers
> * `/`: Common keybind to access search functions within websites. Works in GitHub, Google, YouTube, Linear and many other sites. I think this is the best one.
> 
> Also, I don't think it should be done in Rust. These keybinds are used by many applications, only detecting them when the app is focused is a pretty standard expectation. To go further and hijack them would be irresponsible.

Ok I will happily make these changes if @maxichrome and or @Brendonovich is happy with this proposition.

jamiepine:
Happy to merge this, we will however most likely replace this code as soon as a keybind system is in place. Ideally we'd want a universal way to register and override keybinds rather than directly calling `document.addEventListener` in components

jamiepine (changes requested):
Move code to SearchBar.tsx

maxichrome:
I'll resolve merge conflicts for this PR and make some necessary changes to the feature's working and should have this mergeable shortly

packages/interface/src/components/layout/TopBar.tsx
oscartbeaumont: I don't think this needs to be done through Tauri because we are targeting the interface at both the desktop app and web app (where we don't have Tauri). It definitely does need to be in the Tauri menu tho, so macOS is happy.

oscartbeaumont:
The way the keyboard shortcut was implemented it would preventing typing "/" anywhere in the app.

I fixed that, updated the keyboard shortcut hint on the search box and also made it so pressing `Esc` will unfocus the current input anywhere in the app. I also enabled `Cmd+A` for copy all because that's been bugging me a lot.

oscartbeaumont (approved):

PR #355, opened 2022-08-11
Eslint config + types

utkubakir:
- [X] Configure eslint
- [X] Fix types Landing
- [X] Fix types Mobile
- [X] Fix types Interface
- [X] Add "lint" script (runs eslint and tsc)

Might be good addition?
- [ ] CLI
- [ ] Lefthook

react-hooks rule complained a lot but I didn't wanna potentially break our flow or do unnecessary renders so I disabled the rule on most files.


apps/landing/src/pages/index.page.tsx
```diff
@@ -78,7 +78,7 @@ function Page() {
 				alert('An error occurred while unsubscribing from waitlist');
 			}
 		})();
-	}, []);
+	}, [urlParsed.search]);
```
Brendonovich, line 81: Not sure if we want to do this, the effect is only supposed to run once to check if a user wants to unsubscribe from the waitlist. Not a huge issue though, I doubt we're modifying search params much

Brendonovich (approved):
Nice cleanup, love all the work that's been going into mobile too

apps/landing/src/pages/index.page.tsx
utkubakir: Removed it, it's best to keep everything as it is.

PR #358, opened 2022-08-12
Fixes Tauri dependencies for Debian Buster (stable), fixes warning fo…

### LINKED ISSUE #77: Setup scripts

ned-park:
…r users without cargo or pnpm installed

set -e prevents the if statements from running when which exits with non-zero status are never informed about installing cargo or pnpm, removing it fixes that.  

Updates Tauri dependencies for Debian in setup-system.sh 

This PR completes the Linux - Debian task of issue #77 

Fixes #77 Linux - Debian task

### DIFF
--- a/.github/scripts/setup-system.sh
+++ b/.github/scripts/setup-system.sh
@@ -1,7 +1,5 @@
 #!/bin/bash
 
-set -e
-
 echo "Setting up your system for Spacedrive development!"
 
 which cargo &> /dev/null
@@ -29,7 +27,7 @@ if [[ "$OSTYPE" == "linux-gnu"* ]]; then
                 else
                         DEBIAN_FFMPEG_DEPS="libavcodec-dev libavdevice-dev libavfilter-dev libavformat-dev libavresample-dev libavutil-dev libswscale-dev libswresample-dev ffmpeg" # FFMPEG dependencies
                 fi
-                DEBIAN_TAURI_DEPS="libwebkit2gtk-4.0-dev build-essential curl wget libssl-dev libgtk-3-dev libappindicator3-dev librsvg2-dev" # Tauri dependencies
+                DEBIAN_TAURI_DEPS="libwebkit2gtk-4.0-dev build-essential curl wget libssl-dev libgtk-3-dev libayatana-appindicator3-dev librsvg2-dev" # Tauri dependencies
                 DEBIAN_BINDGEN_DEPS="pkg-config clang" # Bindgen dependencies - it's used by a dependency of Spacedrive
                 
                 sudo apt-get -y update
--- a/.github/scripts/setup-system.sh
+++ b/.github/scripts/setup-system.sh
@@ -1,22 +1,66 @@
 #!/bin/bash
 
+set -e
+
+script_failure() {
+        echo "An error occurred while performing the task on line $1" >&2
+        echo "Setup for Spacedrive development failed" >&2
+}
+
+trap 'script_failure $LINENO' ERR
+
 echo "Setting up your system for Spacedrive development!"
 
-which cargo &> /dev/null
-if [ $? -eq 1 ]; then
+if ! which cargo &> /dev/null; then
         echo "Rust was not detected on your system. Ensure the 'rustc' and 'cargo' binaries are in your \$PATH."
         exit 1
 fi
 
 if [ "${SPACEDRIVE_SKIP_PNPM_CHECK:-}" != "true" ]; then
-        which pnpm &> /dev/null
-        if [ $? -eq 1 ]; then
+        
+        if ! which pnpm &> /dev/null; then
                 echo "PNPM was not detected on your system. Ensure the 'pnpm' command is in your \$PATH. You are not able to use Yarn or NPM."
                 exit 1
         fi
 else
         echo "Skipped PNPM check!"
 fi
+
+if [ "$1" == "mobile" ]; then
+        echo "Setting up for mobile development!"
+
+         # IOS targets
+        if [[ "$OSTYPE" == "darwin"* ]]; then
+                echo "Installing IOS Rust targets..."
+
+                if ! /usr/bin/xcodebuild -version; then
+                        echo "Xcode is not installed! Ensure you have it installed!"
+                        exit 1
+                fi
+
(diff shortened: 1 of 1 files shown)


oscartbeaumont:
I am confused about what the problem with `set -e` is. What problem is it causing in the script? Personally, I think it should stay in place so the script will stop if something goes wrong so it is very clear to the user running the script that something went wrong. I feel that removing `set -e` is just removing the error but not fixing the underlying problem.

I am happy to merge the change to Tauri dependencies as they have updated their official guide in the same way.

ned-park:
> Personally,I think it should stay in place so the script will stop if something goes wrong so it is very clear to the user running the script that something went wrong.

My apologies for the length of this reply.  But mostly it comes down to set -e not actually making it clear to the user that something went wrong in a lot of cases.  Exiting doesn't mean an error happened a lot of the time, and set -e only exits, it doesn't provide any error messages or error handling (beyond exiting).  

It's likely we have different views about set -e, and since this is more your software than mine, using it or not is definitely your call. I view it as generally more trouble than it's worth and have spent more time than I'd like debugging/fixing scripts where someone had used it with good intentions and someone else later either didn't see it, didn't know what it did, or added code that ran afoul of one of its peculiarities.  It does have uses, so I can understand using it around particular parts of code where its functionality is desired and the error messages generated by the failed command are definitely going to be sufficient for the user to know what went wrong (probably the package manager stuff here would be safe), but I often argue that it shouldn't be set globally (or used for any new script) because it rarely works how people think it does and it has some peculiarities that many people aren't aware of.  I can understand why people like to use it, I've just been bitten by it enough times that I generally think it should be avoided.  I'd rather just handle the errors, print a message, and exit manually if something critical goes wrong in case someone decides to add a --quiet to something in 6 years and no one actually hits that particular failed case until a few years after that.  

I'd just note that within 5 or 6 lines of set -e was an if statement that printed a clear error message about why the script would have failed to finish that was unreachable because of set -e so the script bailed with what looked like success instead of failure and this user was none the wiser about it.  That kind of thing happens all the time.

One can always add in (error messages and) exit commands that should terminate the script if something critical fails or build a function that takes in some parameters (not shown here) if one wanted to make it more generically applicable to failed states:

```
PACKAGE_FAILURE() {
    echo "Failed to install the packages required for Spacedrive development"
    exit 1
}

sudo apt-get -y install ${SPACEDRIVE_CUSTOM_APT_FLAGS:-} $DEBIAN_TAURI_DEPS $DEBIAN_FFMPEG_DEPS $DEBIAN_BINDGEN_DEPS || PACKAGE_FAILURE
```

That way one can be sure that a user is notified of the failure before the script exits, whereas with set -e one can't be certain since the script can just bail without any error messages at all.

If, for instance, cargo/rust isn't installed the script exits before running the if statement that would tell the user that they're missing rust on their system.  `set -e` ensures the user gets no notification that script failed unless the command the script is running at the time happens to print an error instead of just exiting quietly (which isn't a safe assumption).  Instead of getting an error message one just gets dumped back to a regular prompt and have to manually `echo $?` if one wanted to see the success or failure of the script.  

```
~/spacedrive/.github/scripts$ ./setup-system.sh
Setting up your system for Spacedrive development!
~/spacedrive/.github/scripts$
```

When I looked at that I assumed the script finished and was successful.  It didn't print any error messages and I was back at a regular prompt without any manual intervention.  Without knowing the contents of the script, or manually checking exit status (rarely happens outside of development or a specific error message) I thought that was a successful run (forgot I'd uninstalled rust at some point).  For all I knew, all it needed to do was to make some directories, copy a couple config files, or ensure that I had a few things in my path and since they were there, there nothing else to do so script finished and exited.

Without set -e 

```
~/spacedrive/.github/scripts$ ./setup-system.sh
Setting up your system for Spacedrive development!
Rust was not detected on your system. Ensure the 'rustc' and 'cargo' binaries are in your $PATH.
~/spacedrive/.github/scripts$

```
The script handles the errors like it's supposed to.  I also know that since the if statement was in the script, the intention was that it would run and set -e was preventing it from doing so.  Admittedly, this issue could also be fixed by changing the if statements to this form (which would probably be better in any case since the current form is also bash anti-pattern):

```
if ! which cargoo &> /dev/null ; then
        echo "Rust was not detected on your system. Ensure the 'rustc' and 'cargo' binaries are in your \$PATH."
        exit 1
fi
```

But that was a relatively larger change at the time and even then if set -e is left in, the script assumes that the programs, commands and utilities being run will print error messages to the console (which isn't a given at all for many utilities running in a shell and shouldn't be relied on for error notification in scripts anyway since that behaviour can change over time, with different options, across different systems, and between different versions of things) to notify users that something went wrong instead of handling the errors in the script itself and notifying the user that something went wrong before exiting.  Since those two statements were a handful of lines apart and obviously conflicted with each other in terms of intention, I figured the set -e was left in accidentally when someone was debugging something because why would the error message code be deliberately made unreachable.

Keeping set -e for parts of the script is definitely an option if you really want it there, but I pretty much always prefer building in the error handling and exit manually so error messages get printed on failed commands and the exit statuses can be set to make debugging the script easier later on and avoid silent failures.

See [https://serverfault.com/a/847016](https://serverfault.com/a/847016) and its links for more on some of the possible pitfalls of set -e

oscartbeaumont:
Thank you for the thorough explanation. I am willing to take your guidance here given you seem to understand it much better than myself. Feel free to make whatever changes you think make sense and we can merge them.

I am fairly sure I added `set -e` because if the first `apt-get install` failed it would just continue executing the second one which would hide the error from the user but the code you propose above for handling the errors looks good as it will stop the program if the first command fails.

ned-park:
I've left set -e in, I was going to change it to turn it on and off as needed but it definitely seems excessive (I've mostly written done scripting on the administration side where stuff the things that break the script make up less of the contents.

I fixed the few if statements that break when `set -e` is on and wrote a small function that prints the line an error occurred on and that the script failed to finish.

oscartbeaumont (approved):

PR #370, opened 2022-08-30
Assignable Tags

jamiepine:
You can now add tags to files in the Explorer via the context menu.

This PR also has:
- Fix for the the window flash on startup bug
- Improves the look of the context menu
- Refactors Explorer state
- Some interface style changes
- Introduces `ObjectKind` ready for Objects PR
- Better starting point for docs
- 0.1.0 planned release notes


core/src/api/files.rs
```diff
@@ -19,6 +21,11 @@ pub struct SetFavoriteArgs {
 
 pub(crate) fn mount() -> RouterBuilder {
 	<RouterBuilder>::new()
+		// debug only
+		.query("object", |_ctx, _args: LibraryArgs<i32>| {
```
Brendonovich, line 25: what this

packages/client/src/stores/useExplorerStore.ts
```diff
@@ -1,31 +1,55 @@
...
+				if (!draft.multiSelectIndexes.length && indexes.length) {
+					draft.multiSelectIndexes = [draft.selectedRowIndex, ...indexes];
+				} else {
+					draft.multiSelectIndexes = [...new Set([...draft.multiSelectIndexes, ...indexes])];
+				}
+			})
+		);
+	},
+	reset: () => set(() => ({})),
+	set: (key, value) => set((state) => ({ ...state, [key]: value }))
```
Brendonovich, line 54: would be better accepting a `Partial<ExplorerStore>` and spreading it on the existing value

core/src/api/files.rs
jamiepine: something that shouldn't be here

core/src/object/mod.rs
```diff
@@ -0,0 +1,62 @@
...
+use serde::{Deserialize, Serialize};
+
+use crate::prisma;
+
+// The response to provide the Explorer when looking at Objects
+#[derive(Debug, Serialize, Deserialize, Type)]
+pub struct ObjectsForExplorer {
+	pub objects: Vec<ObjectData>,
+	// optional sources for context of Explorer
+	pub location: Option<prisma::location::Data>,
```
Brendonovich, line 15: Would be great if this was all one `sources` field that was of type `ObjectSource`. More descriptive + accurate + type safe + cool

oscartbeaumont:
@jamiepine Can we revert your changes to the mobile app (xcodeproj and Info.plist) before merging? I understand they help you run it on your end but the Xcode project should keep the `com.spacedrive.app` bundle identifier in the repository andf it should keep push notifications enabled.

core/src/api/tags.rs
```diff
@@ -80,10 +111,33 @@ pub(crate) fn mount() -> RouterBuilder {
 			|ctx, arg: LibraryArgs<TagAssignArgs>| async move {
 				let (args, library) = arg.get_library(&ctx).await?;
 
-				library.db.tag_on_file().create(
-					tag::id::equals(args.tag_id),
-					file::id::equals(args.file_id),
-					vec![],
+				if args.unassign {
```
Brendonovich, line 114: imo unassign should really be done in another query

jamiepine: you right, will be less painful when with the simplified version of invalidateQuery

core/src/api/locations.rs
```diff
@@ -15,16 +15,32 @@ use tracing::info;
...
+	Location(location::Data),
+	Tag(tag::Data),
+	// Space(object_in_space::Data),
+}
+
+#[derive(Serialize, Deserialize, Type, Debug)]
+#[serde(tag = "type")]
+pub enum ExplorerItem {
+	Path(Box<file_path::Data>),
+	Object(Box<file::Data>),
```
Brendonovich, line 35: Why box these?

core/src/api/tags.rs
```diff
@@ -80,10 +149,33 @@ pub(crate) fn mount() -> RouterBuilder {
...
```
Brendonovich, line 152: Yuck

core/src/api/locations.rs
jamiepine: it wouldn't let me commit otherwise, apparently the variable size of the two structs is too different

core/src/api/tags.rs
jamiepine: So I was thinking of renaming the query, since the amount of duplicated code otherwise is everything minus the prisma query, all the way to the frontend. Two useLibraryQueries just for add/remove feels more yuck to me.

xPolar (approved):
LGTM!

PR #375, opened 2022-09-04
Explorer performance

jamiepine:
- Replaced React Virtuso with Tanstack Virtual
- Refactored Explorer interface
- Replaced Zustand with Valtio (is great but has broken prod builds pls help)
- Refactored libraryStore and explorerStore to use Valtio
- Rebuilt context menu (Thanks Brendan :D)


oscartbeaumont:
I tried to fix one bug and ended up changing a lot, whoops.

Changes:
 - Valito bug fixed
 - Reduces unnecessary rerender and fetching data
 - Remove Tauri platform fetch as blocker for app startup -> This has halved the app startup time
 - Remove need for ‘import React’ in every file
 - We need to rely on React Query for state so that when its cache updates all of the UI updates - Using a Valito store causes issues here.
 - The `useCurrentLibrary` hook is where all library stuff goes through. This ensures the setup code for libraries is always run.
 - The sidebar renders basic version completly without any library queries or needing the library_id set in context.
 - Custom URI schema for loading thumbnails (ported from P2P branch)

Known bugs:
 - Can’t click in create library modal

oscartbeaumont (approved):

PR #376, opened 2022-09-05
Video thumbnails

fogodev:
Introducing video files thumbnail generation webp images using ffmpeg


apps/mobile/rust/src/android.rs
```diff
@@ -78,7 +78,7 @@ pub extern "system" fn Java_com_spacedrive_app_SDCore_handleCoreMsg(
 					env.get_string(data_dir.into()).unwrap().into()
 				};
 
-				let new_node = Node::new(data_dir).await;
+				let new_node = Node::new(data_dir).await.expect("Unable to create node");
```
oscartbeaumont, line 81: It's worth noting panicking in the mobile Rust code is probably a bad idea as I am fairly sure unwrapping across FFI can be unsound. This will be fine for now tho because the code could panic in any number of places internally.
Before we launch the mobile app it would be best for us all to go through and remove all panics and ensure we aren't introducing them in the future but for now this is fine.

oscartbeaumont (approved):

jamiepine:
CleanShot 2022-09-06 at 13 41 11@2x

I'm getting pretty bad thumbnails with my test files, also there's a casing issue on the query `.MP4` would be ignored.

Very excited about this PR though!

jamiepine:
Not sure if the placement of this code matches what Brendan is doing for the sync lib, but it feels out of place either way. maybe `core/crates/thumbnailer` would work better?

Brendonovich:
> Not sure if the placement of this code matches what Brendan is doing for the sync lib

Nah I'm doing the same thing

Plz squash merge this PR lol

jamiepine (approved):
incredible

PR #378, opened 2022-09-06
Use latest rspc + PCR

oscartbeaumont:
This PR utilises the latest versions of rspc and Prisma Client Rust, and introduces some new utilities to improve DX.

### `library_*` builder functions

The `library_query` and `library_mutation` builder functions for the router replace the `LocationArgs<T>` pattern that is currently in use. This changes nothing on the frontend, working as expected with `useLibraryQuery/Mutation`, but in Rust allows resolvers to consume a third argument that contains `library::Data` for library the request was scoped to, and regular arguments can be specified without being wrapped in `LocationArgs<T>`.

Due to Rust being annoying it is necessary to use async functions inside `library_*` resolvers, but this isn't a huge problem.

### Replacing usage of `with`/`fetch` with `include`

Using `include` and declaring shared types using `include!` macros can be a little annoying, but it allows for much better type-safety on the frontend. The model bindings exported by `rspc` now only contain scalar fields, so if a component needs to accept a relation it must specify it manually - eg. `Location & { node: Node }` which is equivalent to `location::include!({ node })` in Rust. In future we would like the model types to be aware of their relations but still require opt-in inclusion via a generic that accepts a string union of relations.

### More app-wide `invalidate_query`

`invalidate_query!` already supports not taking arguments, we just use it more now. Not exactly scalable but it's fine until we get normalised caching.

### `rspc` alphabetically orders exported types

Less annoying PR changelogs, yay. 

*Brendan*


Brendonovich (approved):

jamiepine (approved):

jamiepine:
I have no issue with the preposed changes, already had the run down from Brendan and it all makes sense.

Make sure this doesn't get squashed on merge, changed the setting myself.

oscartbeaumont:
To extend Brendan's message a `library_*` must take in a function pointer (not closure) which returns an async block that itself returns `Result<T, rspc::Error>`. I would like rspc to have official support for something like we are doing here, and if I can get that working it should be able to remove these limitations. If these limitations are a problem we could just "eject" to the old syntax.

The main downside of these limitations right now is it means to make a `todo` endpoint you must do `.library_*(|_, _: (), _| async { Ok(todo!()) })` which is pretty annoying but something we can live with.

PR #380, opened 2022-09-06
Only show a period at the end of a filename if it has an extension

### LINKED ISSUE #379: [ENG-240] Files with no extension still display a period as if there was one
### Describe the bug

If a file doesn't have an extension (e.g.: folders, gitignores), an extra period is displayed at the end of it's name in both the grid view and the inspector.

image

### Reproduction

1. Open a library location
2. Find a folder or file without an extension
3. In both the grid view & inspector you can see an extra period at the end of it

### Expected behavior

No period should be added to files or folders without an extension, and it should just be the name.

### Platform and versions

```Shell
pnpm 7.11.0
cargo 1.63.0 (fd9c4297c 2022-07-01)
rustc 1.63.0 (4b91a6ea7 2022-08-08)

running on macOS 12.5, but is visible across all platforms
```

### Stack trace

_No response_

### Additional context

_No response_

briamoe:
Removes the extra period at the end of a file or folder if it doesn't have an extension.

**before:**
image

**after:**
image

Closes #379

### DIFF
--- a/packages/interface/src/components/explorer/FileItem.tsx
+++ b/packages/interface/src/components/explorer/FileItem.tsx
@@ -65,7 +65,7 @@ export default function FileItem(props: Props) {
 					)}
 				>
 					{props.data?.name}
-					{props.data?.extension && <>.{props.data.extension}</>}
+					{props.data?.extension && `.${props.data.extension}`}
 				</span>
 			</div>
 		</div>
--- a/packages/interface/src/components/explorer/FileItem.tsx
+++ b/packages/interface/src/components/explorer/FileItem.tsx
@@ -64,7 +64,8 @@ export default function FileItem(props: Props) {
 						}
 					)}
 				>
-					{props.data?.name}.{props.data?.extension}
+					{props.data?.name}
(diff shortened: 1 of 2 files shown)


packages/interface/src/components/explorer/FileItem.tsx
```diff
@@ -64,7 +64,8 @@ export default function FileItem(props: Props) {
 						}
 					)}
 				>
-					{props.data?.name}.{props.data?.extension}
+					{props.data?.name}
+					{props.data?.extension && <>.{props.data.extension}</>}
```
Brendonovich, line 68: Could you use string interpolation instead of a fragment?
```
`.${props.data.extension}`
```

Brendonovich (changes requested):

packages/interface/src/components/explorer/FileItem.tsx
briamoe: Sure!

Brendonovich (approved):

oscartbeaumont (approved):

PR #381, opened 2022-09-08
Better db migrations

oscartbeaumont:
Move to using the Prisma migration engines for doing database migrations instead of our custom system. This should make them more reliable.

I also squashed the migrations because this change is already **NOT** backward compatible. Everyone will need to recreate their libraries.


Brendonovich (approved):

Brendonovich:
In future this should be managed by PCR, but this is fine for now as testing.

PR #382, opened 2022-09-08
[Eng 200] enter on keyboard doesn't cause modal to submit

### LINKED ISSUE #332: [ENG-200] Enter on keyboard doesn't cause modal to submit
For example in the modal for adding a new device (trigged from "Add device" on Overview page), when pressing the enter key inside the text field the modal does not get submitted.
<sub>Oscar Beaumont on Linear</sub>

MohammedBajuaifer:
This PR enables submission modals by using the "Enter Key". It also rerun the library queries when creating/removing libraries

Closes #332


oscartbeaumont:
I think we should get the `queryClient` from the hook. Eg. `const queryClient = useQueryClient();` instead of importing it directly.

xPolar:
@jamiepine @oscartbeaumont could this PR be rechecked.

MohammedBajuaifer:
Should I fix the conflict for this PR? there were no conflicts before but I think after merging the other PRs they appeared.

utkubakir (approved):

utkubakir:
This looks like a better approach to what we have on main as with this we can control if the form can be submitted or not (creating a library with no name) + looks cleaner in my opinion.

PR #383, opened 2022-09-14
Use native shortcuts

maxichrome:
This changes keyboard-shortcut functionality to properly use the Tauri (Rust-side) shortcuts engine. This allows, for instance, macOS' app-specific shortcut remapping — which is particularly important for power users — among other benefits (such as macOS' highlighting of shortcut position in the Menu Bar when pressed, which is absent in the current `main` build).

A macOS System Settings window. In the Keyboard Shortcuts section, the user has set a custom shortcut for the Search action; changing it to shift-command-L. Implicit knowledge is that the default shortcut for Search is command-L without a shift command.


oscartbeaumont:
Does using this system still allows for keyboard shortcuts on the web?

Also, these changes will definitely conflict with the explorer performance 2 branches. It's probs best to get that branch merged first and then I can fix the conflicts here when they come up.

maxichrome:
Yeah I've specifically kept in functionality for web shortcuts. Tried to test to no avail though; web now just shows a "No library loaded" message and I've not checked with anyone to see what's up

oscartbeaumont:
For web, you must run `cargo run -p server` in one console and in another console `pnpm web dev`.

Brendonovich (approved):
Love this, would just like to see search being defaulted to `/` instead of `Cmd+L`.
Linear use `/` for search and `Cmd+K` for command palette - `Cmd+L` conflicts with Open Location on a lot of browsers. 
image

maxichrome:
Hard agree Brendan, I remember the same debate being had when this was being originally implemented. If we're going to make this product change we should decide once and for all in this PR.

Consulting @jamiepine's approval/veto on changing to `/` here.

maxichrome:
Ok I went ahead and updated in this PR because slash does seem like much better of a shortcut. I'll give this 24 hours to get a veto from Jamie, and if so I'll revert that change; otherwise we can merge then

jamiepine:
Just remove the shortcut indicator in the search bar if you're going with "/" it'll look weird and be confusing. Otherwise this PR looks good. Hopefully no major conflicts with eng-explorer-performance-2

jamiepine:
Actually I recall preferring CMD+F for search, given that's what other file explorers use.

maxichrome:
Cool, will make that update. I had the same thought after finishing work last night. Then will be ready

PR #391, opened 2022-10-01
upgrade rspc + lazy loading in React

oscartbeaumont:
Changes:
 - [x] Upgrade to new rspc syntax
 - [x] Lazy loading for icons and pages in React. This should have been another PR but too late now.

Outstanding:
 - [x] Fix mobile
 - [x] Fix production/explorer issues from main


Brendonovich:
Looks like this is almost ready, but thumbnails for images aren't generating on my machine. @jamiepine @oscartbeaumont Can you check this?

Brendonovich (approved):

oscartbeaumont:
Huh, my database connection timed out when I added a location and after resetting my DB and restarting the app I am getting `Expected orphan Paths not returned from database query for this chunk` when starting an "identify unique files" job.

I think the thumbnailer is working fine it just has no files to work on or something like that because it finishes in less than a second and doesn't report an error. The thumbnailer dir is being created but is empty.

I have dug into the code but I am super confused. Why do we fail when there are no orphan files found [here](https://github.com/spacedriveapp/spacedrive/blob/5ab432ae464fa5d12f8c572c75359092c2abd1d5/core/src/object/identifier_job.rs#L150) @jamiepine?

Brendonovich (dismissed):

oscartbeaumont:
Turns out the job system is also broken in main. I am gonna merge this PR but it still needs to be investigated.

PR #392, opened 2022-10-04
move prisma schema and migrations to /core

Brendonovich:
This is how things were before moving everything into `/crates`.
IMO having `prisma-cli` in `/crates` makes sense since, well, it's a crate.
Having the schema there doesn't make sense since the schema is like an integral part of the core.


oscartbeaumont (dismissed):

jamiepine (approved):
I had no idea the cli crate could be separated from the schema and this makes me very happy. was annoyed to have to put the schema in the crates folder but was silently coping.

PR #396, opened 2022-10-04
Rename Files to Objects

jamiepine:
This has been a long time coming, a lot of code in the app currently refers to Files as Objects, but since the db schema required a large refactor I was avoiding it.

Also in this PR:
- Fix for a huge bug with invalidate query that was causing all reactive app updates to not work. Now they work!
- some code cleanup removing usage of `eslint-disable-next-line react-hooks/exhaustive-deps` and refactoring component logic accordingly, thanks Brendan!


jamiepine:
@utkubakir you might want to verify all is working in the mobile app with this before we merge...

jamiepine:
Merging anyway cause I need this for next PR + main is very broken anyway. Mobile + landing are not tested but should be fine.

PR #397, opened 2022-10-04
Object Kind

jamiepine:
## Changes
- Added `ObjectKind` struct 
- Added `Extension` enums with variants for some supported* file types with variant macros for easy definition
- Thumbnailer now uses `Extension` enum to query for image and video filetypes 
## Unrelated
- Improved explorer rendering
- Added dev data folder to separate prod and development libraries

*these filetypes are ones we understand, but do not necessarily support preview or open. custom icons can be made though.


core/src/util/db.rs
```diff
@@ -36,7 +39,15 @@ pub async fn load_and_migrate(db_url: &str) -> Result<PrismaClient, MigrationErr
 	}
 
 	#[cfg(not(debug_assertions))]
-	client._migrate_deploy().await?;
+	{
+		if std::env::var("SD_BASELINE_DB")
```
Brendonovich, line 43: This can be removed now that we have separate dev and prod dbs

core/src/util/db.rs
```diff
@@ -11,6 +11,9 @@ pub enum MigrationError {
 	#[error("An error occurred during migration: {0}")]
 	MigrateFailed(#[from] DbPushError),
 	#[cfg(not(debug_assertions))]
+	#[error("An error occurred during base-lining: {0}")]
+	ResolveFailed(#[from] MigrateResolveError),
```
Brendonovich, line 15: This can also be removed

Brendonovich (changes requested):

Brendonovich (approved):

PR #400, opened 2022-10-05
[ENG-250] Crypto Library

brxken128:
This implements the `crypto` library, the specifications of which can be found on Linear.

It does need further documentation writing, and tests, but I think it's good to go for now.

This should be ready to use within Spacedrive, and has been tested externally (it works very well).


crates/crypto/Cargo.toml
```diff
@@ -0,0 +1,29 @@
+[package]
+name = "sd_crypto"
```
Brendonovich, line 2:
```suggestion
name = "sd-crypto"
```

crates/crypto/src/header/keyslot.rs
```diff
@@ -0,0 +1,98 @@
...
+
+/// This defines the keyslot version
+pub enum KeyslotVersion {
+	V1,
+}
+
+impl Keyslot {
+	/// This function is used to serialize a keyslot into bytes
+	#[must_use]
+	pub fn serialize(&self) -> Vec<u8> {
```
Brendonovich, line 31: Would be nice if ser/de was done with `serde` instead of manually. Not sure how nicely that could be done using a binary format though.

crates/crypto/src/keys/hashing.rs
```diff
@@ -0,0 +1,67 @@
...
+	let argon2 = Argon2::new(
+		argon2::Algorithm::Argon2id,
+		argon2::Version::V0x13,
+		params.get_argon2_params(),
+	);
+
+	let result = argon2.hash_password_into(password.expose_secret(), &salt, &mut key);
+
+	// Manual drop so we can ensure that it's gone
+	drop(password);
```
Brendonovich, line 60: is this really necessary? `password` will always be dropped at the end of the function anyway.

brxken128:
Temporarily converting to a draft while I make some last-minute API changes.

Brendonovich (changes requested):
Will chuck in my 2c, can't comment on any of the subscription specifics though since idk what's going on lol

Brendonovich (approved):

brxken128:
I have some performance benchmarks:

```
2.03968GB/s, XChaCha20-Poly1305
1.47887GB/s, AES-256-GCM
```

I think this can only be improved upstream. We could stop `zeroize`ing the buffers, but then we'd possibly leak data (and the time it takes to zero out memory is negligible).

Most drives aren't able to read/write this fast, so performance for users will likely just be limited to the speed of the drive (+ the time it takes to hash their password).

jamiepine (dismissed):
just incredible work

Issue #402, opened 2022-10-06
[ENG-253] Spacedrive landing page has horizontal scroll on ios and mac


kor-pixel:
### Describe the bug

The spacedrive.app landing page has horizonal scroll issues on ios mobile and mac.

### Reproduction

1. Go to spacedrive.app
2. Shrink the page
3. scroll to the right

### Expected behavior

page wont scroll to the right

### Platform and versions

```Shell
M1 Macbook Pro
Deployed Site
```

PR #403, opened 2022-10-07
Fixed horizontal scrolling overflow

kor-pixel:
Fixed the horizontal scrolling issue present on IOS and Mac by adding overflow-hidden to landing page.

I do not have an Iphone to test that this does not cause any unintended issues so some further testing should be done before accepting this. 

Closes #(issue): [ENG-253] Spacedrive landing page has horizontal scroll on ios and mac #402


kor-pixel:
This is causing another issue, will fix it later

PR #404, opened 2022-10-07
fixed the overflow

kor-pixel:
Fixed the horizontal scrolling issue present on IOS and Mac by adding overflow-hidden to all pages by adding it to main tag

I do not have an Iphone to test that this does not cause any unintended issues so some further testing should be done before accepting this.

Closes #(issue): [ENG-253] Spacedrive landing page has horizontal scroll on ios and mac #402

### DIFF
--- a/apps/landing/src/pages/index.page.tsx
+++ b/apps/landing/src/pages/index.page.tsx
@@ -82,7 +82,7 @@ function Page() {
 	}, []);
 
 	return (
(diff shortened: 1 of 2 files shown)


maxichrome:
@oscartbeaumont @Brendonovich any idea why Clippy fails here? is this known?

Brendonovich:
@maxichrome Don't worry about it - Looks like the p2p stuff isn't compiling properly and Clippy isn't happy bc this is a fork, not a branch.

Issue #369, opened 2022-08-30
[ENG-222] libavresample-dev fails to install


Mamun-Developer:
### Describe the bug

When we execute `./.github/scripts/setup-system.sh` the follwing error comes up

> Package libavresample-dev is not available, but is referred to by another package.
> This may mean that the package is missing, has been obsoleted, or
> is only available from another source
> 
> E: Package 'libavresample-dev' has no installation candidate
> An error occurred while performing the task on line 78
> Setup for Spacedrive development failed

### Reproduction

1. Execute `./.github/scripts/setup-system.sh`

### Expected behavior

Must be successful installation.

### Platform and versions

```Shell
7.9.5
cargo 1.63.0 (fd9c4297c 2022-07-01)
rustc 1.63.0 (4b91a6ea7 2022-08-08)

OS: Ubuntu 22.04
Processor: AMD® Ryzen 7 5800h with radeon graphics × 16
Memory: 16GB
```

### Stack trace

```Shell
mamun@mamun-A35S:~/spacedrive$ ./.github/scripts/setup-system.sh
Setting up your system for Spacedrive development!
Detected 'apt' based distro!
Hit:1 https://apt.releases.hashicorp.com jammy InRelease
Hit:2 https://download.docker.com/linux/ubuntu jammy InRelease                                                                                                                       
Hit:3 https://dl.google.com/linux/chrome/deb stable InRelease                                                                                                                        
Hit:4 http://packages.microsoft.com/repos/code stable InRelease                                                                                                                      
Hit:5 http://security.ubuntu.com/ubuntu jammy-security InRelease                                                                                                                     
Hit:6 https://deb.nodesource.com/node_16.x jammy InRelease                                                                                                               
Hit:7 http://bd.archive.ubuntu.com/ubuntu jammy InRelease                                                                              
Hit:8 https://ppa.launchpadcontent.net/deadsnakes/ppa/ubuntu jammy InRelease                                    
Hit:9 http://bd.archive.ubuntu.com/ubuntu jammy-updates InRelease                 
Hit:10 http://bd.archive.ubuntu.com/ubuntu jammy-backports InRelease              
Hit:11 https://ppa.launchpadcontent.net/git-core/ppa/ubuntu jammy InRelease       
Reading package lists... Done                              
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
Package libavresample-dev is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
[...]

Issue #67, opened 2022-04-29
Package 'libavresample-dev' has no installation candidate


sommerper:
**Describe the bug**
On PopOS 21.10.
Trying to run setup shell script after cloning repository.

**To Reproduce**
1. git clone https://github.com/spacedriveapp/spacedrive
2. cd spacedrive
3. curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
4. curl -fsSL https://get.pnpm.io/install.sh | PNPM_VERSION=7.0.0-rc.9 sh - 
5. chmod +x ./.github/scripts/setup-system.sh && ./.github/scripts/setup-system.sh 
Results in output:
```
Setting up your system for Spacedrive development!
Detected 'apt' based distro!
[sudo] password for asdasdasd: 
Hit:1 https://download.docker.com/linux/ubuntu impish InRelease
Hit:2 http://ppa.launchpad.net/mosquitto-dev/mosquitto-ppa/ubuntu impish InRelease                       
Hit:3 http://us.archive.ubuntu.com/ubuntu impish InRelease                                               
Get:4 http://us.archive.ubuntu.com/ubuntu impish-security InRelease [110 kB]                
Hit:5 http://apt.pop-os.org/proprietary impish InRelease                                          
Get:6 http://apt.pop-os.org/release impish InRelease [6,850 B]                                    
Get:7 http://us.archive.ubuntu.com/ubuntu impish-updates InRelease [115 kB]                              
Get:8 http://apt.pop-os.org/release impish/main Sources [66.3 kB]                                 
Get:9 http://us.archive.ubuntu.com/ubuntu impish-backports InRelease [101 kB]                   
Hit:10 https://fman.io/updates/ubuntu stable InRelease                                  
Get:11 http://us.archive.ubuntu.com/ubuntu impish-security/main amd64 DEP-11 Metadata [20.3 kB]
Get:12 http://us.archive.ubuntu.com/ubuntu impish-security/universe amd64 DEP-11 Metadata [3,624 B]
Get:13 http://us.archive.ubuntu.com/ubuntu impish-updates/main i386 Packages [179 kB]
Get:14 http://us.archive.ubuntu.com/ubuntu impish-updates/main amd64 Packages [384 kB]             
Get:15 http://us.archive.ubuntu.com/ubuntu impish-updates/main amd64 DEP-11 Metadata [25.7 kB]
[...]

PR #416, opened 2022-10-16
rolled back 07/07/2022 PopOS fix and instead removed libavresample-de…

PyRo1121:
…v as that is deprecated as of ffmpeg >V 4.3.4

 This PR I have rolled back the commit on 07/07/2022 which made the install script work with Pop. In this case I removed libavresample-dev as a dep as it was deprecated after ffmpeg version 4.3.4. This is a common issue for all Ubuntu based builds as well as any one using ffmpeg version greater than 4.3.4

https://github.com/spacedriveapp/spacedrive/issues/67
https://github.com/spacedriveapp/spacedrive/issues/369

Closes #(issue): 2


mmattbtw (changes requested):
remove the `.idea` folder :^)

oscartbeaumont:
This commit was rebased onto main and had removed some unrelated stuff so I did a coauthored commit to main to merge the mains.
Thanks for the contribution.

PR #417, opened 2022-10-17
put mobile in pnpm workspace

Brendonovich:
Using `node-linker=hoisted` which removes pnpm's symlink functionality, `@sd/mobile` can now be included in the monorepo. Code and asset sharing seems to work, with svgs being displayed properly and `@sd/client` importing properly.

Turborepo has not been configured yet, but will be in future.

Huge thanks to @byCedric and [expo-monorepo-example](https://github.com/byCedric/expo-monorepo-example) for showing how this is possible.


apps/mobile/metro.config.js
```diff
@@ -6,32 +6,42 @@ const [SDAssetsPath, SDAssetsPathExclude] = resolveUniqueModule('@sd/assets', '.
...
 const { getDefaultConfig } = require('expo/metro-config');
 const expoDefaultConfig = getDefaultConfig(__dirname);
 
+const projectRoot = __dirname;
+const workspaceRoot = path.resolve(projectRoot, '../..');
+
 const metroConfig = makeMetroConfig({
-	projectRoot: __dirname,
+	projectRoot,
+	watchFolders: [workspaceRoot],
```
byCedric, line 22: If this turns out to be slow, since Metro will "index" all files in the repository, you can customize this a bit more. The only rule here is that:
> "anything outside these watch folders, won't be found by Metro. Even when they exist." (don't ask me why though)

With that, you can minimize Metro crawling through the Rust source code. You notice this slowness when starting Metro, that's when the crawling happens (during `npx expo start` for example)

utkubakir (approved):
Looks good, it builds and the modules work too. @sd/client is currently not used on mobile but I can do the cleanup when I merge it with my current branch. I had some package upgrades there too so all good.

oscartbeaumont:
I asked Brendan not to refactor anything on mobile to use the new client because I have some changes incoming on that front. Let me take care of that.

apps/mobile/metro.config.js
Brendonovich: Thanks for the feedback Cedric! We'll keep this in mind. In my testing it wasn't slow but it may be in the future.

PR #418, opened 2022-10-17
UI refactor

Brendonovich:
Introduces using class-variance-authority for various UI components.

- [x] Button
- [ ] Dropdown
- [ ] Select (need to discuss difference with Dropdown)

Everything could do with further refactors which will probably be done.


jamiepine:
Beautiful, merging early

PR #421, opened 2022-10-18
Mobile CI

oscartbeaumont:
Add a CI process for running a Typescript typecheck + ESLint and then an `eas build` in CI.

This PR currently only does Android builds because we need to have an Apple Developer account to build IOS. This can be easily added at a later date.

### DIFF
--- a/.easignore
+++ b/.easignore
@@ -1,0 +1,1 @@
+!/core/src/prisma.rs
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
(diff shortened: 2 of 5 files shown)


oscartbeaumont:
This is ready to be merged. It is currently disgustingly slow due to Android building the core for many different architectures ones after the other. I can't see any easy way to parallelise them with the current setup. We could investigate something custom if it becomes a problem. I am hoping the CI cache will make normal runs of the workflow a bit faster than I am observing but we have already gone over the GH Actions storage limit so we will have to wait and see.

@Brendonovich @utkubakir would be great to get a review from you both before merging.

utkubakir:
We had some PRs broke mobile before, so I am happy to see this. Great job as always!

maxichrome (approved):
I would rather use @utkubakir's approval on this, but peeking through this myself it looks solid.

PR #422, opened 2022-10-19
Identifier processes ObjectKind

jamiepine:
- Identifier now uses the internal `file-ext` crate to derive `ObjectKind` when creating an Object from scan
- Improvements to blur effects in the Explorer (thanks @maxichrome)


jamiepine:
Merged in with #431

PR #423, opened 2022-10-20
[ENG-258] Key Manager and Crypto Crate Improvements

brxken128:
This PR adds the entire `KeyManager`, which is an in-memory keystore for both mounted and stored keys.

It handles the mounting, unmounting, creating, retrieving, etc. of keys and will allow for extremely easy integration with other jobs/library mutations. It has been made accessible everywhere as it's included within the library context.

The Prisma schema has also been updated to accommodate these changes.

This PR also brings in a fix for metadata/preview media header objects. They used to be encrypted solely with the first key that was used to encrypt the file - now they inherit the master key from the keyslots. This, in conjunction with the key manager, grants us instant access to all encrypted data within a file (including metadata and preview media).


core/src/library/library_manager.rs
```diff
@@ -260,13 +261,18 @@ impl LibraryManager {
 			.exec()
 			.await?;
 
+		// create a new key manager
+		let key_manager = Arc::new(Mutex::new(KeyManager::new(Vec::<StoredKey>::new(), None)));
```
Brendonovich, line 265: 2 things here:
- Do the arc + mutex when you assign the value to `LibraryContext`, that way you can modify the key manager freely before handing it off
- Use `vec![]` to initialise the vec, it should infer the generic

core/src/util/seeder.rs
```diff
@@ -48,3 +53,42 @@ pub async fn indexer_rules_seeder(client: &PrismaClient) -> Result<(), SeederErr
 
 	Ok(())
 }
+
+pub async fn keystore_seeder(client: &PrismaClient, key_manager: Arc<Mutex<KeyManager>>) -> Result<(), SeederError> {
```
Brendonovich, line 57: I'm not sure this belongs in the `seeder` module. `seeder` is meant to populate data if none exists, but this is populating data from other existing data.
It might be better suited as a `create_key_manager` function in `library_manager.rs` that takes `&PrismaClient` and returns `KeyManager`.

Brendonovich (changes requested):
no idea if the crypto is good but the code looks good :)

crates/crypto/src/header/keyslot.rs
```diff
@@ -64,10 +62,10 @@ impl Keyslot {
...
 		algorithm: Algorithm,
 		hashing_algorithm: HashingAlgorithm,
+		salt: [u8; SALT_LEN],
 		password: Protected<Vec<u8>>,
 		master_key: &Protected<[u8; MASTER_KEY_LEN]>,
 	) -> Result<Self, Error> {
-		let salt = generate_salt();
 		let nonce = generate_nonce(algorithm);
 
 		let hashed_password = hashing_algorithm.hash(password, salt).unwrap();
```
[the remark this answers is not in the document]
brxken128: oh, not a clue how i missed that. thanks 😭

jamiepine (approved):

PR #428, opened 2022-10-21
remove openssl from core

Brendonovich:
shot in the dark but this may fix windows ci

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5182,7 +5182,6 @@ dependencies = [
  "itertools",
  "normi",
  "once_cell",
- "openssl-sys",
  "prisma-client-rust",
  "rmp",
  "rmp-serde",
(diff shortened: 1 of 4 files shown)


brxken128:
could windows CI be setup with the build script that @maxichrome has been working on? it might take too long to runs, or might just not work in a non-rebootable environment though (not too sure on the specifics here)

Brendonovich:
@brxken128 That will be used eventually but atm I just want it to pass.

brxken128:
> @brxken128 That will be used eventually but atm I just want it to pass.

that's fair 🙏

oscartbeaumont:
The warning does seem to be gone and I have no idea why that would be given `openssl-sys` is still being override at a workspace level. Can merge this now and if the warnings come back we can review at that point.

oscartbeaumont (approved):

PR #427, opened 2022-10-21
Refactor dropdown

Brendonovich:
new tw function, more cva use, more organised code

PR #431, opened 2022-10-22
New color system & better UI components

jamiepine:
- Light and dark themes now work and look decent
- UI elements are cleaner and more consistent  
- Began using `cva` and `tw` for cleaner Tailwind

This PR also ate two other PRs (#427 & #422)

spacedrive-ui

Also got this cool black on white style for menus in light theme
CleanShot 2022-10-22 at 04 35 59@2x


apps/desktop/src-tauri/src/main.rs
```diff
@@ -59,9 +59,9 @@ async fn main() -> Result<(), Box<dyn Error>> {
 
 			#[cfg(target_os = "macos")]
 			{
-				use macos::{lock_app_theme, AppThemeType};
+				// use macos::{lock_app_theme, AppThemeType};
 
-				lock_app_theme(AppThemeType::Dark as _);
+				// lock_app_theme(AppThemeType::Dark as _);
 			}
```
maxichrome, line 65: Should we remove this entirely?

jamiepine: probably, can't tell what it would be useful for now we have working themes

apps/desktop/src/index.tsx
```diff
@@ -12,7 +12,7 @@ import '@sd/ui/style';
 
 const isDev = import.meta.env.DEV;
 const client = hooks.createClient({
-	links: [...(isDev ? [loggerLink()] : []), tauriLink()]
+	links: [...(false ? [loggerLink()] : []), tauriLink()]
```
Brendonovich, line 15: should this be undone?

docs/developers/architecture/sync.md
```diff
@@ -4,85 +4,15 @@ index: 12
 
 # Sync
 
-Spacedrive synchronizes data using a combination of master-slave replication and last-write-wins CRDTs,
-with the synchronization method encoded into the Prisma schema using [record type attributes](#record-types).
+Spacedrive synchronizes library data in realtime across the distributed network of Nodes.
```
Brendonovich, line 7: This all needs to be undone. Guessing it's a wrongly resolved merge?

packages/interface/src/AppRouter.tsx
```diff
@@ -7,38 +7,38 @@ import { AppLayout } from './AppLayout';
...
-const DebugScreen = loadable(() => import('./screens/Debug'));
-const SettingsScreen = loadable(() => import('./screens/settings/Settings'));
-const TagExplorer = loadable(() => import('./screens/TagExplorer'));
-const PhotosScreen = loadable(() => import('./screens/Photos'));
-const OverviewScreen = loadable(() => import('./screens/Overview'));
-const ContentScreen = loadable(() => import('./screens/Content'));
-const LocationExplorer = loadable(() => import('./screens/LocationExplorer'));
-const OnboardingScreen = loadable(() => import('./components/onboarding/Onboarding'));
-const NotFound = loadable(() => import('./NotFound'));
+const DebugScreen = lazy(() => import('./screens/Debug'));
```
Brendonovich, line 10: Undo this, bad merge.

packages/interface/src/components/primitive/InputContainer.tsx
```diff
@@ -9,18 +9,15 @@ interface InputContainerProps extends DefaultProps<HTMLDivElement> {
...
 
-export function InputContainer(props: PropsWithChildren<InputContainerProps>) {
+export function InputContainer({ mini, ...props }: PropsWithChildren<InputContainerProps>) {
 	return (
 		<div className="flex flex-row">
-			<div
-				className={clsx('flex flex-col w-full', !props.mini && 'pb-6', props.className)}
-				{...props}
-			>
+			<div className={clsx('flex flex-col w-full', !mini && 'pb-6', props.className)} {...props}>
```
Brendonovich, line 15: Spread props before setting `className`, otherwise `className` will be overwritten.

packages/ui/src/Button.tsx
```diff
@@ -34,47 +31,31 @@ const styles = cva(
 			loading: {
 				true: 'opacity-70'
 			},
-			padding: {
-				thin: '!p-1',
-				sm: '!p-1.5'
-			},
-			noBorder: {
-				true: 'border-0'
+			forIcon: {
```
Brendonovich, line 34: Not sure I agree with this. Variants shouldn't really overlap, and `forIcon` could definitely overlap with some other intent that needs different padding. I'm not a fan of `loading` and `noBorder` either. Maybe I'm nitpicking but I'd prefer `padding: { icon: '!px-1' }`.

Brendonovich (changes requested):
Seems like main wasn't merged in well and some things need to be undone. Would be great if @oscartbeaumont could check to make sure.

Also CI isn't passing yet so fix that plz ❤️

packages/interface/src/AppRouter.tsx
jamiepine: No this is an actual fix, loadable doesn't work with suspense.

Brendonovich: Ohh ffs I can't read, nevermind

oscartbeaumont: Odd. We originally had `lazy` but it was the cause of hot reload breaking. I will give this a look.

Brendonovich: > Odd. We originally had `lazy` but it was the cause of hot reload breaking. I will give this a look.

Didn't we use `React.lazy` in the past? This is using `lazy` from `@loadable/components`, which I overlook initially.

apps/desktop/src/index.tsx
jamiepine: yes, I disabled it cause log spam were making debugging hard

packages/interface/src/components/primitive/InputContainer.tsx
jamiepine: true, will fix

packages/ui/src/Button.tsx
jamiepine: Even though I haven't needed to adjust the padding on any icon button yet, I see why it is bad. Will switch to `padding="icon"`, or make an `IconButton` component that takes care of overriding the padding.

packages/interface/src/AppRouter.tsx
oscartbeaumont: Oh that should be alright. I'm on mobile so it didn't show the import.

Brendonovich (dismissed):

oscartbeaumont (approved):

PR #432, opened 2022-10-22
fix(interface/settings): remove `library` dep when updating general settings

### LINKED ISSUE #429: Library name overriden when dropdown is active and you navigate through libraries
### Describe the bug

This is currently only tested on web. 
If you select a library, navigate to library settings (General) and open the dropdown and select another library the next library will have its name overriden by the previous.

### Reproduction

1. Select a library and navigate to library settings (General)
2. Open the dropdown
3. Select another library
4. You should still be in library settings, but the name of the new library you clicked is now the same as the previous

### Expected behavior

Open the dropdown, select a new library then the dropdown should be closed and no library name overwrite.

### Platform and versions

```Shell
pnpm: 7.13.0
cargo: cargo 1.64.0 (387270bc7 2022-09-16)
rustc: rustc 1.64.0 (a55dd71d5 2022-09-19)

OS: Ubuntu 
Browser: Brave
```

### Stack trace

_No response_

### Additional context

I'll look into it and hopefully propose a fix for it.

itschip:
Fixes #429 where the library name would be overidden. Caused by the side effect responsible for checking for changes in the debounced values.

Side note: I was considering adding functionality to close dropdown on click/clickaway, but decided to wait for your input.

### DIFF
--- a/packages/interface/src/screens/settings/library/LibraryGeneralSettings.tsx
+++ b/packages/interface/src/screens/settings/library/LibraryGeneralSettings.tsx
@@ -35,7 +35,7 @@ export default function LibraryGeneralSettings() {
 				});
 			}
 		}
-	}, [nameDebounced, descriptionDebounced, library, editLibrary]);
+	}, [nameDebounced, descriptionDebounced, editLibrary]);
 
 	useEffect(() => {
 		if (library) {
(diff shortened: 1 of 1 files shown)


oscartbeaumont:
Thanks for this PR as it has been a known issue and I have been meaning to fix it. The entire component makes me sad with all the `useEffect`'s  so I pushed a change refactoring it to use `react-hook-form` which cleans up the code and prevents bugs like this from making it back in.

oscartbeaumont (dismissed):

Brendonovich (approved):

jamiepine:
Awesome PR, thanks for fixing this it was bugging me on stream yesterday attempting to debug it.

jamiepine (approved):

PR #435, opened 2022-10-24
fix(interface/settings): update general values on library change

### LINKED ISSUE #434: General settings values are not updated on library changes.
### Describe the bug

It seems that ad28ed41905a68bd42d6f2c5eaf12e521c9a914a only fixed one aspect of the general settings. This time, the actual values are not updated, so even tho it doesn't _automatically_ override the name, you can do so yourself by either changing the name or description, since the values stays the same throughout renders.

### Reproduction

1. Open dropdown
2. Choose a library and navigate to general settings
3. Choose another library (you should still be in general settings and see the previous values)
4. Update one of the input fields
5. Current library should have the same values as the previous.

### Expected behavior

Update general settings values on navigation from library to library

### Platform and versions

```Shell
pnpm: 7.13.0
cargo: 1.64.0 (387270bc7 2022-09-16)
rustc: 1.64.0 (a55dd71d5 2022-09-19)
```

### Stack trace

_No response_

### Additional context

I'll look into and create a PR

itschip:
Fixes #434

I also took the liberty to create a useDebounce hook 👀

Hopefully, this should fix both previous and current issues with general settings.

### DIFF
--- a/packages/interface/src/hooks/useDebounce.ts
+++ b/packages/interface/src/hooks/useDebounce.ts
@@ -1,0 +1,16 @@
+import { useEffect, useState } from 'react';
+
+export const useDebounce = <T = any>(value: T, delay = 100): T => {
+	const [debouncedValue, setDebouncedValue] = useState(value);
+
+	useEffect(() => {
+		const handler = setTimeout(() => {
+			setDebouncedValue(value);
+		}, delay);
+		return () => {
+			clearTimeout(handler);
+		};
+	}, [value, delay]);
+
+	return debouncedValue;
+};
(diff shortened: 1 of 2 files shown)


oscartbeaumont:
Thanks for your help but this was fixed by @jamiepine and I yesterday on his branch #431 [here](https://github.com/spacedriveapp/spacedrive/blob/130443d689fb3e1e8d06afcb2d4632174a2dbf34/packages/interface/src/screens/settings/library/LibraryGeneralSettings.tsx). It was an oversight on my part when moving to `react-hook-form`.

That `useDebounce` hook looks pretty nice!

PR #440, opened 2022-10-27
Update Team Page

jamiepine:
- Update members
- New animojis for those without face pics
- Bonus: Less blue app UI


apps/landing/src/components/TeamMember.tsx
```diff
@@ -17,6 +17,7 @@ export interface TeamMemberProps {
 		twitter?: string;
 		twitch?: string;
 		github?: string;
+		dribbble?: string;
```
Brendonovich, line 20: 5head

packages/ui/style/colors.scss
```diff
@@ -22,7 +22,7 @@
     --color-sidebar-selected: var(--dark-hue), 15%, 24%;
     --color-sidebar-shade: var(--dark-hue), 15%, 23%;
     // main
-    --color-app: var(--dark-hue), 19%, 14%;
+    --color-app: var(--dark-hue), 15%, 14%;
```
Brendonovich, line 25: Still seems quite blue. I'd be fine with a separate theme that's a lot more gray like the original colours.

xPolar (approved):

PR #442, opened 2022-10-27
Update file-types.json

HeavySnowJakarta:
Adds some file types (mainly about Microsoft Office and Hancom Office) to file-types.json.

Note: There may exist problems about the change of the expression of “Open-XML Office Type”. 

This pull request is to response the Discord message: https://discord.com/channels/949090953497567312/960383223421075507/1031420903449444372


jamiepine:
So we're actually planning on deprecating this file in favour of the `file_ext` crate which can identify files by their magic bytes, though it is easy to add supported formats, but we haven't yet implemented a system for descriptions. Though I think I have a plan for that which would support i18n.

PR #444, opened 2022-10-28
Error Bounary Sentry support

oscartbeaumont:
Currently, this PR hardcodes our Sentry DSN. Vite is refusing to load environment variables from the `.env` file on my machine and I can't figure out why. This will be fine for now unless something else wants to have a look into it.


packages/interface/src/ErrorFallback.tsx
```diff
@@ -1,13 +1,25 @@
...
+		console.log('TODO', error);
+		captureException(error);
+		// platform.openLink(
+		// 	`https://github.com/spacedriveapp/spacedrive/issues/new?assignees=&labels=kind%2Fbug%2Cstatus%2Fneeds-triage&template=bug_report.yml&logs=${encodeURIComponent(
+		// 		error.toString()
+		// 	)}&info=${encodeURIComponent(
+		// 		`App version ${version} running on ${guessOperatingSystem() || 'unknown'}`
+		// 	)}`
+		// );
+		resetErrorBoundary();
```
maxichrome, line 20: Where is this defined? I don't see it across the diffs here but could have just missed it

oscartbeaumont: If you referring to `resetErrorBoundary` it is a prop to the component that comes from the error boundary. Idk how the error boundary was set up as that was done previously. This function in practice just resets the app back to the overview page.

jamiepine (dismissed):

PR #447, opened 2022-10-30
Added open suse

PyRo1121:
I updated the DIstro identifier as the one in place was having issues with openSUSE. Also had to update FFMPEG to ^4.4.0 due to openSUSE not having on main PPA. Added in node check to setup as a requirement needed for all builds except mobile. I am working towards automating the checks next so instead of failing we would be able to install rust/pnpm/node through script.

Closes #(issue)


oscartbeaumont:
I'm not sure downgrading ffmpeg because a single Linux distribution doesn't have it in their main PPA is a good move. The changes to the script look pretty good but they have macOS downloading ffmpeg `5.0.1` while you have downgraded the Rust package which I assume would cause issues or at minimum inconsistencies.

PyRo1121:
> I'm not sure downgrading ffmpeg because a single Linux distribution doesn't have it in their main PPA is a good move. The changes to the script look pretty good but they have macOS downloading ffmpeg `5.0.1` while you have downgraded the Rust package which I assume would cause issues or at minimum inconsistencies.

The downgrade is actually just a revert back to the original ffmpeg required for macOS. I will look for alternative routes so we can maintain the required ffmpeg all other distros are using.

PyRo1121:
I have updated it now so it is running off the latest ffmpeg by using their experimental repo. I also put in the script to remove the repo after running to clean up the install

PyRo1121:
On this request I would say because of lack of Tauri bundling I could remove the openSUSE. With that I still think the remaining portion is beneficial for identifying the distro correctly instead of by a package manager

Brendonovich:
Seeing as a lot has changed with this PR, I think the existing changes should be contributed separately.
Make an issue where we can discuss what changes should be made, then the changes can be attempted again.
Usually we'd auto-close PRs that aren't linked to an issue, so it'd be good for the next one to follow that rule.

PyRo1121:
I agree and I will follow the correct route as set forth by the rules. I have learned a ton since opening this one and :| it is a cluster.

PR #450, opened 2022-10-31
[ENG-262] Key Manager Integration

brxken128:
This PR is massive, please criticize it to shreds - it's important that we get this right.

The Key Manager now uses [DashMap](https://github.com/xacrimon/dashmap) in order to provide concurrency and better performance over `HashMap`.

There are a few known bugs/issues:

* The dropdown UI is pretty bad, and doesn't fit in too well with the rest of Spacedrive (and the TS formatting is bad too, I don't know how we format that though)
* The key numbering is volatile, and can/will be confusing
* We have no way to set key names within the UI
* We have no way to view the plaintext key within the UI
* Object/container stats are not attached (this isn't major, as they would all be 0 and therefore wouldn't show anyway)
* The UI offers no differentiation between the default key and normal keys
* There is a bug where you are unable to remove a key instantly while mounting another key
* We have no way to set the master password, and possibly no way to verify that it's correct until the user attempts to run an action. I will be adding a master password validation feature, which will essentially insert a ghost key into the library, which is only ever used to ensure the entered master password is correct. If we're going to give the user a master password during library onboarding, we'll need a way to make sure it's correct and doesn't deviate from what was provided.

Go ahead and close this if needs be, there is still quite a lot to do - I just didn't want this branch to be sat around forever.

Here is a preview of the current key manager (key mounting on dev builds isn't this fast, but they will be in prod):

https://user-images.githubusercontent.com/77554505/199038410-0d21c5c5-4ec6-48e3-aad8-2e139978c22f.mp4

### DIFF
--- a/core/prisma/schema.prisma
+++ b/core/prisma/schema.prisma
@@ -196,7 +196,7 @@ model Key {
   // is this key the default for encryption?
   // was not tagged as unique as i'm not too sure if PCR will handle it
   // can always be tagged as unique, the keys API will need updating to use `find_unique()`
-  default      Boolean
+  default      Boolean @default(false)
   // nullable if concealed for security
   date_created DateTime? @default(now())
   // encryption algorithm used to encrypt the key
@@ -216,7 +216,7 @@ model Key {
   // the *encrypted* key
   key Bytes
 
-  automount Boolean
+  automount Boolean @default(false)
 
   objects    Object[]
   file_paths FilePath[]
--- a/core/prisma/schema.prisma
+++ b/core/prisma/schema.prisma
@@ -194,6 +194,8 @@ model Key {
   // the name that the user sets
   name         String?
   // is this key the default for encryption?
+  // was not tagged as unique as i'm not too sure if PCR will handle it
+  // can always be tagged as unique, the keys API will need updating to use `find_unique()`
   default      Boolean
   // nullable if concealed for security
   date_created DateTime? @default(now())
--- a/core/prisma/schema.prisma
+++ b/core/prisma/schema.prisma
@@ -192,7 +192,7 @@ model Key {
   // uuid to identify the key
   uuid         String    @unique
   // the name that the user sets
-  name         String
+  name         String?
   // is this key the default for encryption?
   default      Boolean
   // nullable if concealed for security
--- a/core/prisma/schema.prisma
+++ b/core/prisma/schema.prisma
@@ -192,7 +192,7 @@ model Key {
   // uuid to identify the key
   uuid         String    @unique
   // the name that the user sets
-  name         String?
+  name         String
   // is this key the default for encryption?
   default      Boolean
   // nullable if concealed for security
--- a/core/prisma/schema.prisma
+++ b/core/prisma/schema.prisma
@@ -214,6 +214,8 @@ model Key {
   // the *encrypted* key
   key Bytes
 
+  automount Boolean
+
   objects    Object[]
   file_paths FilePath[]
 
--- a/core/src/api/keys.rs
+++ b/core/src/api/keys.rs
@@ -38,12 +38,7 @@ pub(crate) fn mount() -> RouterBuilder {
 		})
 		.library_mutation("mount", |t| {
 			t(|_, key_uuid: uuid::Uuid, library| async move {
-				library.key_manager.mount(key_uuid).map_err(|_| {
-					rspc::Error::new(
-						rspc::ErrorCode::InternalServerError,
-						"Error while mounting the key within the key manager".into(),
-					)
-				})?;
+				library.key_manager.mount(key_uuid).map_err(rspc::Error)?;
 				// we also need to dispatch jobs that automatically decrypt preview media and metadata here
 				invalidate_query!(library, "keys.listMounted");
 				Ok(())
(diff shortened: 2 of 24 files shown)


jamiepine:
CleanShot 2022-10-31 at 23 29 03@2x
I'm getting a core crash when adding a key

brxken128:
> CleanShot 2022-10-31 at 23 29 03@2x
> I'm getting a core crash when adding a key

That's pretty odd, I'll take a look into it

brxken128:
Converted to draft for now, there's quite a few unwraps that I'm not happy with

core/src/api/keys.rs
```diff
@@ -38,7 +38,12 @@ pub(crate) fn mount() -> RouterBuilder {
 		})
 		.library_mutation("mount", |t| {
 			t(|_, key_uuid: uuid::Uuid, library| async move {
-				library.key_manager.mount(key_uuid).unwrap();
+				library.key_manager.mount(key_uuid).map_err(|_| {
```
Brendonovich, line 41: It would be much better for these `Error` -> `rspc::Error` implementations to be done in an `Into<rspc::Error>` implementation. Repeating this code everywhere is not great haha.

brxken128: I did ask Oscar about this as I wasn't too confident or happy with it, but yeah I'll fix that. Thanks!

brxken128:
Going to close this for now, after some thorough testing there's a few more bugs that I'm not happy with.

brxken128:
Re-opening this as I believe the most critical issues/bugs have been solved. I am going to be adding a dedicated master password verification feature, but I think that needs a separate PR.

core/prisma/schema.prisma
```diff
@@ -214,6 +216,8 @@ model Key {
   // the *encrypted* key
   key Bytes
 
+  automount Boolean
```
Brendonovich, line 219: Before this PR is merged a new migration must be generated

core/src/api/keys.rs
```diff
@@ -0,0 +1,247 @@
...
+			t(|_, _: (), library| async move {
+				library.key_manager.empty_keymount();
+				invalidate_query!(library, "keys.listMounted");
+				Ok(())
+			})
+		})
+		// this also mounts the key
+		.library_mutation("add", |t| {
+			t(|_, args: KeyAddArgs, library| async move {
+				let algorithm = match &args.algorithm as &str {
```
Brendonovich, line 196: This and `hashing_algorithm` should be enums

packages/interface/src/components/key/Key.tsx
```diff
@@ -17,11 +17,82 @@ export interface Key {
...
+
+	return (
+		<DropdownMenu.Root open={open} onOpenChange={setOpen}>
+		<DropdownMenu.Trigger>
+			{trigger}
+		</DropdownMenu.Trigger>
+		{transitions(
+				(styles, show) =>
+					show && (
+		<DropdownMenu.Portal forceMount>
```
Brendonovich, line 64: formatting

packages/interface/src/components/key/Key.tsx
```diff
@@ -17,11 +17,82 @@ export interface Key {
...
+			</DropdownMenu.Content>
+		</DropdownMenu.Portal>
+		))}
+		</DropdownMenu.Root>
+	);
+};
+
+
 export const Key: React.FC<{ data: Key; index: number }> = ({ data, index }) => {
+	const { mutate: mountKey } = useLibraryMutation('keys.mount');
```
Brendonovich, line 91: Don't destructure mutations, just use `.mutate` without aliasing

packages/interface/src/components/key/KeyList.tsx
```diff
@@ -1,47 +1,61 @@
...
 
 export type KeyListProps = DefaultProps;
 
+const ListKeys = () => {
+	const keys = useLibraryQuery(['keys.list']);
+	const mounted_uuids = useLibraryQuery(['keys.listMounted']);
+	const default_key = useLibraryQuery(['keys.getDefault']);
+
+	const mountedKeys: QueryKey[] = keys.data?.filter((key) => mounted_uuids.data?.includes(key.uuid)) ?? []
+	const unmountedKeys: QueryKey[] = keys.data?.filter(key => !mounted_uuids.data?.includes(key.uuid)) ?? []
```
Brendonovich, line 16: Might be worth putting these in `useMemo`

packages/interface/src/components/key/KeyMounter.tsx
```diff
@@ -9,13 +10,20 @@ const KeyHeading = tw(CategoryHeading)`mb-1`;
...
+	const mounted_uuids = useLibraryQuery(['keys.listMounted']);
+
 	const [showKey, setShowKey] = useState(false);
 	const [toggle, setToggle] = useState(true);
 
 	const [key, setKey] = useState('');
 	const [encryptionAlgo, setEncryptionAlgo] = useState('XChaCha20Poly1305');
 	const [hashingAlgo, setHashingAlgo] = useState('Argon2id');
 
+	const { mutate: createKey } = useLibraryMutation('keys.add');
```
Brendonovich, line 26: Said it earlier but don't destructure mutations. Might try get it in an eslint rule.

Brendonovich (changes requested):

core/src/api/keys.rs
brxken128: I tried to do this, but `value=` within the `KeyMounter` is required to be a string AFAIK and I just couldn't get it to work.

core/prisma/schema.prisma
brxken128: Should be done in ed3e4ca58aadb7fb2e130a5cd0bf35d1499a00f6.

core/src/api/keys.rs
brxken128: Got it. It's not the cleanest, nor is it perfect, but it's better than giving Rust a String.

jamiepine (approved):

packages/interface/src/components/key/KeyMounter.tsx
utkubakir: Any reason for not destructuring? I guess it looks cleaner if the component has several mutations.

Brendonovich (dismissed):

Brendonovich:
Please rustfmt and should be good to go

jamiepine (approved):

PR #458, opened 2022-11-03
Enforce minumum Node version

### LINKED ISSUE #457: NPMRC
### Describe the bug

So I was looking through the `package.json` in the root and I see the app requires Node `>= 14.0.0`. If this is not put in the `.npmrc` as `engine-strict=true` that rule is not applying across the project. A perfect example is when I run `pnpm install` I get the error provided below (under **Stack trace**) when adding `engine-strict=true`

Please also reference the PNPM docs: [package.json](https://pnpm.io/package_json) goes over the specifics of using engine requirements. I feel this would be better suited in `.npmrc` so we can avoid any packages being added that are not compatible with the required Node version. We may want to update the `package.json` to be consistent with `CONTRIBUTING.md` which references a requirement for Node 17.

### Reproduction

1. Update `.npmrc` to include
   
   ```
   engine-strict=true
   ```
1. Run `pnpm install`

### Expected behavior

Should install, as all packages in theory should match the project's requirements set forth in `CONTRIBUTING.md` and `package.json`

### Platform and versions

```Shell
7.14.1
cargo 1.67.0-nightly (7e484fc1a 2022-10-27)
rustc 1.67.0-nightly (11ebe6512 2022-11-01)
```

### Stack trace

When running with invalid Node version:

```
cloud spacedrive [npmrc] pnpm install
Scope: all 13 workspace projects
Lockfile is up to date, resolution step is skipped
 ERR_PNPM_UNSUPPORTED_ENGINE  Unsupported environment (bad pnpm and/or Node.js version)

Your Node version is incompatible with "/amqplib/0.5.2".

Expected version: >=0.8 <=9
Got: v14.19.2
```

PyRo1121:
Updated .NPMRC to include `engine-strict=true` as per the pnpm docs listed in issue by putting the node version only in package.json it does not carry they requirement across the project. This also would allow future contributions to include dependencies that don't fall in line with the projects path and requirements.

Closes #457


.npmrc
```diff
@@ -1,3 +1,5 @@
+; make all engine requirements (e.g. node version) strictly kept
+engine-strict=true
```
maxichrome, line 2: Good addition!

package.json
```diff
@@ -39,9 +39,9 @@
 		"@types/node": "^16.0.0"
 	},
 	"engines": {
-		"pnpm": ">=6.0.0",
+		"pnpm": ">=7.0.0",
```
maxichrome, line 42: We might even consider bumping this to contain a minimum minor version, just in case bug fixes/additions were made which we utilize and don't realize are missing in older minor versions (unlikely but possible).

```suggestion
		"pnpm": ">=7.14.0",
```

package.json
```diff
@@ -39,9 +39,9 @@
...
 		"npm": "pnpm",
 		"yarn": "pnpm",
-		"node": ">=14.0.0"
+		"node": ">=17.0.0"
```
maxichrome, line 45: We might consider:

- adding a minimum Node version here
- adding a `.nvmrc` file so users using _nvm_ or _fnm_ can take advantage of those tools' easy Node version switching

maxichrome (dismissed):
Thanks for the suggestion & the patch. A few notes in my review but overall I approve. Thanks PyRo!

package.json
PyRo1121: I thought about the .nvmrc as well but wasn't sure if y'all wanted to add more. I think it would be really beneficial as user myself of nvm

PyRo1121: I will say with adding `.nvmrc` I don't believe there is a way to set the version to be >= so I thought the best solution if we add that would be LTS/*

PyRo1121:
Adding to this and I apologize I took your advice and didn't mean to dismiss it @maxichrome. We might want to update the `@type/node` on all 'package.json` as it is the type definitions for node.

.nvmrc
```diff
@@ -0,0 +1 @@
+lts/*
```
Brendonovich, line 1: The [contributing guide](https://github.com/spacedriveapp/spacedrive/blob/main/CONTRIBUTING.md) recommends Node v17, this should probably be enforced here.
```suggestion
v17
```

Brendonovich (changes requested):

Brendonovich:
Turns out Vercel only support Node 16, guess we're using 16 instead of 17 then.

maxichrome (approved):
All looks good to me now

PR #466, opened 2022-11-08
[ENG-283] Key manager refinements

brxken128:
This PR refines the key manager pretty heavily, and I'm a lot more comfortable calling it stable now. The UI for these changes isn't perfect.

Main changes:

* Key manager is locked until a valid password+secret key combo are provided. This is currently "password" and `MDAwMDAwMDAwMDAwMDAwMA==` (16 zeroes encoded in base64). This will be changed once we have proper onboarding.
* Nothing plaintext is stored in memory. The master password is now hashed on entry, and that is stored in memory. We have instant access to all of the keys, and only need to hash each plaintext key with the content salt now.
* We can verify the master password and secret key combination are correct (with the use of a nil-UUID "verification key")
* Key mounting time was reduced by half (due to hashing the master password on entry)
* An "Unmount all & lock" button was added, for quickly removing everything from memory.
* The "sync with library" button is functional, and if disabled the key will only be stored in memory.
* The crypto crate got `serde` feature gating
* The key settings page was populated, and it now functions just the same as the key manager itself
* Passphrase generation! Current passphrases are 7-words long, using the EFF large wordlist (the index has been removed so we can easily select words), and they're separated with `-`. This gives each passphrase 7776^7 possible combinations (which is an insane amount!), but we can always swap this wordlist out if needed.

Below is a preview video of these changes. This is in real-time, with no cuts (in prod):

https://user-images.githubusercontent.com/77554505/200555466-d8eb258b-a61b-4adb-bf11-af7bec99e873.mp4

This also closes ENG-305.


brxken128:
Converting to draft until tomorrow morning, the settings page needs to be updated to follow the new "lock" system.

PR #467, opened 2022-11-21
[ENG-307] Key manager features

brxken128:
This PR consumed #466. It closes ENG-283, ENG-305 and ENG-307. Consequently, it is quite large.

The changes for this PR:
* Master password changing functionality
* Keystore backups
* Keystore backup restoration
* Massive UI overhauls (including a very nice password strength meter)

The notes imported from #466:
* Key manager is locked until a valid password+secret key combo are provided. This is currently "password" and `MDAwMDAwMDAwMDAwMDAwMA==` (16 zeroes encoded in base64). This will be changed once we have proper onboarding.
* Nothing plaintext is stored in memory. The master password is now hashed on entry, and that is stored in memory. We have instant access to all of the keys, and only need to hash each plaintext key with the content salt now.
* We can verify the master password and secret key combination are correct (with the use of a nil-UUID "verification key")
* Key mounting time was reduced by half (due to hashing the master password on entry)
* An "Unmount all & lock" button was added, for quickly removing everything from memory.
* The "sync with library" button is functional, and if disabled the key will only be stored in memory.
* The crypto crate got `serde` feature gating
* The key settings page was populated, and it now functions just the same as the key manager itself
* Passphrase generation! Current passphrases are 7-words long, using the EFF large wordlist (the index has been removed so we can easily select words), and they're separated with `-`. This gives each passphrase 7776^7 possible combinations (which is an insane amount!), but we can always swap this wordlist out if needed.


Brendonovich:
For future reference, don't make PRs this big. They should only close a single Linear issue and even then the previous PRs were pretty large. It's fine for now but plz not again.

oscartbeaumont (approved):

crates/crypto/assets/eff_large_wordlist.txt
```diff
@@ -0,0 +1,7776 @@
+abacus
```
oscartbeaumont, line 1: Is there a reason we are doing this instead of just letting a library do this for us? I have been using [bip39](https://crates.io/crates/bip39) for P2P stuff and it's a bit quirky but works.

oscartbeaumont (approved):

oscartbeaumont:
Would be great to get this merged once your happy @brxken128. It all looks good to me.

A squash merge is probably in order.

crates/crypto/assets/eff_large_wordlist.txt
brxken128: As far as I'm aware, `bip39` only has 2048 total words/choices (in the English list), as opposed to 7776 with the EFF large wordlist. The considerably higher permutations of the EFF wordlist make it more suitable for an application such as this, and the entropy of diceware-esque passwords is well discussed and calculated (a good example would be [here](https://www.privacyguides.org/basics/passwords-overview/#diceware-passphrases)).

It's worth noting that BIP39 is great for wallet/other mnemonics, as you only need the first 4 letters of the word to have the total word (as no word starts with the first same 4 letters) - recovery is a whole lot easier. I do believe BIP39 also has support for other languages, which is something the current wordlist falls short with, and we'll probably need to fix this in the future. For this application though, the total entropy/permutations is pretty critical.

Some math:
BIP39 total permutations: 2048^7 ~= `1.511157275×10²³`
EFF wordlist total permutations: 7776^7 ~= `1.7190708×10²⁷`

Combine that with `argon2id` (+moderate parameters), and an attacker would have a pretty challenging task ahead of them.

PR #468, opened 2022-11-23
ENG 223 Location Awareness

fogodev:
Introducing location awareness feature

Now when Spacedrive is running, we watch each location for file system events, like file creation, updates or removals, acting accordingly to each event. And we also store some metadata on a hidden file at each location root path, checking when a location is online or offline and relinking a pre-existing location that was missing and re-added.

Some other fixes, like a small revamp on job system typings and fixes on file-ext subcrate.

I'm setting up this PR to have access to CI tests, as I included some tests about file system events that I wish to run on multiple platforms. If they run just fine, this PR is ready to be reviewed, otherwise I will do some fixing.


jamiepine (changes requested):
Not working correctly on macOS, discussion on Slack

apps/server/package.json
```diff
@@ -4,6 +4,7 @@
 	"main": "index.js",
 	"license": "GPL-3.0-only",
 	"scripts": {
-		"dev": "cargo watch -x 'run -p server'"
+		"dev": "RUST_LOG=\"sd_core=info\" cargo watch -x 'run -p server'",
+		"build": "DATA_DIR=/Users/jamie/Projects/spacedrive/apps/server/sdserver_data cargo run -p server --release"
```
brxken128, line 8: this is specific to @jamiepine's system and should probably be removed before merge

core/src/location/manager/watcher/linux.rs
```diff
@@ -0,0 +1,66 @@
...
+		Self {}
+	}
+
+	async fn handle_event(
+		&mut self,
+		location: indexer_job_location::Data,
+		library_ctx: &LibraryContext,
+		event: Event,
+	) -> Result<(), LocationManagerError> {
+		debug!("Received Linux event: {:#?}", event);
```
brxken128, line 33: i strongly think these (windows/macos/linux) should be removed before merge - adding a location that constantly updates (even the spacedrive library directory) will cause the terminal window to be flooded with tens of thousands of debug logs

brxken128 (commented):
Great work, this will make filesystem actions a lot smoother! The code looks pretty great, but there's a few bugs that I've encounted:

On Fedora 37:
* Creating a new directory and then deleting it from a watched location does not remove the directory from within Spacedrive (an event is caught though).
* Sometimes locations cannot be added. My test directory is `Documents/Kicad`, and it gave me an error relating to the location metadata
* Using the location manager and removing a location will produce a mass amount of errors relating to the file watcher, cluttering the terminal window (see attached image)
* The amount of logs in general are massive - this should be fairly trivial to disable though

Overall, great work! The UI is cleaner, things feel more responsive and I'm quite excited to see this merged.

image

core/src/location/manager/watcher/linux.rs
fogodev: I agree, changing these to `trace` logs

apps/server/package.json
fogodev: Removed

oscartbeaumont:
On macOS rename and file create events are working but file deletion is not. I have also found a few conditions where changes are not causing the queries to be invalidated which I have fixed.

Below is the trace logs for when I delete a file.
```
2022-12-13T05:37:59.384416Z  INFO sd_core::location::manager::watcher::macos: Received MacOS event: Event {
    kind: Modify(
        Data(
            Any,
        ),
    ),
    paths: [
        "/Users/oscar/Pictures/assets",
    ],
    attr:tracker: None,
    attr:flag: None,
    attr:info: None,
    attr:source: None,
}
2022-12-13T05:37:59.385141Z  WARN sd_core::location::manager::watcher::macos: Unexpected MacOS modify event on a directory
2022-12-13T05:37:59.388000Z  INFO sd_core::location::manager::watcher::macos: Received MacOS event: Event {
    kind: Modify(
        Name(
            Any,
        ),
    ),
    paths: [
        "/Users/oscar/Pictures/assets/bruh.rtf", // This is the name of the file I deleted
    ],
    attr:tracker: None,
    attr:flag: None,
    attr:info: None,
    attr:source: None,
}
```
However, the file remains in the DB after this.

I tried to find a solution but I couldn't get it working. I can give it another look later in the week if @fogodev isn't sure what is causing it.

core/src/lib.rs
```diff
@@ -107,16 +112,39 @@ impl Node {
...
 			NodeContext {
 				config: Arc::clone(&config),
 				jobs: Arc::clone(&jobs),
+				location_manager: Arc::clone(&location_manager),
 				event_bus_tx: event_bus.0.clone(),
 			},
 		)
 		.await?;
 
+		// Adding already existing locations for location management
```
oscartbeaumont, line 127: If we support encrypted libraries we will want this in the library manager instead of here because encrypted libraries won't be loaded on startup (due to needing a password).
I have fixed this in #493 but just putting it as a note.

oscartbeaumont (approved):

oscartbeaumont:
The macOS Delete event is still not working but I discussed with @fogodev and it's probably best for us to merge this huge PR and then continue with fixes on a new smaller PR. I am working on an improved file system watcher for macOS and that can be part of a separate PR and should fix the delete event.

This PR has so many outstanding changes that I am creating branches off it instead of main which is a sign this PR has been open for too long!

@fogodev great work on this! It's really gonna make the Spacedrive app feels so much better!

PR #475, opened 2022-12-07
[ENG-251] Crypto Jobs

brxken128:
This PR adds encrypt and decrypt jobs to the context menu. A few things in this PR aren't final, and input surrounding both the functionality and the UX of these changes would be greatly appreciated. The main points are below:

* Currently, encrypt (by default) just takes the file and encrypts it to *another* file in the same local directory - it appends ".sdenc" to the name. The user can change this through the UI.
* Decrypt (by default) adds ".decrypted" to the end of the file name. I really dislike this functionality, but cannot decide what should happen instead.
* No file size (making sure there's enough space for the job) or overwrite checks are in place. The former could be implemented trivially, by checking Prisma's `volume` table. I may do this before it gets merged.

We are able to add additional UX-focused switches to the encrypt and decrypt dialogs (such as "encrypt in place" or "encrypt to separate file", that kind of thing).

Hitting encrypt or decrypt will open a specific dialog, with configurable options for the user. Once this has closed, an alert dialog will tell the user that their job has started and they can track the progress in the job manager. I'm don't mind this as it makes it clear to the user, but someone else's two cents would be very appreciated.


maxichrome (approved):
I think we should pass over this for UI & UX refinements but that can be a separate task. Looks nice overall, happy to have some user facing encryption mechanism now! ❤️

brxken128:
> I think we should pass over this for UI & UX refinements but that can be a separate task. Looks nice overall, happy to have some user facing encryption mechanism now! ❤️

Yeah, at least we have something functional - refinements aren't as large of a task at all. Appending ".decrypted" to the end of each decrypted file is most definitely **not** user friendly 🤣

PR #478, opened 2022-12-09
[ENG-313] Key auto-generation and viewing

brxken128:
This PR adds a key generator, and a key viewer. It also adds the default password/secret key to the onboarding screen. We're able to do this for now as they're static, but once proper onboarding has been developed the user will be provided with their (random!) details then.

It also makes some (pretty large) changes to the way I've been handling alert dialogs, and aims to make the whole thing a lot easier. I think this'll require @oscartbeaumont's approval, just so we can make sure none of our changes clash. If they do, I can just cherry-pick/pull out the key viewer and key generation changes specifically and we can drop the dialog work in here.

The generated keys are now done in JS with the help of `crypto-random-string`. This means everything is a lot more responsive, and we aren't bombarding rspc with queries each time the user adjusts the slider.

Key automounting is also functional, and I plan to add a dropdown menu item to enable/disable this per-key in another PR.

image showing a slider for generating keys in the key mounting screen
image showing the default password and secret key in the temporary onboarding screen


oscartbeaumont:
Looks great! I put a few comments in but this can be squash merged when your ready @brxken128

utkubakir (approved):

PR #484, opened 2022-12-13
[ENG-316] Decrypt with Password

brxken128:
This implements password decryption, and offers the user a chance to import the key and content salt into our key manager. This is super helpful for either one-off decryption, or importing lost key/content salt relations. This removes the check for any mounted keys before opening the decrypt dialog, but it disables keymanager-based decryption if no keys are mounted.

I may implement this for encryption too, just as a one-off "encrypt with this password" (this could also have the option to save the key to the key manager). I'm not too sure if this would be worthwhile though, as it's just as easy to create a memory-only key within the key manager.

I have tested this, and it successfully imports the key into the key manager.

This PR also adds the (hex encoded) content salt to the key viewer.

### DIFF
--- a/core/src/api/keys.rs
+++ b/core/src/api/keys.rs
@@ -301,6 +301,7 @@ pub(crate) fn mount() -> RouterBuilder {
 					args.hashing_algorithm,
 					!args.library_sync,
 					args.automount,
+					None,
 				)?;
 
 				let stored_key = library.key_manager.access_keystore(uuid)?;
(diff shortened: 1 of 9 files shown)


brxken128:
Temporarily converting this to a draft while I try to find the source of encrypt/decrypt jobs failing to even start in #487. This has been squash-merged in, so I'll test both and see what the issue could be.

oscartbeaumont (approved):

PR #485, opened 2022-12-13
Add `security.txt` and a public key to the landing page

brxken128:
This PR adds a public key to our landing page, as well as a `security.txt` file that was signed with the private key.

I wouldn't merge this until we get `security@spacedrive.com` set up.

Additionally, we should set up a GitHub/landing page security policy at some point (although those shouldn't stop this PR).

Please thoroughly check all URLs, make sure the signatures/keys are functioning correctly, etc. I have, but I may have missed something.


oscartbeaumont:
I set up the `security at spacedrive.com` alias in Google Admin. For future reference, the alias was set up by creating a "Group" and setting the permission matrix so that external users as "Who can post" to the group, then adding everyone into the group who should receive the emails.

oscartbeaumont (approved):

PR #486, opened 2022-12-14
Bump Node Version To LTS

utkubakir:
- Bumped node version to LTS (v18)
- Add eslint to `@sd/client`.


Brendonovich:
Will need to update Vercel to use node 18 but apart from that looks good

oscartbeaumont (approved):

oscartbeaumont:
Web deployment is still failing in the same way it has been a long time so we can just ignore it. Everything else looks good!

PR #487, opened 2022-12-14
[ENG-318] Implement a KDF to derive new keys

brxken128:
This PR implements BLAKE3's KDF in order to derive new encryption keys, so we're not re-using both the root key and user key hashes directly across multiple types of algorithms. This change not only allows us to follow correct practices, but it heavily minimizes the damage caused by an exposed master key somewhere in the chain.

It uses 3 context strings - one for root key related tasks, one for (user key + content salt) hashes, and one for the (master password + secret key) related hashes. For this to work, we did have to introduce an additional salt to both Prisma and the keyslots. Migrations have been re-generated.

This also refactors some of the `crypto` crate, most of the work being focused on the key manager. I apologise in advance for the amount of effort this may take to review, but things do still work from what I've tested. On the plus side, the code is a **lot** more idiomatic and easy to follow.

This squash-merged #484 in order to prevent merge conflicts, so I'll keep this as a draft until that is merged.


brxken128:
Keeping this as a draft for now - there's a bug somewhere with file encryption.

brxken128:
I cherry-picked the old squash merge out, and I believe this PR is good to go! Apologies for not getting around to it sooner.

crates/crypto/src/header/file.rs
```diff
@@ -344,149 +343,149 @@ impl FileHeader {
...
-			]
-			.to_vec(),
-			metadata: None,
-			preview_media: None,
-		};
-
-		header.serialize().unwrap();
-	}
-}
+// #[cfg(test)]
```
oscartbeaumont, line 346: Should these tests be reenabled before merging?

oscartbeaumont (dismissed):

crates/crypto/src/header/file.rs
brxken128: I need to go through and fix them before they're re-enabled. The tests use an older version of the header format (as the bytes are hardcoded), and new ones need to be generated. I was putting this off until I'm certain we won't have any more structural changes for V1 of the crypto code, so that we don't have to change the tests again.

oscartbeaumont:
@brxken128 feel free to squash merge once you're happy it looks good to me!

Brendonovich (approved):

PR #491, opened 2022-12-21
Integrate KeyManager with library creation flow

oscartbeaumont:
When creating a library you are now guided through setting up a password and secret for the keymanager.


crates/crypto/src/keys/keymanager.rs
```diff
@@ -151,17 +158,19 @@ impl KeyManager {
 	pub fn onboarding(
 		algorithm: Algorithm,
 		hashing_algorithm: HashingAlgorithm,
+		creds: PasswordAndSecret,
```
brxken128, line 161: the credentials need to be generated in here, as we provide the master password and secret key to begin with (annoying, but extremely secure). the user is free to change it afterwards - i plan to revise this eventually.

crates/crypto/src/keys/keymanager.rs
```diff
@@ -151,17 +158,19 @@ impl KeyManager {
...
 	) -> Result<OnboardingBundle> {
-		let _master_password = generate_passphrase();
-		let _salt = generate_salt();
-
 		// BRXKEN128: REMOVE THIS ONCE ONBOARDING HAS BEEN DONE
```
brxken128, line 163: we can un `_` the variables above, and remove the statically set password/salt beneath (and the length check) as the master password/salt are generated in this function

crates/crypto/src/keys/keymanager.rs
brxken128: we could keep it this way, i don't inherently have anything against it. maybe i should add a passphrase generation function to master password changing.

brxken128 (commented):
sorry for the delay with checking over it! things look great, i'll give it a build quickly and make sure everything is looking good there too. thank you :)

crates/crypto/src/keys/keymanager.rs
brxken128: ^ closing this as it's related to the other comment

packages/interface/src/components/dialog/CreateLibraryDialog.tsx
```diff
@@ -2,55 +2,104 @@ import { useBridgeMutation } from '@sd/client';
...
+					<p className="text-sm mt-3">Name:</p>
+					<Input
+						className="flex-grow w-full"
+						placeholder="My Cool Library"
+						disabled={form.formState.isSubmitting}
+						{...form.register('name', { required: true })}
+					/>
+				</div>
+
+				{/* TODO: Proper UI for this. Maybe checkbox for encrypted or not and then reveal these fields. Select encrypted by default. */}
```
brxken128, line 81: Library encryption itself isn't supported yet, but the key manager shouldn't be restricted to encrypted libraries only

packages/interface/src/components/dialog/CreateLibraryDialog.tsx
```diff
@@ -2,55 +2,104 @@ import { useBridgeMutation } from '@sd/client';
...
+				<span className="text-sm">Make password and secret fields empty to skip key setup.</span>
```
brxken128, line 82: Setup can't be skipped, the key manager *needs* a verification key created before it can ever be unlocked. There isn't support for adding one after the fact

packages/interface/src/components/dialog/CreateLibraryDialog.tsx
```diff
@@ -2,55 +2,104 @@ import { useBridgeMutation } from '@sd/client';
...
+					/>
+				</div>
+				<div className="relative flex flex-col">
+					<p className="text-sm mt-2">Secret Key:</p>
+					<Input
+						className="flex-grow !py-0.5"
+						{...form.register('encrypted_cfg.secret')}
+						disabled={form.formState.isSubmitting}
+						placeholder="30303030-30303030-30303030-30303030"
+					/>
```
brxken128, line 100: The secret key *can't* be generated by the user - this is what the onboarding function did. If we allow the user to arbitrarily set a salt, security will be heavily lowered. Once we move the secret key to a pepper, we can remove the need for a secret key at all for users who don't like the functionality.

packages/interface/src/components/dialog/CreateLibraryDialog.tsx
```diff
@@ -2,55 +2,104 @@ import { useBridgeMutation } from '@sd/client';
...
+				{/* TODO: Proper UI for this. Maybe checkbox for encrypted or not and then reveal these fields. Select encrypted by default. */}
+				<span className="text-sm">Make password and secret fields empty to skip key setup.</span>
+
+				<div className="relative flex flex-col">
+					<p className="text-sm mt-2">Password:</p>
+					<Input
+						className="flex-grow !py-0.5"
+						disabled={form.formState.isSubmitting}
+						{...form.register('encrypted_cfg.password')}
+						placeholder="password"
```
brxken128, line 90: I think the password here should be an autogenerated passphrase at first, so the user can either keep that or change it to one of their own that they like. That offers a nice balance between secure-by-default and not too annoying, imo. Maybe we could have a refresh icon to auto-generate a new passphrase too.

core/src/library/library_manager.rs
```diff
@@ -71,58 +71,19 @@ impl From<LibraryManagerError> for rspc::Error {
...
-				verification_key.salt.to_vec(),
-				vec![],
-			)
-			.exec()
-			.await?;
-	}
-
-	let db_stored_keys = client.key().find_many(vec![]).exec().await?;
-
 	let mut default = Uuid::nil();
```
brxken128, line 78: Not your fault, just an afterthought, but we can probably use an option for this. Not sure why I decided a nil-UUID would be suitable here.

core/src/api/keys.rs
```diff
@@ -190,31 +174,6 @@ pub(crate) fn mount() -> RouterBuilder {
...
-				write_storedkey_to_db(library.db.clone(), &verification_key).await?;
-
-				let keys = OnboardingKeys {
-					master_password: bundle.master_password.expose().clone(),
-					secret_key: base64::encode(bundle.secret_key.expose()),
-				};
-
-				Ok(keys)
-			})
-		})
```
brxken128, line 217: We'll need this to some extent. Secret keys should be generated within Rust, and should **not** have the opportunity to be changed at all through JS, at least not yet.

brxken128 (changes requested):

brxken128 (approved):
Awesome work! Things are functioning great, and I'll open a PR to address the changes we discussed on Slack (optional secret key, optional key manager initialisation, etc).

oscartbeaumont:
Thanks, for the review and your help with this PR @brxken128!

PR #493, opened 2022-12-27
Job system improvements

oscartbeaumont:
Changes:
 - [x] We now pass around `&mut WorkerContext` to avoid need for Mutex. I was able to move some stuff around so we only ever need one instance of `WorkerContext` which satisfies Rust's borrow checker rule of having a single mutable reference. This means no need for `lock().await` to update the job report.
 - [x] Remove `WorkerEvent` enum along with the associated channel and Tokio task. This channel was extra complexity along with another point of failure and was just not needed. For example [here](https://github.com/spacedriveapp/spacedrive/blob/bbca699e3397ed8588fa954eca76a3f670b9434c/core/src/job/worker.rs#L158) we send a message than use `done_tx` to await the response. There is no reason we shouldn't just do the [DB call and other logic](https://github.com/spacedriveapp/spacedrive/blob/bbca699e3397ed8588fa954eca76a3f670b9434c/core/src/job/worker.rs#L230) on the current thread given it's already awaiting that operation.
 - [x] ~~`CurrentDirFileIdentifierJob`, `FileEncryptorJob` and `FileDecryptorJob` were not put into the `JobManager`'s resume functionality. This means they would never be resumed.~~ (@brxken128 let me know this was intentional but it could well have been a mistake given that isn't explictly set by whoever made the job) 
      - [x] The longer-term fix for this is to implement the `JobRestored` trait and a constant called `const JOB_RESTORER: Lazy<HashMap<&'static str, Box<dyn JobRestorer>>>`. When you ingest a job if it doesn't exist in the `JOB_RESTORER` constant it will refuse to run. This will help us catch the mistake before it causes issues.
 - [x] Instead of resuming jobs on core startup we now resume them when each library is loaded within the library manager. Resuming them at startup is gonna cause problems when we have encrypted libraries that can't be loaded at startup (because they have a password). The same thing was done for the `LocationManager` as it's gonna have the same problems. For the `LocationManager` it would mean when creating a library it would not enabled the FS listener without restarting the app.
 - [x] Removed external job queuing (`LibraryCtx.queue_job(...)`, `JobManager.ingest_queue(...)`). These methods facilitate writing code that assumes we only have a single worker thread. This is currently the case but I don't think it will always be the case so writing code around it is just gonna cause us errors. Eg. We do `.queue_job(...).queue_job(...).spawn_job(...)` in one point which assumes the spawned job will run first then the two queued jobs. With multiple worker threads, they could all just start at the same time.
      - [ ] As of writing I haven't come up with a solution but just using `.spawn(...).spawn(...).spawn(...)` and reordering the jobs achieves the equivalent functionality of the previous code without needing the queue method which encourages the single-threaded assumptions. This can be dealt with in a future PR.
 - [x] Removed `DynJob` and replace it with a simple system that uses a tokio task for both queued and spawned jobs. Upon queuing a job it will spawn onto a tokio::task and await a channel that triggers the job to start. This system is so much easier to understand as the code doesn't do a heap of message passing. In terms of performance, if a web server can handle 10k reqs/seconds and spawns a new tokio task for each thread, we should be fine spawning a few extra tasks for the added simplicity. This also mostly removes the concept of a `Worker`.
 - [x] Simplify `spawn_job` call. From `library.spawn_job(Job::new(FileEncryptorJobInit { ... }, FileEncryptorJob {})).await` to `library.spawn_job(FileEncryptorJobInit { ... }).await`
 - [x] Reintroduce using `Hash` on `StatelessJob::Init` so that two identical jobs can't be run at the same time. This was a regression from my changes.
 - [ ] Make the job's time elapsed on the frontend work again.
     - The old system would push updates to the frontend every second pointlessly using up rspc IPC bandwidth. Instead the frontend will just keep updating the time elapsed from a static `time_started` field on the job. When it receives the update that the job has been paused or stopped it can stop ticking up the time elapsed and use the value in that payload. This is similar to what Vercel does and it means if the stop or pause events never reach the frontend it will keep counting forever but this is a fine tradeoff, imo.
 - [x] Pushing certain data directly to frontend's RQ cache to prevent the core from feeling lagging while under heavy load.
      - When indexing a large location a lot of `invalidate_query` events are sent over the rspc IPC bus. This causes it to become saturated and makes the app lag. The way an `invalidate_query` event works is as follows: "Frontend invalidate key x" -> "Backend what is the new data for key x" -> "Frontend here is the new data for key x". This can be improved.
      - [x] Batch and deduplicate invalidations on the backend. Ask the frontend to invalidate a bunch of keys at once so it can fetch all the new values at once with its built-in query batching.
      - [x] Pushing new data. Eg. "Frontend I have the following new data for key x.". This means 2 less IPC calls for the same functionality. This feature is significantly more error-prone but is worth it for performance-sensitive calls.
 - [ ] Thoroughly test everything

This was gonna include the prune preview media job but it's already huge and has been dragging out so let's get this PR merged as is and do that in a future PR.


oscartbeaumont:
Kinda replaced by #707.

PR #506, opened 2023-01-06
[ENG-323] Key manager onboarding fixes

brxken128:
This PR makes quite a few changes, and fixes the key manager in relation to onboarding.

Changes:

- The "Welcome to Spacedrive" screen had a minor overhaul, and looks a lot less out of place
- The secret key was moved to a pepper-based system
- As an extension to the above, the secret key is also now completely optional and user-settable
- Both the Rust and TS crypto-related code was heavily cleaned up
- `key` was added to `SelectOption`, and it's just set to the (string) `value`. This prevents a lot of log errors.
- `HashingAlgorithm` -> `string` and vice versa conversions were fixed, and are a lot neater.

image displaying the new "welcome to spacedrive" screen with a blue-purple gradient background

image showing the new library creation dialog, with key manager configuration options


crates/crypto/src/keys/keymanager.rs
```diff
@@ -357,9 +340,17 @@ impl KeyManager {
...
+			old_verification_key.hashing_algorithm.hash_with_secret(
+				master_password,
+				old_verification_key.content_salt,
+				sk,
+			)?
+		} else {
+			old_verification_key
+				.hashing_algorithm
+				.hash(master_password, old_verification_key.content_salt)?
+		};
```
Brendonovich, line 353: can this be put in a function? looks like it's done in a couple of places

packages/interface/src/components/dialog/BackupRestoreDialog.tsx
```diff
@@ -19,20 +20,23 @@ export interface BackupRestorationDialogProps {
...
+	const form = useForm<FormValues>({
 		defaultValues: {
 			masterPassword: '',
-			secretKey: ''
+			secretKey: '',
+			filePath: ''
 		}
 	});
 
 	const onSubmit: SubmitHandler<FormValues> = (data) => {
```
Brendonovich, line 31: use `form.handleSubmit` here
```suggestion
	const onSubmit = form.handleSubmit((data) => {
```
would mean u can do `onSubmit={onSubmit}` in `<form>` and u don't have to manually type it

packages/interface/src/components/dialog/CreateLibraryDialog.tsx
```diff
@@ -1,25 +1,42 @@
-import { useBridgeMutation } from '@sd/client';
-import { Input } from '@sd/ui';
-import { Dialog } from '@sd/ui';
+import { Algorithm, useBridgeMutation } from '@sd/client';
+import { Button, Dialog, Input, Select, SelectOption } from '@sd/ui';
 import { useQueryClient } from '@tanstack/react-query';
-import { PropsWithChildren } from 'react';
+import { writeText } from '@tauri-apps/api/clipboard';
```
Brendonovich, line 4: this a big no no, try using `navigator.clipboard.writeText` which isn't bound to Tauri. if it doesn't work then we can add a clipboard section to `Platform` which uses this API outside of `@sd/interface`

packages/interface/src/components/dialog/MasterPasswordChangeDialog.tsx
```diff
@@ -1,34 +1,78 @@
-import { useLibraryMutation } from '@sd/client';
+import { Algorithm, useLibraryMutation } from '@sd/client';
 import { Button, Dialog, Input, Select, SelectOption } from '@sd/ui';
-import { zxcvbn, zxcvbnOptions } from '@zxcvbn-ts/core';
-import zxcvbnCommonPackage from '@zxcvbn-ts/language-common';
-import zxcvbnEnPackage from '@zxcvbn-ts/language-en';
-import clsx from 'clsx';
-import { Eye, EyeSlash } from 'phosphor-react';
+import { writeText } from '@tauri-apps/api/clipboard';
```
Brendonovich, line 3: refer to earlier comment

packages/interface/src/components/dialog/MasterPasswordChangeDialog.tsx
```diff
@@ -1,34 +1,78 @@
...
 
-	const onSubmit: SubmitHandler<FormValues> = (data) => {
+	const [show, setShow] = useState({
+		masterPasswordDialog: false,
+		masterPassword: false,
+		masterPassword2: false,
+		secretKey: false
+	});
+
+	const MP1CurrentEyeIcon = show.masterPassword ? EyeSlash : Eye;
```
Brendonovich, line 71: making a `PasswordInput` component for this would be nice

packages/interface/src/components/explorer/Explorer.tsx
```diff
@@ -93,7 +90,9 @@ export default function Explorer(props: Props) {
 			</div>
 			<AlertDialog
 				open={alertDialogData.open}
```
Brendonovich, line 92: if `AlertData` has a `GenericAlertDialogState`, can it just consume the entire state as `state` and `setOpen` rather than 5 individual props?

packages/interface/src/components/key/PasswordMeter.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+	const options = {
+		dictionary: {
+			...zxcvbnCommonPackage.dictionary,
+			...zxcvbnEnPackage.dictionary
+		},
+		graps: zxcvbnCommonPackage.adjacencyGraphs,
+		translations: zxcvbnEnPackage.translations
+	};
+	zxcvbnOptions.setOptions(options);
```
Brendonovich, line 17: this needs to be done outside the component - atm it's rerunning every time the component rerenders. it'd also be nice if `zxcvbn` stuff was dynamically loaded so that we didn't have to load a huge dictionary just for this one component to render

packages/interface/src/components/onboarding/Onboarding.tsx
```diff
@@ -1,27 +1,19 @@
-import clsx from 'clsx';
 import { useState } from 'react';
 import { useNavigate } from 'react-router';
 
-import { Button, Input } from '../../../../ui/src';
-import { useOperatingSystem } from '../../hooks/useOperatingSystem';
+import { Button } from '../../../../ui/src';
```
Brendonovich, line 4:
```suggestion
import { Button } from '@sd/ui';
```

packages/ui/src/Select.tsx
```diff
@@ -58,6 +58,7 @@ export function SelectOption(props: PropsWithChildren<{ value: string }>) {
 				'focus:outline-none hover:text-white radix-disabled:opacity-50 hover:bg-accent '
 			)}
 			value={props.value}
+			key={props.value}
```
Brendonovich, line 61: bad

Brendonovich (changes requested):

Brendonovich (dismissed):

core/src/library/library_manager.rs
```diff
@@ -204,20 +201,19 @@ impl LibraryManager {
...
-			// TODO: Make setting up keys optional with rest of system before removing this.
-			todo!();
-		}
+		// setup master password
+		let verification_key = KeyManager::onboarding(
+			km_config.algorithm,
+			km_config.hashing_algorithm,
+			km_config.password,
+			km_config.secret_key,
+		)?;
```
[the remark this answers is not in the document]
brxken128: `km_config` is declared in the core, onboarding is setup in the crypto crate. can move the struct though

packages/interface/src/components/dialog/BackupRestoreDialog.tsx
brxken128: will do this in another PR, would be a bit too large for this one as there's at least 3/4 dialogs that need refactoring to the same standard

packages/interface/src/components/dialog/CreateLibraryDialog.tsx
brxken128: ^ didn't realise it was tauri-specific, will try that and if not i'll go the `Platform` route

packages/interface/src/components/dialog/MasterPasswordChangeDialog.tsx
brxken128: i spent far too long trying to get this work, so will hold off on it for now. need to brush up on TS a little more

crates/crypto/src/keys/keymanager.rs
brxken128: i'll include this in my refactor, i'm not happy with lots of similar patterns and such too

packages/interface/src/components/dialog/MasterPasswordChangeDialog.tsx
Brendonovich: fairo

Brendonovich (approved):

Brendonovich (approved):

PR #509, opened 2023-01-09
[ENG-329] Crypto dialog refactor

brxken128:
This re-organises the dialogs (code-wise), and makes them a lot neater (and use they now use `react-hook-form`). Quite a small PR from me for once (with the amount done in it, not necessarily the additions/deletions).

I have tested all changed dialogs, and they seem to function as intended.


Brendonovich (approved):
forms and dialog need a whole refactor but this is alright for now

PR #510, opened 2023-01-09
[ENG-327] Crypto refactor

brxken128:
Quite a large PR, I apologise in advance.

This PR doesn't make many fundamental code changes, but here is a list of everything:

- Fixed a JS bug returning the (`Option`) secret key as `Some("")`
- Ensured the Rust hashing code works exactly the same, irregardless of the above. We can't always prevent JS oddities such as this
- Lots of dead code was removed
- Comments were updated
- Rust is a **lot** more idiomatic and more optimised
- Updated `sd-crypto`'s clippy configuration to be stricter, and be correctly compatible with the `rspc` feature
- Preview media is now actually embedded into encrypted files when selected within the UI
- Marginally more efficient header item serialization
- `std::fs::File` was changed to the non-blocking `tokio::fs::File` where easily possible
- `Protected<T>` got `into_inner` where `T` implements `Debug` - this prevents some ugly `value.expose().clone()` patterns
- Metadata and preview media clones were massively reduced


packages/interface/src/components/dialog/BackupRestoreDialog.tsx
```diff
@@ -29,7 +29,7 @@ export const BackupRestoreDialog = (props: BackupRestorationDialogProps) => {
 	});
 
 	const onSubmit: SubmitHandler<FormValues> = (data) => {
-		const sk = data.secretKey ?? null;
+		const sk = data.secretKey === '' ? null : data.secretKey;
```
Brendonovich, line 32: This can be `data.secretKey || null`. `??` is for null/undefined, `||` is for any falsy value. You can also put `required` on the input to ensure it will never be empty in the first place.

Brendonovich (changes requested):

packages/interface/src/components/dialog/BackupRestoreDialog.tsx
brxken128: it's allowed to be empty, that's the thing. `data.secretKey ?? null` ended up returning `Some("")` to Rust, if it was empty.

brxken128: haven't experimented with any other values for it, as my main concern was properly getting `Some` and `None` functioning

brxken128: i think it returns `Some("")` because `data.secretKey ?? null` doesn't take into account that it won't be null, but `""`

core/src/object/fs/encrypt.rs
```diff
@@ -164,24 +164,29 @@ impl StatefulJob for FileEncryptorJob {
...
+						let extension = path.extension().map_or_else(
+							|| Ok("sdenc".to_string()),
+							|extension| {
+								Ok::<String, JobError>(
+									extension
+										.to_str()
+										.ok_or(JobError::EarlyFinish {
+											name: self.name().to_string(),
+											reason: "path isn't valid UTF-8".to_string(),
+										})?
```
fogodev, line 179: `JobError::EarlyFinish` isn't actually an error, at the `run` method for jobs, when a early finish is received we just emit a warn log and the `Err` is dropped. Is this the intended use?

fogodev (changes requested):

core/src/object/fs/encrypt.rs
```diff
@@ -227,12 +235,29 @@ impl StatefulJob for FileEncryptorJob {
...
+						let pvm_path = ctx
+							.library_ctx
+							.config()
+							.data_directory()
+							.join("thumbnails")
+							.join(object.cas_id + ".webp");
+
+						if fs::metadata(pvm_path.clone()).is_ok() {
+							let mut pvm_bytes = Vec::new();
+							let mut pvm_file = File::open(pvm_path)?;
```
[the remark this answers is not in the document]
brxken128: i can't change this, as `tokio::fs::File` doesn't seem to implement `std::io::Read`, and the crypto functions require `std::io::Read`. will have to find a solution at some point

packages/interface/src/components/dialog/BackupRestoreDialog.tsx
brxken128: nevermind, just understood what you said, will fix

fogodev (approved):

fogodev (approved):

jamiepine (approved):
LGTM

PR #511, opened 2023-01-09
Detect SD-encrypted files and display them as such

brxken128:
This adds ".sdenc" and our encrypted file's magic bytes (`ballapp` in hex) to the `file-ext` crate, so that the UI is able to detect them. It also uses the `Archive` icon for encrypted files, as there wasn't one more suitable.


crates/file-ext/src/extensions.rs
```diff
@@ -182,11 +182,11 @@ extension_category_enum! {
...
 	EncryptedExtension _ALL_ENCRYPTED_EXTENSIONS {
 		// Spacedrive encrypted file
-		Bit = [0x73, 0x64, 0x62, 0x69, 0x74],
+		Sdenc = [0x62, 0x61, 0x6C, 0x6C, 0x61, 0x70, 0x70],
 		// Spacedrive container
-		Box = [0x73, 0x64, 0x62, 0x6F, 0x78],
+		Sdcontainer = [0x73, 0x64, 0x62, 0x6F, 0x78],
 		// Spacedrive block storage,
-		Block = [0x73, 0x64, 0x62, 0x6C, 0x6F, 0x63, 0x6B],
+		Sdblock = [0x73, 0x64, 0x62, 0x6C, 0x6F, 0x63, 0x6B],
```
Brendonovich, line 189: can these be pascal case like `SdEnc`, `SdContainer`, `SdBlock`

Brendonovich (approved):

PR #513, opened 2023-01-10
[ENG-331] `StoredKey` overhaul

brxken128:
This only really makes two main changes:

* `serde` is now used for serialization/deserialization with crypto `enum`s in relation to Prisma
* `StoredKey` now has an appropriate `StoredKeyVersion`, and functions should now be set-up to correctly handle any potential differences between versions. Functions that create `StoredKey`s will always use `STORED_KEY_LATEST`, but functions that read/use keys will correctly handle the version.

This makes the KM code a little less neat, but the backwards-compatibility most definitely outweighs this issue.


crates/crypto/src/keys/keymanager.rs
```diff
@@ -509,48 +523,52 @@ impl KeyManager {
...
-				};
-
-				self.keymount.insert(uuid, mounted_key);
-
-				Ok(())
-			}
-			None => Err(Error::KeyNotFound),
-		}
+		self.keystore.get(&uuid).map_or_else(
+			|| Err(Error::KeyNotFound),
```
fogodev, line 527: As you're passing a value that can be eagerly evaluated without extra costs (`Err(Error::KeyNotFound)`), you can use `map_or` here too, as you used later in this snippet

Brendonovich (approved):

crates/crypto/src/keys/keymanager.rs
brxken128: good catch, didn't realise `map_or` would've done the trick 🙏🏻

Brendonovich (approved):

PR #514, opened 2023-01-11
[ENG-328] Keymanager mounting queue

brxken128:
This PR adds a key manager mounting queue, and a few routes for checking the mount status of certain keys. It uses a Rust `DashSet` internally (for the hardcoded/externally immutable queue - this one will help protect us against malicious clients and DOS attacks due to memory-hard hashing algorithms).

The Rust queue is used for only one (external) route - checking if the key manager is currently being unlocked. This is used multiple times throughout the app, so a route will likely be best. The `set_master_password()` function also temporarily takes an `invalidate` closure, so we can invalidate the query correctly within the crypto crate. With this, users cannot attempt to unlock the key manager multiple times, no matter where in the app they are (or where they first started the unlock).

The Rust queue is used and updated heavily within the key manager, and protections have been added to relevant functions in order to protect against a malicious client (one where the React queue has been removed).

We also use a `const Set<string>` within React, so that the UI is reactive without cluttering RSPC. This `Set` is passed to each key individually, so they can view and update their status accordingly. Users won't be able to mount the same key multiple times. A "Mounting..." status was also added for keys that are in the process of being mounted, so that users stay informed.

I have tested these changes and didn't notice any oddities! Things work great.

### DIFF
--- a/packages/interface/src/components/key/KeyList.tsx
+++ b/packages/interface/src/components/key/KeyList.tsx
@@ -48,7 +48,7 @@ export const ListOfKeys = () => {
 						data={{
 							id: key.uuid,
 							name: `Key ${key.uuid.substring(0, 8).toUpperCase()}`,
-							queue: mountingQueue,
+							queue: mountingQueue.current,
 							mounted: mountedKeys.includes(key),
 							default: defaultKey.data === key.uuid,
 							memoryOnly: key.memory_only,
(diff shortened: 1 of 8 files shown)


packages/interface/src/components/key/KeyList.tsx
```diff
@@ -18,6 +18,9 @@ export const SelectOptionKeyList = (props: { keys: string[] }) => (
 		))}
 	</>
 );
+
+const mountingQueue = new Set<string>();
```
Brendonovich, line 22: put this in a `useRef` in the component and access it with `mountingQueue.current`. not a huge change but is more react-y

Brendonovich (changes requested):
defs need a review from @oscartbeaumont, idk what's going on with that frontend queue but if it's necessary then all good i guess

oscartbeaumont:
Just chatted with @Brendonovich and he thinks we can remove the `Set` and instead use optimistic UI with React Query. If this can be done it makes way more sense. I will give this a shot tomorrow.

Brendonovich (approved):

PR #515, opened 2023-01-11
Begin form abstraction

Brendonovich:
Form management right now is a fragmented nightmare, this PR addresses some of that using tips from [my blog post](https://www.brendonovich.dev/blog/the-ultimate-form-abstraction).

- `useZodForm` replaces `useForm` and requires defining a zod schema
- `Form` automatically wraps a `<form>` with a `FormProvider` and `<fieldset>`
- `Dialog` now requires `form` and `onSubmit` instead of not properly using a form internally and using the `ctaAction` callback
- Form-aware UI elements will be available in `@sd/ui/src/forms`. They all work with `form.register`, with some using `useController` under the hood.

A nice side effect of these changes is how it ties dialogs together. The dialog's disabled state is the same as the form's disabled state, all inputs automatically get disabled thanks to `<fieldset>`, and if `onSubmit` is async and uses `mutateAsync`, the dialog's loading state will be synced to the form's `isLoading` state.

Could definitely do with more work, as well as more standardisation around how dialogs are handled, but this is a good start.

Also since when was `type="submit"` the default for buttons in the html spec?!


packages/interface/src/components/dialog/AlertDialog.tsx
```diff
@@ -28,16 +30,18 @@ export interface AlertDialogProps {
 }
 
 export const AlertDialog = (props: AlertDialogProps) => {
+	const form = useZodForm({ schema: z.object({}) });
 	// maybe a copy-to-clipboard button would be beneficial too
 	return (
 		<Dialog
+			form={form}
```
Brendonovich, line 37: yeah cringe but i'll come up with something better later

oscartbeaumont (approved):

brxken128 (approved):
the forms look a lot better and easier to follow lol, nice job!

PR #521, opened 2023-01-13
[ENG-333] Handle .spacedrive existing instead of throwing error

fogodev:
Removing metadata files that failed to be deserialized, this can happen due to an older version of these `.spacedrive` metadata files, before this feature was actually implemented, or if the file was corrupted.

### DIFF
--- a/core/src/location/metadata.rs
+++ b/core/src/location/metadata.rs
@@ -49,16 +49,28 @@ impl SpacedriveLocationMetadataFile {
 				metadata: match serde_json::from_slice(&data) {
 					Ok(data) => data,
 					Err(e) => {
-						error!(
-							"Failed to deserialize corrupted metadata file, \
-						we will remove it and create a new one; File: {}; Error: {e}",
-							metadata_file_name.display()
-						);
-						fs::remove_file(&metadata_file_name).await.map_err(|e| {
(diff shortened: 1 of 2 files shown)


core/src/location/metadata.rs
```diff
@@ -45,10 +46,22 @@ impl SpacedriveLocationMetadataFile {
...
 			Ok(data) => Ok(Some(Self {
+				metadata: match serde_json::from_slice(&data) {
+					Ok(data) => data,
+					Err(e) => {
+						error!(
+							"Failed to deserialize corrupted metadata file, \
+						we will remove it and create a new one; File: {}; Error: {e}",
+							metadata_file_name.display()
+						);
+						fs::remove_file(&metadata_file_name).await.map_err(|e| {
```
Brendonovich, line 57: Might be worth doing this conditionally so that files aren't accidentally deleted in prod

Brendonovich (changes requested):
Works fine for dev but we probs wanna be careful to not forget about this

core/src/location/metadata.rs
fogodev: You're totally right, this is an excellent point, I introduced back the deserialize error, but only for production builds. This way, if this problem happens again in the future, we can introduce some kind of versioning schema without the using losing important metadata.

Brendonovich (approved):

PR #522, opened 2023-01-14
[ENG-342] Block location watcher for files that are being actively modified

fogodev:
Introducing methods to temporarily stops and reinit watching over a location

### DIFF
--- a/core/src/location/manager/helpers.rs
+++ b/core/src/location/manager/helpers.rs
@@ -10,7 +10,10 @@ use tokio::{fs, io::ErrorKind, time::sleep};
 use tracing::{error, warn};
 use uuid::Uuid;
 
-use super::{watcher::LocationWatcher, LocationId, LocationManagerError, ManagerMessage};
+use super::{
+	watcher::LocationWatcher, IgnorePathManagerMessage, LocationId, LocationManagerError,
+	ManagerMessage,
+};
 
 type LibraryId = Uuid;
 type LocationAndLibraryKey = (LocationId, LibraryId);
@@ -278,6 +281,20 @@ pub(super) async fn handle_reinit_watcher_request(
 					reason: String::from("failed to fetch location from db"),
 				})
 			}
+		} else {
+			Ok(())
+		},
+	); // ignore errors, we handle errors on receiver
+}
+
+pub(super) fn handle_ignore_path_request(
+	message: IgnorePathManagerMessage,
+	locations_watched: &HashMap<LocationAndLibraryKey, LocationWatcher>,
+) {
+	let (location_id, library_ctx, path, ignore, response_tx) = message;
+	let _ = response_tx.send(
+		if let Some(watcher) = locations_watched.get(&(location_id, library_ctx.id)) {
+			watcher.ignore_path(path, ignore)
 		} else {
 			Ok(())
 		},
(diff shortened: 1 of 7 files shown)


Brendonovich:
Does this PR address the entirety of the linked issue? Surely while encrypting a file we don't have to stop watching the location _entirely_, just ignore events for the files being processed.

fogodev:
> Does this PR address the entirety of the linked issue? Surely while encrypting a file we don't have to stop watching the location _entirely_, just ignore events for the files being processed.

Just introduced this part of ignoring events based on a path, used a guard to temporarily ignore those events.

core/src/location/manager/helpers.rs
```diff
@@ -163,3 +166,137 @@ pub(super) fn subtract_location_path(
...
+						locations_watched,
+						locations_unwatched,
+					);
+					forced_unwatch.remove(&key);
+
+					Ok(())
+				} else {
+					Err(LocationManagerError::LocationMissingLocalPath(location_id))
+				}
+			} else {
```
brxken128, line 279: a tad messy (along with the other one), but if it's too much hassle to change it'll be fine

core/src/location/manager/helpers.rs
```diff
@@ -163,3 +166,137 @@ pub(super) fn subtract_location_path(
...
+				Err(LocationManagerError::FailedToStopOrReinitWatcher {
+					reason: String::from("failed to fetch location from db"),
+				})
+			}
+		} else {
+			Ok(())
+		},
+	); // ignore errors, we handle errors on receiver
```
brxken128, line 287: a tad messy, but if it's too much hassle to change then not an issue (same with the other similar pattern)

brxken128 (approved):
looks great, haven't tested but provided you have i don't see anything that should really break (and it fully addresses the issue, which is a major plus in my eyes haha)

core/src/location/manager/helpers.rs
fogodev: Changed to a more functional style

Brendonovich (approved):
I wonder if we could use something like Actix or build our own actor abstraction to simplify the actors & message channels, it's a bit unwieldy atm 😅

jamiepine:
> Also, could the location watching be moved to its own crate?

Potentially, but I think it's good to merge for now.

jamiepine (dismissed):
Great work!

core/src/location/manager/mod.rs
```diff
@@ -115,10 +126,48 @@ impl LocationManager {
...
+			Ok(())
+		}
+	}
+
 	#[cfg(feature = "location-watcher")]
 	async fn run_locations_checker(
 		mut add_locations_rx: mpsc::Receiver<ManagerMessage>,
 		mut remove_locations_rx: mpsc::Receiver<ManagerMessage>,
+		mut stop_watcher_rx: mpsc::Receiver<ManagerMessage>,
+		mut reinit_watcher_rx: mpsc::Receiver<ManagerMessage>,
```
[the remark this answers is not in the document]
fogodev: I agree with you, just did some refactoring to introduce nicer abstractions and unifying those channels in a more extensible way, so it's easier to introduce more features in the future.

Brendonovich (approved):

PR #524, opened 2023-01-16
[ENG-257] File erasure and deletion

brxken128:
This PR brings the context menu's delete and erase options to life.

They both do not have "delete all occurrences of this file" config options yet, but those can wait for now.

I have tested erase, and the correct amount of bytes are overwritten (after an annoying typo was found).

This also brings forward what I think is an issue with the location watcher, where we get a `file not found` error after deleting a file (after the file was erased, truncated, dropped and then deleted with `fs::delete()`).

They are currently not `async`, but this will change once the crypto crate gets async support for I/O operations.


packages/interface/src/components/explorer/FileItem.tsx
```diff
@@ -29,6 +29,8 @@ interface Props extends HTMLAttributes<HTMLDivElement> {
 	index: number;
 	setShowEncryptDialog: (isShowing: boolean) => void;
 	setShowDecryptDialog: (isShowing: boolean) => void;
+	setShowDeleteDialog: (isShowing: boolean) => void;
+	setShowEraseDialog: (isShowing: boolean) => void;
```
jamiepine, line 33: at this point we'd be much better off with `setShowDialog(name, isShowing)`, but not needed for this PR.

jamiepine (approved):
tested and it works magically, love love love to see it! :D

packages/interface/src/components/explorer/FileItem.tsx
Brendonovich: Have done the form refactor, my next project is a dialog refactor haha

PR #527, opened 2023-01-18
[ENG-294] Volume awareness

maxichrome:
On macOS, directory listings from `sysinfo` are less than ideal — sometimes the primary drive (usually "Macintosh HD") is missing entirely, and sometimes system recovery volumes show up (such as "iSCPreboot"). This poses an obvious user experience deficit for a file manager; so we've decided the easiest solution is to use Swift, which plays very nicely with macOS' native system APIs, to fetch this volume listing and related info instead where possible/ergonomic.

Closes ENG-294


Brendonovich:
`packages/macos` should be moved to `core/macos` and have the `swift-rs` bindings included alongside it

PR #529, opened 2023-01-18
[ENG 239] Onboarding Flow & Location Settings

jamiepine:
This is a very basic implementation of onboarding. It is missing animated transitions, extra screens and better loading state during library creation.

Location settings are complete aside from indexer rules functionality, will complete in a separate PR.

https://user-images.githubusercontent.com/32987599/216482984-dd317753-dc93-4c5f-9ceb-403e8694b77c.mov


jamiepine:
Okay I'm about to build the "creating library" screen, also means we can have skippable steps in the library creation process

core/prisma/schema.prisma
```diff
@@ -182,6 +182,10 @@ model FilePath {
     parent_id Int?
     key_id    Int? // replacement for encryption
     // permissions       String?
+    
+    // this path will exist when the owner node of the location comes online
+    // allows for optimistic file management
+    pending Boolean @default(false)
```
Brendonovich, line 188: Can you elaborate on this? I'm a little nervous creating owned data for a non-owner node.

core/src/api/libraries.rs
```diff
@@ -80,12 +80,29 @@ pub(crate) fn mount() -> RouterBuilder {
 			#[derive(Deserialize, Type)]
 			pub struct CreateLibraryArgs {
 				name: String,
-				password: Protected<String>,
+				password: Option<Protected<String>>,
+				tokenized_password: Option<String>,
```
Brendonovich, line 84: Could this be an enum allowing for one or the other but not both?

core/src/api/locations.rs
```diff
@@ -293,4 +295,19 @@ fn mount_indexer_rule_routes() -> RouterBuilder {
...
+		// list indexer rules for location, returning the indexer rule
+		.library_query("listForLocation", |t| {
+			t(|_, location_id: i32, library| async move {
+				library
+					.db
+					.indexer_rules_in_location()
+					.find_many(vec![indexer_rules_in_location::location_id::equals(
+						location_id,
+					)])
+					.include(indexer_rules_in_location::include!({ indexer_rule }))
```
Brendonovich, line 307: This might work better as a query on `indexer_rules` with a filter of `indexer_rules::locations::some(vec![indexer_rules_in_location::location_id::equals(location_id)])`

core/src/location/mod.rs
```diff
@@ -167,13 +170,26 @@ impl LocationUpdateArgs {
...
+		if let Some(gpm) = self.generate_preview_media {
+			params_to_update.push(location::generate_preview_media::set(gpm))
+		}
+		if let Some(spm) = self.sync_preview_media {
+			params_to_update.push(location::sync_preview_media::set(spm))
+		}
+
+		if let Some(hidden) = self.hidden {
+			params_to_update.push(location::hidden::set(hidden))
+		}
```
Brendonovich, line 187: Personally I'd prefer this done as an array of `Option<SetParam>` which you `flatten` to get rid of `None` values, that way you can avoid most of the `if let Some` checks.

packages/client/src/stores/onboardingStore.ts
```diff
@@ -0,0 +1,50 @@
...
+
+import { valtioPersist } from './util';
+
+export enum UseCase {
+	CameraRoll = 'cameraRoll',
+	MediaConsumption = 'mediaConsumption',
+	MediaCreation = 'mediaCreation',
+	CloudBackup = 'cloudBackup',
+	Other = 'other'
+}
```
Brendonovich, line 11: would rather a `const` than an `enum` but ehh

packages/interface/src/components/onboarding/OnboardingCreatingLibrary.tsx
```diff
@@ -0,0 +1,105 @@
...
+				library
+			]);
+			resetOnboardingStore();
+			navigate('/overview/');
+		},
+		onError: (err: any) => {
+			console.log({ libraryCreationError: err });
+			setError(err.message);
+			alert('fuck');
+		}
```
Brendonovich, line 44: This could be moved into `create` if you use `mutateAsync`

packages/interface/src/components/onboarding/OnboardingMasterPassword.tsx
```diff
@@ -0,0 +1,146 @@
...
+			} else {
+				form.setError('password_validate', {
+					type: 'manual',
+					message: 'Passwords do not match'
+				});
+			}
+		} else {
+			tokenizeSensitiveKey.mutate({
+				secret_key: data.password
+			});
```
Brendonovich, line 63: `mutateAsync` would be nice here so that the form gets disabled until the operation completes + you can move `onSuccess` and `onError` in

packages/interface/src/components/onboarding/OnboardingNewLibrary.tsx
```diff
@@ -0,0 +1,91 @@
...
+			name: (ob_store.newLibraryName as string) || ''
+		}
+	});
+
+	useUnlockOnboardingScreen();
+
+	const _onSubmit = form.handleSubmit(async (data) => {
+		getOnboardingStore().newLibraryName = data.name;
+		navigate('/onboarding/master-password');
+		return;
```
Brendonovich, line 39:
```suggestion
```

packages/interface/src/components/onboarding/OnboardingNewLibrary.tsx
```diff
@@ -0,0 +1,91 @@
...
+						<Button onClick={() => setImportMode(false)} variant="outline" size="sm">
+							Create new library
+						</Button>
+					</div>
+				) : (
+					<>
+						<Input
+							{...form.register('name')}
+							//@ts-expect-error - size prop conflicts for some reason, despite being a valid variant
+							size="md"
```
Brendonovich, line 71: this happens bc `InputProps` is an intersection of cva + regular props, doing `Omit<..., 'size'>` for the base props would fix this and only use cva's `size` prop.

packages/interface/src/components/onboarding/OnboardingPrivacy.tsx
```diff
@@ -0,0 +1,82 @@
...
+import { Controller } from 'react-hook-form';
+import { useNavigate } from 'react-router';
+
+import { useUnlockOnboardingScreen } from './OnboardingProgress';
+import { OnboardingContainer, OnboardingDescription, OnboardingTitle } from './OnboardingRoot';
+
+const { Input, z, useZodForm, Form } = forms;
+
+const schema = z.object({
+	shareTelemetryDataWithDevelopers: z.string()
```
Brendonovich, line 12: maybe use a literal union for added typesafety?

packages/interface/src/components/onboarding/OnboardingPrivacy.tsx
```diff
@@ -0,0 +1,82 @@
...
+				<OnboardingTitle>Your Privacy</OnboardingTitle>
+				<OnboardingDescription>
+					Spacedrive is built for privacy, that's why we're open source and local first. So we'll
+					make it very clear what data is shared with us.
+				</OnboardingDescription>
+				<div className="m-4">
+					<Controller
+						control={form.control}
+						name="shareTelemetryDataWithDevelopers"
+						render={({ field }) => (
```
Brendonovich, line 54: ideally this would be a `useController` and be hidden away in `ui/src/forms/RadioGroup.tsx`, I can implement that if you want.

packages/interface/src/screens/settings/library/location/EditLocation.tsx
```diff
@@ -0,0 +1,189 @@
...
+	const onSubmit = form.handleSubmit(async (data) => {
+		console.log({ data });
+		updateLocation.mutate({
+			id: Number(id),
+			name: data.displayName,
+			sync_preview_media: data.syncPreviewMedia,
+			generate_preview_media: data.generatePreviewMedia,
+			hidden: data.hidden,
+			indexer_rules_ids: []
+		});
```
Brendonovich, line 66: i'm guessing you're not using `mutateAsync` so that the whole page isn't disabled?

packages/ui/src/RadioGroup.tsx
```diff
@@ -0,0 +1,41 @@
...
+export const Item = ({ children, ...props }: RadioGroup.RadioGroupItemProps) => {
+	return (
+		<div className="flex max-w-sm px-4 py-3 space-x-2 border rounded-md border-app-line bg-app-box/50">
+			<RadioGroup.Item
+				id={'radio' + props.value}
+				className={cx(
+					'peer relative w-4 h-4 rounded-full mr-1 mt-1 flex-shrink-0 border border-transparent',
+					'radix-state-checked:bg-accent',
+					'radix-state-unchecked:bg-gray-100 dark:radix-state-unchecked:bg-gray-900',
+					'focus:outline-none focus:ring-0 focus:ring-offset-0 focus-visible:ring focus-visible:ring-accent focus-visible:ring-opacity-75 focus-visible:ring-offset-2 dark:focus-visible:ring-offset-gray-800'
```
Brendonovich, line 28: i didn't think we use `dark:` since we do themes?

Brendonovich (changes requested):
make ci pass and answer questions and i'm happy

core/prisma/schema.prisma
jamiepine: Well as the app inherently allows you to browse locations that are offline, they would currently need to be read-only. I had always imagined that we would allow users to organize files for online and offline locations just the same. Changes to offline locations would happen optimistically and be applied later once the location comes online. This was one way I thought it could be accomplished.

core/src/api/libraries.rs
jamiepine: It very much should be an enum

core/src/api/locations.rs
jamiepine: Yeah I let copilot do this one, I'm almost clueless when it comes to querying relations with PCR, might need a refresher crash course. I did notice the data was not returned ideally.

core/src/location/mod.rs
jamiepine: Not sure how to accomplish this one.

packages/client/src/stores/onboardingStore.ts
jamiepine: Like an object? What would the const

packages/interface/src/components/onboarding/OnboardingCreatingLibrary.tsx
jamiepine: create?

packages/interface/src/components/onboarding/OnboardingMasterPassword.tsx
jamiepine: You might have to show me what you mean here cause I don't follow. I was under the impression mutateAsync went against the conventions of react hook form.

packages/interface/src/components/onboarding/OnboardingNewLibrary.tsx
jamiepine: Where would the Omit type go?

packages/interface/src/components/onboarding/OnboardingPrivacy.tsx
jamiepine: Will look into how to do that.

packages/interface/src/components/onboarding/OnboardingPrivacy.tsx
jamiepine: Thought of doing this, figured I'd just do it later. You can if you want.

packages/client/src/stores/onboardingStore.ts
Brendonovich: Same as the enum, dw about it

packages/interface/src/components/onboarding/OnboardingCreatingLibrary.tsx
Brendonovich: The function you define further down

packages/interface/src/components/onboarding/OnboardingMasterPassword.tsx
Brendonovich: Nah `mutateAsync` is great since `handleSubmit` takes an async function, we can keep the form in a disabled/loading state until the function is done.

packages/interface/src/components/onboarding/OnboardingNewLibrary.tsx
Brendonovich: On the `React.HTMLInputAttributes` of `InputProps` in `ui/src/Input.ts`

packages/interface/src/screens/settings/library/location/EditLocation.tsx
jamiepine: Same as above, I'm just not used that pattern with React Query

packages/ui/src/RadioGroup.tsx
jamiepine: Copilot moment, missed that.

core/src/location/mod.rs
Brendonovich: I'll do it if u want, also should `location.name` really be optional? I'd imagine each location should be required to have a name, even if it is empty.

jamiepine:
CI did pass until I merged main 😭

core/src/location/mod.rs
jamiepine: Oh absolutely, but what if I don't want to update the name? I should submit the same name anyway?

Brendonovich: I mean optional in the schema, being optional in the args is fine.
image

packages/interface/src/components/explorer/FileItem.tsx
```diff
@@ -29,7 +29,7 @@ interface Props extends HTMLAttributes<HTMLDivElement> {
 
 function FileItem({ data, selected, index, ...rest }: Props) {
 	const objectData = data ? (isObject(data) ? data.item : data.item.object) : null;
-	const isVid = isVideoExt(data.item.extension || '');
+	const isVid = ObjectKind[objectData?.kind || 0] === 'Video';
```
utkubakir, line 32: FYI: Mobile needs this too, I moved most of the utils (ObjectKind, isObject, isPath etc.) desktop uses to `@sd/client` on my `mobile-inspector` branch

core/prisma/schema.prisma
Brendonovich: I'd like if we held off on this, `pending` isn't used anywhere yet and I'd like for some more discussion around this topic.

core/src/location/mod.rs
```diff
@@ -465,3 +481,24 @@ pub async fn delete_directory(
 
 	Ok(())
 }
+
+// check if a path exists in our database at that location
+pub async fn check_virtual_path_exists(
+	library_ctx: &LibraryContext,
+	location_id: i32,
+	subpath: &PathBuf,
```
fogodev, line 489: We're using `impl AsRef<Path>` in most places to accept any kind of type that can be a reference to a `Path`. If this function will still be used, it would be nice to change the type, otherwise just remove the unused function.

fogodev (changes requested):

core/src/api/libraries.rs
```diff
@@ -80,12 +80,29 @@ pub(crate) fn mount() -> RouterBuilder {
...
+				let password = match args.password {
+					Some(password) => password,
+					None => {
+						if args.tokenized_password.is_none() {
+							return Err(Error::new(
+								ErrorCode::BadRequest,
+								"No password or tokenized password provided".into(),
+							));
+						}
+						// TODO: remove unwraps
```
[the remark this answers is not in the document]
Brendonovich: Fixed with the `AuthOptions` enum

core/prisma/schema.prisma
jamiepine: I'll write up a linear issue instead.

jamiepine: I'm having an issue running `pnpm prisma migrate dev`: 

> Migration `20230202133507_keytype` failed to apply cleanly to the shadow database.

Could you remove this and run migrations for me? I already reverted because I want to continue.

core/src/object/fs/create.rs
```diff
@@ -0,0 +1,30 @@
...
+// 		.ok_or(LocationError::IdNotFound(location_id))?;
+
+// 	let name = name.unwrap_or("Untitled Folder");
+
+// 	let exists = check_virtual_path_exists(library_ctx, location_id, subpath).await?;
+
+// 	std::fs::create_dir_all(&obj_path)?;
+
+// 	Ok(())
+// }
```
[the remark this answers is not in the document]
jamiepine: Its fine to leave it there for now I'll, make a separate PR to pick up from here.

fogodev (dismissed):

PR #531, opened 2023-01-19
New dialog system

Brendonovich:
Wrapping dialogs in triggers and managing states is time consuming, complex, and inefficient as React has to process every single dialog.
To remedy this, dialogs are now created using an imperative api via `DialogManager` (names still up in the air). `create` accepts a function which should return JSX for a dialog components, and will control the entire lifecycle of the dialog.


jamiepine (approved):
solid improvement

packages/interface/src/components/explorer/ExplorerContextMenu.tsx
```diff
@@ -282,8 +290,14 @@ export function FileItemContextMenu(props: FileItemContextMenuProps) {
...
 					variant="danger"
 					keybind="⌘DEL"
-					onClick={(e) => {
-						props.setShowDeleteDialog(true);
+					onClick={() => {
+						dialogManager.create((dp) => (
+							<DeleteFileDialog
+								{...dp}
+								location_id={getExplorerStore().locationId!}
+								path_id={props.item.id}
```
utkubakir, line 298: Why not `locationId` and `pathId`? :D

Brendonovich: That's what the prop names were beforehand I'm pretty sure, I just want bothered to go through and update them

PR #533, opened 2023-01-20
Enforce consistent formatting

maxichrome:
After lots of debate back and forth, the Semi-Executive:tm: Decision has been made to enforce Prettier formatting via a pre-commit hook. This hook will silently format the staged file contents before they are committed. This takes the responsibility for proper formatting away from the individual developer's dev environment, even though it's still highly recommended to keep in-editor formatting so there's no surprise about Prettier's output. (Also gives you ample opportunity to add any necessary `// prettier-ignore` comments (though they should be verrry rare considering how generally acceptable Prettier's changes are)


maxichrome:
CI BROKE, WORKING ON IT

Brendonovich (approved):

PR #534, opened 2023-01-20
Fix webkit reload

Brendonovich:
reloading wasn't accounted for when upgrading swift-rs

### DIFF
--- a/apps/desktop/src-tauri/native/macos/Sources/webview.swift
+++ b/apps/desktop/src-tauri/native/macos/Sources/webview.swift
@@ -1,7 +1,8 @@
 import WebKit
 
 @_cdecl("reload_webview")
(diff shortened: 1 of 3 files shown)


apps/desktop/src-tauri/native/macos/Sources/webview.swift
```diff
@@ -1,7 +1,8 @@
 import WebKit
 
 @_cdecl("reload_webview")
-public func reloadWebview(webview: WKWebView) -> () {
+public func reloadWebview(webviewPtr: UnsafePointer<WKWebView>) -> () {
+    let webview = webviewPtr.pointee;
```
maxichrome, line 5: tab

```suggestion
	let webview = webviewPtr.pointee;
```

maxichrome (changes requested):
tab

PR #535, opened 2023-01-20
Refactor routes

Brendonovich:
The huge collection of routes in `AppRouter` was annoying me so I separated them into nested routes objects. Colocating the routes and their components is nice since you don't end up with super long import statements. No one asked for this, I was just tired and felt like doing it.


jamiepine:
no one asked bro, but lgtm

jamiepine (approved):

PR #538, opened 2023-01-21
Fix prettier issues

maxichrome:
Building off the back of #533...

- ~~**Remove [`packages/client/src/core.ts`](https://github.com/spacedriveapp/spacedrive/blob/e4e80fa603c283f240884a7c76c3e0e3badcf7b4/packages/client/src/core.ts) and Git ignore it**~~
  
  ~~I've been on this train for a while -- I believe generated code should stay uncommitted where possible & reasonable. This file just supplies types for rspc queries and is always generated during `pnpm prep` anyway.~~
  
  If there's any reason to keep it around, it should be added to [`.prettierignore`](https://github.com/spacedriveapp/spacedrive/blob/a0975ec8c0036b944ad1f762548ffc0785318c2a/.prettierignore).

- **Add types to `.prettierrc.js`**

  Just useful for editor hinting

- **Fix unparseable file issues**

  Prettier previously had problems parsing the following files, both of which issues have been resolved:

  - [`packages/config/src/vite.d.ts`](https://github.com/spacedriveapp/spacedrive/blob/a0975ec8c0036b944ad1f762548ffc0785318c2a/packages/config/vite/index.d.ts#L5)
  - [`packages/interface/scripts/generateSvgImports.mjs`](https://github.com/spacedriveapp/spacedrive/blob/a0975ec8c0036b944ad1f762548ffc0785318c2a/packages/interface/scripts/generateSvgImports.mjs#L1).

  This means we no longer need to keep these files in Prettier's ignore file and thus they can stay formatted properly!


maxichrome:
ok i see that the CI typecheck requires the rspc core.ts file, @Brendonovich & @oscartbeaumont would it be ergonomic to add a generation step just for it to CI or would that require a lot of dependency building & therefore follow my backup plan listed above?

Brendonovich:
core.ts needs to stay, it's required by Vercel but Vercel can't do the rust codegen stuff. I don't like it either but trust me it's there for a reason

maxichrome:
OK ty for clarifying!! Updating this PR with that change now

Brendonovich (approved):

PR #539, opened 2023-01-22
Eslint overhaul

Brendonovich:
Eslint has been kind of ignored, and I haven't had it set up in my editor until now. I've gone ahead and reorganised our eslint configs, used `extend` instead of spreading `require` so we get expected behaviour, and added a CI job for eslint to make sure we're being consistent.


utkubakir (dismissed):
Looks good!

PR #540, opened 2023-01-22
Add clean command

maxichrome:
In a team discussion earlier the idea was floated to create some form of `pnpm clean` command — this PR does just that, removing the following:

- root `/node_modules/`
- nested `**/node_modules/`
- rust compilation `/target/`
- swift compilation `**/.build/`
- next.js compilation/cache `**/.next/`
- subdirectories of nested desktop compilation `**/dist/**` (not removing directory because the `.gitignore` in there is purposeful)

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,8 @@
 		"client": "pnpm --filter @sd/client -- ",
 		"prisma": "cd core && cargo prisma",
 		"codegen": "cargo test -p sd-core api::tests::test_and_export_rspc_bindings -- --exact",
-		"typecheck": "turbo run typecheck"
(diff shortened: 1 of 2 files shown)


maxichrome:
Requesting others' review (specifically on desktop & mobile) for input on other directories that may be useful for a full refresh

Brendonovich:
For cleaning Rust artifacts there's already the `cargo clean` command, which removes the target folder and some others. Whether it is part of this clean script or not doesn't bother me, I won't be using it very often anyway

Brendonovich (approved):

PR #548, opened 2023-01-22
sortable tags

niikeec:
Adds the ability to sort tags. Tags are assigned a `position` and are sorted based on that.

Closes #547

### DIFF
--- a/core/prisma/migrations/20230122220754_tag_position/migration.sql
+++ b/core/prisma/migrations/20230122220754_tag_position/migration.sql
@@ -1,0 +1,31 @@
+/*
+  Warnings:
+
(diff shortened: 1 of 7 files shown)


Brendonovich (commented):
Since Spacedrive is a distributed application, ordering is going to be much more complex than just a single integer. We'll probably use fractional indexing but we need to have more discussions first.
The DND stuff is nice, but this solution isn't going to work long term.

niikeec:
Is that something that could be explored now or rather kept for the future?

Brendonovich:
If you've got ideas then pop them in [discussions](https://github.com/spacedriveapp/spacedrive/discussions), or if you want to have a chat then join our [Discord server](https://discord.com/invite/gTaF2Z44f5)!

PR #549, opened 2023-01-24
[ENG-348] Asynchronous crypto

brxken128:
This PR adds asynchronous crypto to the crypto crate, namely to everything that deals with streams/IO.

This also fixes broken crypto dialogs that were missing default values.

I have tested the changes, and things work as expected.

Benches have been disabled for the time being, as they need refactoring (and we could probably do with updating them in general).

I plan to do some performance benchmarking, to see if we gain any noticeable improvement. I don't expect much, but a little boost would be nice.

This also adds type aliases for `Salt`, `Key`, `EncryptedKey` and `ProtectedVec`.


brxken128:
Marking this as ready for review. Encryption is pretty slow due to `tokio`'s internal read buffer size, but I plan to use `tokio` as a git dependency and use 2MiB internal buffers (at least until it's changed upstream) - this will require some manual testing on all platforms to make sure nothing breaks.

For now, this does successfully bring in working asynchronous crypto.

crates/crypto/src/header/metadata.rs
```diff
@@ -137,26 +143,27 @@ impl FileHeader {
...
-				)?;
+				)
+				.await?;
 
 				serde_json::from_slice::<T>(&metadata).map_err(|_| Error::Serialization)
-			},
-		)
+			}
+			None => Err(Error::NoMetadata),
+		}
```
fogodev, line 166: This match can also be an if let with else

crates/crypto/src/header/preview_media.rs
```diff
@@ -86,53 +86,55 @@ impl FileHeader {
...
-				)?;
+				)
+				.await?;
 
 				Ok(pvm)
-			},
-		)
+			}
+			None => Err(Error::NoPreviewMedia),
+		}
```
fogodev, line 109: This match can also be an if let with else

crates/crypto/src/header/preview_media.rs
```diff
@@ -86,53 +86,55 @@ impl FileHeader {
...
```
fogodev, line 137: Another match on Option

crates/crypto/src/keys/keymanager.rs
```diff
@@ -39,14 +39,11 @@ use std::sync::Mutex;
...
-use crate::{
-	crypto::stream::Algorithm,
-	primitives::{ENCRYPTED_KEY_LEN, SALT_LEN},
-	Protected,
+	derive_key, generate_master_key, generate_nonce, generate_salt, to_array, EncryptedKey, Key,
+	OnboardingConfig, Salt, KEY_LEN, LATEST_STORED_KEY, MASTER_PASSWORD_CONTEXT, ROOT_KEY_CONTEXT,
 };
+use crate::protected::ProtectedVec;
+use crate::{crypto::stream::Algorithm, primitives::ENCRYPTED_KEY_LEN, Protected};
 use crate::{Error, Result};
```
fogodev, line 47: These `use crate` can be unified

crates/crypto/src/keys/keymanager.rs
```diff
@@ -99,7 +96,7 @@ pub struct MountedKey {
 ///
 /// Use the associated functions to interact with it.
 pub struct KeyManager {
-	root_key: Mutex<Option<Protected<[u8; KEY_LEN]>>>, // the root key for the vault
+	root_key: Mutex<Option<Protected<Key>>>, // the root key for the vault
```
fogodev, line 99: Beware that these sync Mutexes's locks mustn't be held across await points

crates/crypto/src/keys/keymanager.rs
```diff
@@ -512,16 +525,15 @@ impl KeyManager {
...
 		if self.keymount.get(&uuid).is_some() {
 			return Err(Error::KeyAlreadyMounted);
 		} else if self.is_queued(uuid) {
 			return Err(Error::KeyAlreadyQueued);
 		}
 
-		self.keystore
-			.get(&uuid)
-			.map_or(Err(Error::KeyNotFound), |stored_key| {
+		match self.keystore.get(&uuid) {
```
fogodev, line 535: Matching on Option instead of if let with else

crates/crypto/src/keys/keymanager.rs
```diff
@@ -575,23 +589,25 @@ impl KeyManager {
...
 
 	/// This function is used for getting the key value itself, from a given UUID.
 	///
 	/// The master password/salt needs to be present, so we are able to decrypt the key itself from the stored key.
-	pub fn get_key(&self, uuid: Uuid) -> Result<Protected<Vec<u8>>> {
-		self.keystore
-			.get(&uuid)
-			.map_or(Err(Error::KeyNotFound), |stored_key| {
+	pub async fn get_key(&self, uuid: Uuid) -> Result<ProtectedVec<u8>> {
+		match self.keystore.get(&uuid) {
```
fogodev, line 605: Match on Option

fogodev (changes requested):

core/src/api/keys.rs
```diff
@@ -169,7 +172,8 @@ pub(crate) fn mount() -> RouterBuilder {
 				for key in automount {
 					library
 						.key_manager
-						.mount(Uuid::from_str(&key.uuid).map_err(|_| Error::Serialization)?)?;
+						.mount(Uuid::from_str(&key.uuid).map_err(|_| Error::Serialization)?)
+						.await?;
```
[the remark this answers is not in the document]
brxken128: Not wise at all, that could potentially result in 6gb+ of memory used with just 12 keys.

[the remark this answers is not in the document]
fogodev: Damn boy, didn't know that lol, all good then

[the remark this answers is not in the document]
brxken128: I have thought about the idea previously, where we could potentially use `rayon` and dispatch them in batches relative to the total memory cost, although it's likely more effort than it's worth. It would probably result in faster automounting time, but at the cost of pretty high memory usage.

In prod builds on a modern machine, it happens quick enough that it's not an issue and it doesn't delay unlocking the key manager or anything like that.

core/src/api/keys.rs
```diff
@@ -290,11 +297,10 @@ pub(crate) fn mount() -> RouterBuilder {
...
-					secret_key,
-					&stored_keys,
-				)?;
+				let updated_keys = library
+					.key_manager
+					.import_keystore_backup(Protected::new(args.password), secret_key, &stored_keys)
+					.await?;
 
 				for key in &updated_keys {
 					write_storedkey_to_db(&library.db, key).await?;
```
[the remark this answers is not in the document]
Brendonovich: This should actually be a `create_many` so that they're all batched

fogodev (approved):

crates/crypto/src/primitives.rs
```diff
@@ -44,6 +44,10 @@ pub const MASTER_PASSWORD_CONTEXT: &str =
 	"spacedrive 2022-12-14 15:35:41 master password hash derivation"; // used for deriving keys from the master password hash
 pub const FILE_KEY_CONTEXT: &str = "spacedrive 2022-12-14 12:54:12 file key derivation"; // used for deriving keys from user key/content salt hashes (for file encryption)
 
+pub type Key = [u8; KEY_LEN];
+pub type EncryptedKey = [u8; ENCRYPTED_KEY_LEN];
+pub type Salt = [u8; SALT_LEN];
```
Brendonovich, line 49: I'd like if newtype structs were used here, they allow for much more flexibility with traits and functions, and make types much stricter.

packages/interface/src/components/dialog/DecryptFileDialog.tsx
```diff
@@ -59,26 +59,24 @@ export const DecryptFileDialog = (props: DecryptDialogProps) => {
...
+	const onSubmit = form.handleSubmit((data) =>
+		decryptFile.mutateAsync({
 			location_id: props.location_id,
 			path_id: props.path_id,
-			output_path: output,
-			password: pw,
-			save_to_library: save
-		});
-	});
+			output_path: data.outputPath !== '' ? data.outputPath : null,
```
Brendonovich, line 75: Not a fan of this check, would much rather the core deal with a blank string than the client have to specify default values. Keep in mind you can use `required` on inputs if you really need them to specify a value.

packages/interface/src/components/dialog/MasterPasswordChangeDialog.tsx
```diff
@@ -52,7 +52,9 @@ export const MasterPasswordChangeDialog = (props: MasterPasswordChangeDialogProp
 		schema,
 		defaultValues: {
 			encryptionAlgo: 'XChaCha20Poly1305',
-			hashingAlgo: 'Argon2id-s'
+			hashingAlgo: 'Argon2id-s',
+			masterPassword: '',
+			masterPassword2: ''
```
Brendonovich, line 57: Same stuff about default values here, if they must have a value then try to use `required`.

Brendonovich (changes requested):

packages/interface/src/components/dialog/DecryptFileDialog.tsx
brxken128: It's not required though. The check is there so rspc properly sends through a `None` value.

Brendonovich: Ehh why not just make rspc handle the empty string and not need a `None` in the first place

packages/interface/src/components/dialog/MasterPasswordChangeDialog.tsx
brxken128: Encryption and hashing algorithms need default values otherwise the Select boxes remain completely empty and broken on first launch. The first password needed a default value because the password meter crashed the entire app while it tried evaluating `password.length` with no password at all.

packages/interface/src/components/dialog/DecryptFileDialog.tsx
Brendonovich: gonna worry about this later

packages/interface/src/components/dialog/MasterPasswordChangeDialog.tsx
Brendonovich: gonna worry about this later

Brendonovich (approved):

crates/crypto/src/primitives.rs
Brendonovich: can do this later

PR #550, opened 2023-01-25
fix `spacedrive://` custom protocol on Windows

### LINKED ISSUE #520: ERR_UNKNOWN_URL_SCHEME - Windows
### Describe the bug

The URL scheme does not seem to be working on Windows. The dev tools console spits out the following, indicating an issue with the URL scheme on Windows.

```GET spacedrive://thumbnail/6bb08002b967e052 net::ERR_UNKNOWN_URL_SCHEME```

The thumbnails are being generated fine.

One thing that is maybe worth noting is that if you use the URL `https://spacedrive.anything/thumbnail/6bb08002b967e052`, the function for `register_uri_scheme_protocol` runs with `spacedrive://anything/thumbnail/6bb08002b967e052` as the value for the parsed URI. It does not run at all when trying to use `spacedrive://` though.

### Reproduction

Run the dev or production build on Windows.

### Expected behavior

_No response_

### Platform and versions

```Shell
pnpm 7.24.3
cargo 1.66.1 (ad779e08b 2023-01-10)
rustc 1.66.1 (90743e729 2023-01-10)

Operating System:
Edition: Windows 11 Pro
Version: 22H2
System Type: 64-bit operating system, x64-based processor
```

### Stack trace

_No response_

### Additional context

_No response_

oscartbeaumont:
Changes:
 - [x] Hopefully fixes #520 
 - [x] Move the `handle_custom_uri` function to use primitives from the `http` crate instead of my old janky tuple. Required for steaming and is probably how I should have built this system originally.
 - [x] Streaming the response for files using the protocol. This is so we don't need to load the whole thing into memory.
    - Streaming the response for any arbitrary file format requires chunked encoding. Support chunked encoding is hard so this PR **only deals with chunking of video files** which use a whole different system composed of multiple HTTP requests for the same resource which is way easier to implement.
 - [x] Add a new part to the protocol for loading in files (`spacedrive://file/{library_id}/{location_id}/{filepath_id}`)
 - [x] LRU cache for video file metadata in `spacedrive://file/*` proto
 - [x] JS helpers for generating the URL on web, desktop Unix and desktop Windows.
 - [x] Replace enums with string unions in Typescript. Enum's have many footguns!
 - To fix Linux I had to upgrade Tauri. This then necessitated updating both rspc and PCR for it all to work.


oscartbeaumont:
@jamiepine this works and is ready for use.

I would appreciate it if @brxken128 could give it a shot on Linux because I am concerned about if video files will work on Linux. I don't think it will stream video files on Linux and as far as I'm aware this is a Tauri limitation. It should instead, I am hoping, just load it all into memory and sends it to the webview. We could build a custom protocol to work around this limitation but that is a future concern.

@james-elicx I would appreciate a test of preview media loading on Windows (although don't feel obligated to). I was unable to use the Rust code snippet you sent as it was causing issues on macOS. The system I designed converts the `spacedrive://` (Linux/macOS style) to `http://spacedrive.localhost` (Windows style) and then the code works on that. This choice was because the URL parser in the `http` crate expects a full URL. Thanks for reporting the Windows issue!

Right now I added a button to the right-click menu on an explorer item to test this feature. It does a full webview redirect which is something that @jamiepine is going to improve when the UI he is working on is ready.

james-elicx:
> @james-elicx I would appreciate a test of preview media loading on Windows (although don't feel obligated to). I was unable to use the Rust code snippet you sent as it was causing issues on macOS. The system I designed converts the `spacedrive://` (Linux/macOS style) to `http://spacedrive.localhost` (Windows style) and then the code works on that. This choice was because the URL parser in the `http` crate expects a full URL. Thanks for reporting the Windows issue!
> 

I'm happy to try it out. Thanks for looking into the issue. Unfortunately, at the moment, it throws `Bad request: Invalid operation!`.

This is because on Windows, the request URI you get in the `register_uri_scheme_protocol` handler is in the format `spacedrive://localhost/thumbnail%2F27cc1db92478b45b`. Tauri appears to internally transform the request from `spacedrive.localhost` to `spacedrive://localhost`.

This means that the replaced URI ends up being `https://spacedrive.localhost/localhost/thumbnail%2F4b6c8875aef65af3` since you replace the `spacedrive://` part, leaving the first item in the path inside your `handle_custom_uri` function as `localhost`.

In the Tauri examples, they use `.strip_prefix("spacedrive://localhost/")` to get rid of this, although if you want to go down the replace route, I imagine you could do `let uri = uri.replace("spacedrive://localhost/", "spacedrive://");` so that requests from both OS start the same.

oscartbeaumont:
@james-elicx Thanks for taking the time to check it. I just pushed a change following your recommendation and hopefully, that should do it. Let me know if you find the time to test it!

james-elicx:
> @james-elicx Thanks for taking the time to check it. I just pushed a change following your recommendation and hopefully, that should do it. Let me know if you find the time to test it!

If you swap the two replaces around, it works :p

Just gotta do the Windows one first, otherwise, it's trying to replace a `spacedrive://localhost/` that isn't there due to the Unix one replacing the `spacedrive://` part already.

Many thanks for working on this :)

oscartbeaumont:
@james-elicx hopefully 3rd times a charm. Updated it again. Thanks for all of your help!

oscartbeaumont:
I have moved this back to draft and will try and fix it tommorow.

To make the video streaming work on Linux we need to upgrade Tauri to enable to the `linux-protocol-headers` feature. Upgrading Tauri breaks rspc. Upgrading rspc stops it working PCR. Updating PCR then allows the app to build but rspc is not:
 - Formatting the bindings file (would be super bad for PRs if not fixed)
 - Is just not exporting `FileEraserJobInit`, `FilePathWithObject`, `ObjectWithFilePaths` types

I will work on fixing these tomorrow and then hopefully we will have a bit more luck with Linux support. In the meantime this PR may not work and should be able to be fixed by downgrading the Tauri version in the `src/desktop/src-tauri/Cargo.toml`

oscartbeaumont:
This PR should be working now and it would be great if @brxken128 could check if video playback is working on Linux when he gets a minute. I would also appreciate a Windows test by @james-elicx if you can find the time.

I have upgraded Tauri (which required upgrading rspc, PCR & Axum to work). This has taken a while because in doing so I have revealed a bunch of issues in the main branch of rspc and the PCR integration for rspc but they are all fixed now.

Some changes to Spacedrive in this PR related to upgrading rspc:
 - Crypto package uses `rspc::Type` instead of `specta::Type` - Minor but will make sure we are always use the Specta version supported by rspc (very soon they are gonna split into two repos and will be versioned independently)
 - All bigint types (`usize`, `isize`, `i64`, `u64`, `i128`, `u128`) are now unsupported by rspc and must be string encoded to be sent through rspc. Previously these types were supported but`JSON.parse` would have lossily decoded them so I have decided to stop supporting them in rspc until I can build a solution.

The purpose behind upgrading Tauri was to allow us to enable the Tauri `linux-protocol-headers` feature which I am hoping makes video playback work on Linux. If it doesn't we might be a little screwed.

james-elicx:
> This PR should be working now and it would be great if @brxken128 could check if video playback is working on Linux when he gets a minute. I would also appreciate a Windows test by @james-elicx if you can find the time.

Looks like the protocol and thumbnails are working correctly on Windows now! I appreciate all the work you've done to fix this :)

brxken128:
I've already spoken to @oscartbeaumont about this, but on Linux (Gentoo) it's a mixed bag. Thumbnails work, and opening images work, but video playback does not. It could either be a Tauri thing, or something to do with our protocol - I'm not too sure here.

We're making progress though as only thumbnails worked before.

crates/crypto/src/crypto/stream.rs
```diff
@@ -23,7 +23,7 @@ use tokio::io::{AsyncReadExt, AsyncWriteExt};
 	derive(serde::Serialize),
 	derive(serde::Deserialize)
 )]
-#[cfg_attr(feature = "rspc", derive(specta::Type))]
+#[cfg_attr(feature = "rspc", derive(rspc::Type))]
```
Brendonovich, line 26: I'm guessing this has been changed bc of dependency versioning? Would be nice if it could stay as `specta` with a workspace patch or something.

Brendonovich (approved):

crates/crypto/src/crypto/stream.rs
oscartbeaumont: This wasn't changed because it was needed but because it's my general recommendation when using rspc and I am going to start pushing it harder in the docs soon so wanted us to align with that.

Doing it this way is different to a workspace patch because instead of using _Spacedrive's_ version of Specta, the change uses the version of Specta _rspc_ expects. This makes it impossible for the rspc and Specta versions being mismatched and causing the annoying and very unhelped `the trait IntoLayerResult<_> is not implemented for type` compile error.

I also do not see any benefit to it staying `specta::Type` but given `rspc::Type` has this benefit I think it makes sense.

apps/desktop/src-tauri/src/main.rs
```diff
@@ -38,26 +46,89 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+	let app = tauri::Builder::default().plugin(rspc::integrations::tauri::plugin(router, {
+		let node = node.clone();
+		move || node.get_request_context()
+	}));
+
+	// This is a super cringe workaround for: https://github.com/tauri-apps/tauri/issues/3725 & https://bugs.webkit.org/show_bug.cgi?id=146351#c5
+	// TODO: Secure this server against other apps on the users machine making requests to it using a HTTP header and random token or something
+	let endpoint = create_custom_uri_endpoint(node.clone());
+	#[cfg(target_os = "linux")]
+	let app = {
```
Brendonovich, line 58: can this be in another file

Brendonovich (changes requested):

Brendonovich (approved):

PR #554, opened 2023-01-31
Eng 341 os keychains

brxken128:
This PR adds a `KeyringInterface` to the `crypto` crate, so that we can make use of OS-controlled secure keystores for storing the secret key externally to Spacedrive.

The secret key acts as 2FA, and is required to unlock the vault but it would be extremely laborious to enter the key every time. By using OS keychains, we can automatically store/retrieve the secret key with minimal input from the user (a password may be required on the very first retrieval, depending on the OS).

The current implementation works with:

* Linux (this requires `DBus`, and either `gnome-keyring` or `kwallet`)
* MacOS and iOS (via the `security_framework` crate)

Windows support has yet to be added, due to the extreme complexity of developing for the platform. I only have Linux/MacOS development environments configured, but will setup one for Windows and add support for it in the very near future.


fogodev (approved):
LGTM

PR #557, opened 2023-02-02
Tailwind Prettier Plugin

utkubakir:
I implemented tailwind's prettier plugin...

Pros
- Consistent ordering can help find what you are looking for.
- Follows the recommended class ordering.
- Better than Headwind or similar extensions.

Cons
- ~~[Doesn't work inside variables and with `cva` etc.](https://github.com/tailwindlabs/tailwindcss/discussions/7558)~~
- ~~Doesn't work for mobile~~
- Can get annoying while styling the component


Brendonovich:
I wonder if we should instead use [`eslint-plugin-tailwindcss`](https://github.com/francoismassart/eslint-plugin-tailwindcss)? It uses `prettier-plugin-tailwind` for class sorting and is compatible with `cva` and possibly mobile.

utkubakir:
> I wonder if we should instead use [`eslint-plugin-tailwindcss`](https://github.com/francoismassart/eslint-plugin-tailwindcss)? It uses `prettier-plugin-tailwind` for class sorting and is compatible with `cva` and possibly mobile.

Eslint complaining at you is already annoying, wouldn't this take that to another level? Sure if they support `--fix`, we can run a script but prettier does its thing on every save.

Brendonovich:
you can tell `eslint` to `--fix` on save:
image
or could use both eslint + prettier 🤷‍♂️

Brendonovich (approved):

Brendonovich (approved):

PR #558, opened 2023-02-03
[ENG-355] Keychain integration

brxken128:
This PR leverages OS keychains in order to securely and externally store the secret key for each library.

I've only tested this on Linux so far, but things should work on MacOS too as the keyring functionality did. I'll do some testing to confirm though.

The TS to detect whether or not the user can use the OS keyring is not the best, and I cannot figure out a way to make it work flawlessly.

No config options are present as of yet, and we will attempt to use OS keyrings where we can.

This makes some large changes, the TL;DR:
* The secret key is no longer optional
* The secret key is no longer user-selectable
* The secret key is 18 bytes, hex encoded and separated with `-` every 6 characters
* The crypto MSRV was raised to 1.67.0
* Some routes had names changed in order to better represent their actions
* UI was updated to accommodate these changes
* Users can view the secret key in the settings page (with a QR code), provided OS keychains can be used to store it. If this is not the case, we have no way of knowing what the secret key is ourselves. In order to make this work with #529, we will need to provide the user with a pop-up showing the secret key while they set up their key manager.
* `StoredKey` now has a `key_type` attribute. This will prove to be **very** useful for ENG-354, but for now it cleans things up. The possible values are `User` and `Root` - `Root` being the "verification key". This lays out some groundwork for multiple ways to unlock the vault, and we will need some fancy UI in order to manage/configure these options.
* Related to the above, this will probably be expanded, and secret keys will be designated their own UUIDs within OS keystores, and these key types will become `Root(UUID)` to link them to a secret key. This will be needed when we offer more than one way to unlock the vault (e.g. 2x passwords).
* Migrations were updated
* Fixed a typo where `Hardened` and `Paranoid` parameter levels were the wrong way around
* `BLAKE3-Balloon` parameters were tweaked, and now they're a bit closer to `Argon2id` (memory usage is tiny in comparison, but hashing takes longer).
* Great typesafety for: `Key`, `EncryptedKey`, `Nonce`, `Salt`, `SecretKey` and `SecretKeyString`. This is the bulk of the PR (it felt somewhat fitting to include it here while I was also making architectural changes).
* The `ProtectedVec` type alias was removed. It didn't fit in with everything else
* A lot of code was cleaned up

The secret key is not removed from the OS keychain on library deletion. I'm not too sure on the best way to handle this - but no matter how it's managed, it will need some serious refactoring once ENG-354 is done (so we're probably better off leaving it for now).

`dbg!()` calls are present during `KeyManager::onboarding()` and for changing the master password, but these are required otherwise users wouldn't know their secret key in the event that OS keychains aren't present. Again, once #529 gets merged we can remove the `dbg!()`s and present the secret key for the user to write down.

I will keep this as a draft until CI passes and I'm happy with the code quality.


Brendonovich (approved):

core/src/api/keys.rs
```diff
@@ -79,6 +81,25 @@ pub(crate) fn mount() -> RouterBuilder {
 				Ok(())
 			})
 		})
+		.library_query("keyringHasSecretKey", |t| {
```
oscartbeaumont, line 84: I am not the biggest fan of any `hasSomething` queries. The main issue is that when it comes to invalidation the `getSomething` and `hasSomething` queries both need to be invalidated and could potentially get out of sync (especially an issue when we do optimistic UI on the frontend which I am sure we will at some point).

I understand for many use cases you may need both but a quick glance here seems like it doesn't but if it does feel free to disregard this comment.

oscartbeaumont (approved):

core/src/api/keys.rs
brxken128: Fixed. I completely agree with this, and we do have some `hasX`-style routes (like `keys.isUnlocked` for the key manager being unlocked) and they're necessary sometimes, but this one probably wasn't. Good catch and thank you!

utkubakir (approved):

PR #560, opened 2023-02-07
Move location online status out of database

Brendonovich:
Whether a location is online or not isn't really something that belongs in the db - it's directly related to the node's connections at the current moment, whether that's over the network or watching at the current moment.
It also doesn't make sense in the sync system, since online status shouldn't really be persisted.
Instead, locations are tracked in a global `BTreeSet` inside the `LocationManager`, with changes reported via a broadcast channel and subscribed to on the frontend.

### DIFF
--- a/core/src/location/manager/helpers.rs
+++ b/core/src/location/manager/helpers.rs
@@ -18,18 +18,22 @@ type LocationAndLibraryKey = (LocationId, LibraryId);
 const LOCATION_CHECK_INTERVAL: Duration = Duration::from_secs(5);
 
 pub(super) async fn check_online(location: &location::Data, library_ctx: &LibraryContext) -> bool {
+	let pub_id = &location.pub_id;
+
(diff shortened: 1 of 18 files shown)


core/src/location/manager/helpers.rs
```diff
@@ -39,34 +43,35 @@ pub(super) async fn check_online(location: &location::Data, library_ctx: &Librar
...
-		.exec()
-		.await
-	{
-		error!(
-			"Failed to update location to online: (id: {}, error: {:#?})",
-			location_id, e
-		);
-	}
-}
+// pub(super) async fn set_location_online(
```
oscartbeaumont, line 54: Can this just be removed?

oscartbeaumont (dismissed):

core/src/location/manager/helpers.rs
Brendonovich: oh yeah was gonna remove them lol

PR #562, opened 2023-02-08
[ENG-358] Malicious client protection for the KM

brxken128:
This PR adds checks to quickly and very easily verify a few conditions:

* The key manager is unlocked
* A key is not already mounted
* A key is not already in the process of being mounted

These checks were implemented pretty sparsely beforehand, and in the form of unsightly `if else` statements.


Brendonovich (approved):
love the `ensure_x` functions, quite elegant

PR #563, opened 2023-02-08
[Fix] Dropdown focus when app is launching

### LINKED ISSUE #191: Library selector has focus ring on initial render
On the app's initial render, the library chooser is focused by default and displays a visible focus ring. (See screenshot)

A screenshot of the very top-left corner of the macOS app. There is a dropdown with a light purple outline around it indicating that it currently has keyboard focus.

## Reproduction

1. Open the desktop app (in my case using `pnpm desktop dev`)
2. Look for the focus ring on the library chooser

## Expected behavior

- No element should initially have visible focus.
- Unless determined otherwise, pressing <kbd>Tab</kbd> should apply focus to the library chooser as it's the first visible onscreen element.

## Platform and versions

Commit: f193f45
OS: macOS Monterey version 12.4 (21F79)
Safari: Version 15.5 (17613.2.7.1.8)

```
❯ pnpm --version && cargo --version && rustc --version
7.1.5
cargo 1.60.0 (d1fd9fe2c 2022-03-01)
rustc 1.60.0 (7737e0b5c 2022-04-04)
```

JulesGuesnon:
Hey! 
Here is a fix for the dropdown button that is auto-focused when launching app. The fix is a bit hacky, but I saw a similar one [here](https://github.com/spacedriveapp/spacedrive/blob/b94eab8517d53e61ad2ae6016d90462fa9503abc/packages/interface/src/components/key/KeyMounter.tsx#L33-L37), so if you're fine with it, it's working.

Closes #191

### DIFF
--- a/packages/interface/src/components/layout/Sidebar.tsx
+++ b/packages/interface/src/components/layout/Sidebar.tsx
@@ -1,7 +1,7 @@
 import { ReactComponent as Ellipsis } from '@sd/assets/svgs/ellipsis.svg';
 import clsx from 'clsx';
 import { CheckCircle, CirclesFour, Gear, Lock, Planet, Plus } from 'phosphor-react';
-import React, { PropsWithChildren } from 'react';
+import React, { PropsWithChildren, useEffect } from 'react';
 import { NavLink, NavLinkProps } from 'react-router-dom';
 import {
 	Location,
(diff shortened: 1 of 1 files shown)


Brendonovich:
Appreciate the PR, but it'd be nice if we could get to the bottom of why it's happening in the first place - would you be up for that?

jamiepine (approved):
I fully accept the hacky fix, we should still look into it later @Brendonovich

JulesGuesnon:
I have a few ideas to try to debug, not sure if I'll find anything, but I'll try

PR #564, opened 2023-02-08
[ENG-357] Automatically mount associated key during decryption

brxken128:
This PR adds the (default) option to automatically mount a key while trying to decrypt a file.

It matches the content salts in the header's keyslots. If any content salt from a keyslot matches one of the keys in the key manager, it is mounted. Decryption then continues as normal.

### DIFF
--- a/core/src/object/fs/decrypt.rs
+++ b/core/src/object/fs/decrypt.rs
@@ -6,6 +6,7 @@ use serde::{Deserialize, Serialize};
 use specta::Type;
 use std::{collections::VecDeque, path::PathBuf};
 use tokio::fs::File;
+use uuid::Uuid;
 
 use crate::job::{JobError, JobReportUpdate, JobResult, JobState, StatefulJob, WorkerContext};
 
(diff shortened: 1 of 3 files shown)


core/src/object/fs/decrypt.rs
```diff
@@ -118,6 +120,26 @@ impl StatefulJob for FileDecryptorJob {
...
+							.keyslots
+							.iter()
+							.any(|k| k.content_salt == x.content_salt)
+					})
+					.map(|k| k.uuid)
+					.collect();
+
+				for key in keys.into_iter() {
+					ctx.library_ctx.key_manager.mount(key).await.ok();
+				}
```
Brendonovich, line 140: You should be able to move the entire iterator inside `in .. {`, so that `map` and `collect` aren't necessary.
Also if you're gonna use `key_manager` multiple times it's probs worth assigning to a variable at the top of the function.

Brendonovich (changes requested):
small pr good nice i like

Brendonovich (approved):

PR #566, opened 2023-02-09
[Fix] File unhighlightning

JulesGuesnon:
Hey!
While using the application it was kinda frustrating that you can't remove the highlight of a file once you clicked it. As it's the behavior of the native explorers, I implemented it

https://user-images.githubusercontent.com/36401048/217800481-325341ca-8895-4853-8e29-ccdf909e7bd9.mov

### DIFF
--- a/packages/interface/src/components/explorer/VirtualizedList.tsx
+++ b/packages/interface/src/components/explorer/VirtualizedList.tsx
@@ -121,13 +121,6 @@ export const VirtualizedList = memo(({ data, context, onScroll }: Props) => {
 				ref={scrollRef}
 				className="h-screen custom-scroll explorer-scroll"
 				onClick={(e) => {
-					if (
-						!scrollRef.current ||
-						(!(e.target as HTMLElement).classList.contains('file-row') &&
-							scrollRef.current !== e.target)
-					)
-						return;
-
 					getExplorerStore().selectedRowIndex = -1;
 				}}
 			>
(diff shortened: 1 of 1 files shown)


jamiepine (dismissed):
Great addition, the existing way was to click the file again, which was temporary until I or someone else added this, so thank you!

packages/interface/src/components/explorer/VirtualizedList.tsx
```diff
@@ -117,7 +117,20 @@ export const VirtualizedList = memo(({ data, context, onScroll }: Props) => {
...
+			<div
+				ref={scrollRef}
+				className="h-screen custom-scroll explorer-scroll"
+				onClick={(e) => {
+					if (
+						!scrollRef.current ||
+						(!(e.target as HTMLElement).classList.contains('file-row') &&
+							scrollRef.current !== e.target)
+					)
+						return;
```
Brendonovich, line 129: Could this check be avoided if the virtualizer's children had `stopPropagation` in their `onClick` handlers?

JulesGuesnon: I actually tried and it wasn't working, not sure why tho. Let me double check

JulesGuesnon: Ok I don't get why it wasn't working the first time, but it is now, so definitely shorter and more efficient to do it with `stopPropgation`

Brendonovich (approved):

PR #568, opened 2023-02-11
[ENG-363] Spacedrop UI + Misc Improvements

jamiepine:
As requested by Oscar this is the UI for the Spacedrop functionality. This branch will also include a bunch of misc improvements, as listed below: 

- New Object icons for encrypted and compressed
- Better error fallback screen
- Default to Inspector closed
- Replaced `.sdenc` extension with `.bytes`
- Add inspector button
- Slightly better 404 screen
- Added `_Layout.tsx` as a standard for screens
- Added `SearchInput` component
- Improved About Spacedrive screen
- Improved `Input` component sizes to align with `Button` component

Screenshot 2023-02-21 at 11 33 38 AM
[NOT FINAL - ITS A WIP]


packages/interface/src/screens/index.tsx
```diff
@@ -15,11 +15,12 @@ const routes: RouteProps[] = [
 	{ path: 'media', element: lazyEl(() => import('./Media')) },
 	{ path: 'spaces', element: lazyEl(() => import('./Spaces')) },
 	{ path: 'debug', element: lazyEl(() => import('./Debug')) },
+	{ path: 'spacedrop', element: lazyEl(() => import('./Spacedrop')) },
 	{ path: 'location/:id', element: lazyEl(() => import('./LocationExplorer')) },
 	{ path: 'tag/:id', element: lazyEl(() => import('./TagExplorer')) },
 	{
 		path: 'settings',
-		element: lazyEl(() => import('./settings/Layout')),
+		element: lazyEl(() => import('./settings/_Layout')),
```
Brendonovich, line 23: not a huge fan of `_Layout`, I think a nicer pattern would be for `./settings` to export the layout as a default, and `settingsScreens` be a named export. If we're going this far we also may as well switch to [`useRoutes`](https://reactrouter.com/en/main/hooks/use-routes) instead of the hybrid we're currently using.

Brendonovich:
@jamiepine hello sir

packages/interface/src/components/layout/Sidebar.tsx
```diff
@@ -399,7 +418,11 @@ function LibraryScopedSection() {
 			{!!tags.data?.length && (
 				<SidebarSection
 					name="Tags"
-					actionArea={<SidebarHeadingOptionsButton to="/settings/tags" />}
+					actionArea={
+						<NavLink to="/settings/tags">
```
Brendonovich, line 422: should be a relative path

Brendonovich (changes requested):

oscartbeaumont (dismissed):

core/src/object/fs/decrypt.rs
```diff
@@ -74,7 +74,7 @@ impl StatefulJob for FileDecryptorJob {
 			|| {
 				let mut path = info.fs_path.clone();
 				let extension = path.extension().map_or("decrypted", |ext| {
-					if ext == ".sdenc" {
+					if ext == ".bytes" {
```
brxken128, line 77: i quite like this change, very assertive as `.bytes` is extroadinarily generic

brxken128 (approved):
i quite like the change regarding encrypted file extensions, very assertive (in a great way) as `.bytes` is extroadinarily generic.

awesome work!

core/src/object/fs/decrypt.rs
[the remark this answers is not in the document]
Brendonovich: i'm pretty sure this shouldn't even be `.bytes` anyway
image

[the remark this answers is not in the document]
brxken128: it should be, jamie mentioned the change a while back iirc

Brendonovich (approved):

PR #575, opened 2023-02-16
[ENG-134] OSS Licenses and Projects

brxken128:
This PR adds the `deps-generator` binary, and a basic placeholder page for displaying all of our front/back-end dependencies in-app.

I removed the changes from CI as they weren't working well enough. We can always custom-parse and manually import reports from Fossa, which may be easier if we just update this for each release.

The UI is **_extremely_** basic, and this PR contains the minimum amount of data to make it work. I have attempted to lazy-load the data, I'm not too sure how correct the code is (but it works).

`deps-generator` usage:

* `deps-generator frontend c04897d902c9a55baf14d532f4590bc711f08c16 frontend-deps.json`
* `deps-generator backend Cargo.toml backend-deps.json`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -188,6 +188,12 @@ jobs:
       - name: Check core
         run: cargo check -p sd-core --release
 
+      - name: Compile frontend dependencies
+        run: cargo run --release -p deps-generator -- frontend $GITHUB_SHA frontend-deps.json
+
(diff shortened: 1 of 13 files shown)


fogodev (approved):
Just introduced Reqwest blocking feature, as it was the only place that needed to be async and as this executable doesn't need async extra performance, its nice to get away from it.

PR #580, opened 2023-02-25
[ENG-379] Explorer item resizer

jamiepine:
Adds the ability to resize items in the explorer grid view
- Added item resizer
- Fixed explorer store bug
- Refactored file image component
- Fixed explorer thumbnail rendering
- Improved list view, added more columns and column headings

Note: there is a bug where icons do not render in the FileIcon component, images, video and folders are fine, but custom icons require a forced re-render to show (you can trigger it using the context menu). This only happens on list view and in the inspector, grid view is fine.

https://user-images.githubusercontent.com/32987599/221331773-4cadb0af-3570-41d2-b89f-b7b30c2672fd.mov

Screenshot 2023-02-24 at 7 20 51 PM

### DIFF
--- a/packages/interface/src/components/explorer/FileColumns.tsx
+++ b/packages/interface/src/components/explorer/FileColumns.tsx
@@ -11,14 +11,14 @@ export function ensureIsColumns<T extends IColumn[]>(data: T) {
 	return data;
 }
 
-export const columns = ensureIsColumns([
-	{ column: 'Name', key: 'name', width: 280 } as const,
-	// { column: 'Size', key: 'size_in_bytes', width: 120 } as const,
-	{ column: 'Type', key: 'extension', width: 150 } as const,
-	{ column: 'Size', key: 'size', width: 100 } as const,
-	{ column: 'Date Created', key: 'date_created', width: 150 } as const,
-	{ column: 'Content ID', key: 'cas_id', width: 150 } as const
-]);
+export const columns = [
+	{ column: 'Name', key: 'name', width: 280 },
+	// { column: 'Size', key: 'size_in_bytes', width: 120 },
+	{ column: 'Type', key: 'extension', width: 150 },
+	{ column: 'Size', key: 'size', width: 100 },
+	{ column: 'Date Created', key: 'date_created', width: 150 },
+	{ column: 'Content ID', key: 'cas_id', width: 150 }
+] as const satisfies Readonly<IColumn[]>;
 
 export type ColumnKey = (typeof columns)[number]['key'];
 
--- a/packages/interface/src/components/explorer/FileColumns.tsx
+++ b/packages/interface/src/components/explorer/FileColumns.tsx
@@ -11,14 +11,14 @@ export function ensureIsColumns<T extends IColumn[]>(data: T) {
 	return data;
 }
 
-export const columns = [
-	{ column: 'Name', key: 'name', width: 280 },
-	// { column: 'Size', key: 'size_in_bytes', width: 120 },
-	{ column: 'Type', key: 'extension', width: 150 },
-	{ column: 'Size', key: 'size', width: 100 },
-	{ column: 'Date Created', key: 'date_created', width: 150 },
-	{ column: 'Content ID', key: 'cas_id', width: 150 }
-] as const satisfies IColumn[];
+export const columns = ensureIsColumns([
+	{ column: 'Name', key: 'name', width: 280 } as const,
+	// { column: 'Size', key: 'size_in_bytes', width: 120 } as const,
+	{ column: 'Type', key: 'extension', width: 150 } as const,
+	{ column: 'Size', key: 'size', width: 100 } as const,
+	{ column: 'Date Created', key: 'date_created', width: 150 } as const,
+	{ column: 'Content ID', key: 'cas_id', width: 150 } as const
+]);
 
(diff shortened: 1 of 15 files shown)


packages/interface/src/components/explorer/FileColumns.tsx
```diff
@@ -0,0 +1,42 @@
...
+}
+
+export const columns = ensureIsColumns([
+	{ column: 'Name', key: 'name', width: 280 } as const,
+	// { column: 'Size', key: 'size_in_bytes', width: 120 } as const,
+	{ column: 'Type', key: 'extension', width: 150 } as const,
+	{ column: 'Size', key: 'size', width: 100 } as const,
+	{ column: 'Date Created', key: 'date_created', width: 150 } as const,
+	{ column: 'Content ID', key: 'cas_id', width: 150 } as const
+]);
```
Brendonovich, line 21: `ensureIsCoumns` is redundant - we have `satisfies`
```suggestion
export const columns = [
	{ column: 'Name', key: 'name', width: 280 } as const,
	// { column: 'Size', key: 'size_in_bytes', width: 120 } as const,
	{ column: 'Type', key: 'extension', width: 150 } as const,
	{ column: 'Size', key: 'size', width: 100 } as const,
	{ column: 'Date Created', key: 'date_created', width: 150 } as const,
	{ column: 'Content ID', key: 'cas_id', width: 150 } as const
] as const satisfies IColumn[];
```

packages/interface/src/components/explorer/FileColumns.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+export function ListViewHeader() {
+	return (
+		<div
+			style={{ height: LIST_VIEW_HEADER_HEIGHT }}
+			className="mr-2 flex w-full flex-row rounded-lg border-2 border-transparent"
+		>
+			{columns.map((col) => (
+				<div
+					key={col.key}
```
Brendonovich, line 33: may as well key on `col.column`, `col.key` is kinda redundant

packages/interface/src/components/explorer/FileRow.tsx
```diff
@@ -11,57 +18,73 @@ interface Props extends HTMLAttributes<HTMLDivElement> {
...
-				<div
-					key={col.key}
-					className="table-body-cell flex items-center px-4 py-2 pr-2"
-					style={{ width: col.width }}
-				>
-					<RenderCell data={data} colKey={col.key} />
-				</div>
-			))}
-		</div>
+		<ExplorerItemContextMenu className="w-full" data={data}>
```
Brendonovich, line 21: I seem to recall us not wrapping `File*` stuff with individual context menus before for perf reasons - is this not the case anymore?

packages/interface/src/components/explorer/VirtualizedList.tsx
```diff
@@ -127,10 +107,11 @@ export const VirtualizedList = memo(({ data, context, onScroll }: Props) => {
 					ref={innerRef}
 					style={{
 						height: `${rowVirtualizer.getTotalSize()}px`,
-						marginTop: `${TOP_BAR_HEIGHT}px`
+						marginTop: `${TOP_BAR_HEIGHT + LIST_VIEW_HEADER_HEIGHT}px`
```
Brendonovich, line 110: I don't think these need to be strings, react should let them just be numbers

Brendonovich (changes requested):

packages/interface/src/components/explorer/FileRow.tsx
jamiepine: It doesn't seem to be an issue, but it was not changed in this PR, just renamed.

packages/interface/src/components/explorer/FileColumns.tsx
jamiepine: @Brendonovich the suggested change didn't work here, IColumn has a type error because its "readonly".

Brendonovich: Ah, use `satisfies Readonly<IColumn[]>` instead

Brendonovich (approved):

SusieInPhx:
Image Resizer Bug
powertoys windows 11 image resizer hides folder names

Power Toys was installed on my Windows 11 PC today.  I ran lots of pics thru it successfully until now.
Suddenly from the current folder down the line thru all subsequent folders, even through to the Recycle Bin, the list of file names all appear to be a 'l' or a '!'.  Subfolders are not plagued this way, pressing the arrow to open up the heirarchy looks normal..
Clicking on the folder itself shows the path the way it should, including the correct folder name.

It is as though the folder names are hidden until clicking on them.
image

brxken128:
> Image Resizer Bug
> powertoys windows 11 image resizer hides folder names
> 
> Power Toys was installed on my Windows 11 PC today.  I ran lots of pics thru it successfully until now.
> Suddenly from the current folder down the line thru all subsequent folders, even through to the Recycle Bin, the list of file names all appear to be a 'l' or a '!'.  Subfolders are not plagued this way, pressing the arrow to open up the heirarchy looks normal..
> Clicking on the folder itself shows the path the way it should, including the correct folder name.
> 
> It is as though the folder names are hidden until clicking on them.
> image
> 
> 
> 

I'm failing to see how this is relevant? Did changing some settings in PowerToys break our app (Spacedrive)?

PR #581, opened 2023-02-25
[ENG-380] Interface code structure improvement

Brendonovich:
After discussion with @jamiepine, it has been decided that @sd/interface will be getting overhauled thanks to new ideas made possible by fully adopting nested routing.

## App Directory

Taking a page out of Next 13's book, most code will now follow a specific structure inside interface/app (no src folder).

Files and folders will be organised in similarity to the app's routing structure, with camel-case files and folders being routes (index files can be an exception), and PascalCase files and folders being components.

## Dynamic Routes

Routes with dynamic parameters should also match how they look in the router, being a : followed by the name of the dynamic parameter.

## Nesting

camelCase folders will be groups of nested routes, and should contain an index file that exports something like:

`export const routes: RouteObject[] = []`
## Layouts

If a folder of routes has a layout that should be shared between all of the routes, it should be in a Layout.tsx file that is imported by the parent folder's index file and used as that route's element.

Layouts that are only used by a selection of a folder's routes should be named *Layout.tsx, but not Layout.tsx itself. Using them as wrapper components for the routes they apply to is acceptable, but the recommended approach is to use a [layout route](https://reactrouter.com/en/main/route/route#layout-routes).

## Non-Route Components
Non-Route components should be kept as close to the route they are used in as possible. Same-file or sibling file is preferred, but if used across multiple routes or nesting levels then they should be located only as high as necessary.

## Convention > Configuration
Unlike Next 13, none of these rules will be enforced by any sort of compiler or Vite plugin. They simply serve to make the codebase more uniform and approachable.

## No src Directory
It doesn't really do much so why have it

## Moving from packages/interface to interface

Being the central entrypoint of the app (like core is for the backend), Jamie wants to lift up interface and I'm not opposed to it.


interface/app/$libraryId/KeyManager/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+								unmountAll.mutate(null);
+								clearMasterPassword.mutate(null);
+							}}
+							variant="subtle"
+							className="text-ink-faint"
+						>
+							<Lock className="text-ink-faint h-4 w-4" />
+						</Button>
+						<ButtonLink
+							to={`/${library.uuid}/settings/overview`}
```
brxken128, line 45: This doesn't work for some reason (page not found), but it should be `/settings/keys` too.

brxken128 (changes requested):
Looks great and 99% of things are golden from what I've seen! Pretty happy with these changes as finding things will definitely be more intuitive.

The other issue I found is that the blue sidebar border(?) is cut off on the left/right sides - probably isn't a difficult fix and isn't major.

jamiepine (approved):
Great work, it has such a nice structure now.

PR #582, opened 2023-02-26
[ENG-376] Fix double libraries being created during onboarding

brxken128:
This PR fixes a bug where two libraries are created during onboarding in development mode. This happens as, since React 18, `useEffect` hooks are ran twice in dev (while strict mode is enabled) to ensure that state restoration is handled correctly.

This PR will have to remain a draft, as I'm not actually sure it works. Currently there seems to be a bug where after onboarding has been completed, a `useEffect` somewhere goes on an infinite loop and the app crashes until a reload. On the plus side, only one library is created so this PR functions as intended.

We'll hopefully know for sure what causes the aforementioned issue once #581 has been finished/merged. I have tested both this branch and `main`, and the bug is present in both. We can also possibly remove the `useEffect` from the library creation code entirely - I tried but couldn't tell if it worked or not, so I just stuck with this (as it _should_ work from what I've seen).


brxken128:
Tested with the changes from main - works great and onboarding is now buttery smooth.

interface/app/onboarding/creating-library.tsx
```diff
@@ -55,20 +55,25 @@ export default function OnboardingCreatingLibrary() {
...
-			if (debugState.enabled) {
-				setStatus(`You're running in development, this will take longer...`);
-			}
-		}, 5000);
-		return () => {
-			clearTimeout(timer);
-			clearTimeout(timer2);
-		};
+		if (created.current == false) {
+			created.current = true;
```
Brendonovich, line 62: I reckon an early return if it's already been ran would be cleaner

Brendonovich (approved):

PR #583, opened 2023-02-27
[ENG-227] Desktop app and landing page telemetry using Plausible Analytics

oscartbeaumont:
This PR makes some pretty significant changes, and you may find a summary below.

## Landing page:

* Our landing page now has a Plausible Analytics script attached, which is proxied through Vercel.

## Web, desktop and `@sd/client`:

* Telemetry sharing is tracked on a per-client basis, configured either during onboarding or via the client privacy settings (stored within a custom `valtio` store). This does persist even after our configuration directory has been deleted, so it may be wise to reset these to `false` at the start of onboarding, as to not infringe on any privacy.
* We are able to track app versions (currently hardcoded, but will eventually be in-sync with the core) and platform types (`mobile`, `web` and `desktop`)
* We have 2 new telemetry-related hooks: `usePlausibleEvent()` and `usePlausiblePageViewMonitor()`.
    * `usePlausibleEvent()` returns a callback and is used for submitting events to Plausible
    * the latter is used to watch the router's current URL - if a change is detected, a `pageview` event will be submitted via the aforementioned callback
* All telemetry-related hooks are in `@sd/client`, meaning they are not directly platform-specific (see Tasks below).
* Custom events are fired on certain actions:
    * library creation
    * library deletion
    * location creation
    * location deletion
    * tag creation
    * tag deletion
    * assigning a tag to an object
* Library UUIDs and other dynamic/unique items are removed or replaced from the router's path before it is submitted to Plausible - e.g. `/ac12ca58-261f-4e2b-a888-8696fc4bf1f9/location/26` becomes `/location`.
* Extensive documentation has been written to make it extremely easy to add new events.
* Zero PII is tracked. More in-depth information can be found here<sup>1</sup>.
* Telemetry sharing is disabled by default if the app is in development mode, regardless of the currently selected telemetry sharing option.
* The `DebugPopover` component was added into onboarding (only in debug mode)
* We can allow sending telemetry in debug mode via the aforementioned popover (provided telemetry sharing is also enabled for the client). Events fired in debug mode will be tagged accordingly in the Plausible dashboard.
* We can log all successfully submitted telemetry events to the console, with a toggle to enable/disable this functionality in the debug popover.
* All `ob_store` declarations were renamed to `obStore` for uniformity, and `obStore.shareTelemetryDataWithDevelopers` was renamed to `obStore.shareTelemetry`
* Validity checks were added to prevent multiple of the exact same events from firing consecutively (primarily an issue in debug mode). This has a pretty large issue though, as sometimes multiple consecutive requests are legitimate but only the first instance will be submitted. I've outlined a solution in the tasks below.

Tasks:
- [x] Allow telemetry to be enabled/disabled via the node settings page
- [ ] Automatically keep the telemetry version parameter in-sync with the core's version (currently it's hardcoded). This should be provided to Sentry too, and we should also only submit Sentry reports if either: the user presses "submit report" in the error page, or the user has telemetry sharing enabled and has agreed to send error reports. I'm not too sure regarding this though.
- [x] Allow telemetry sharing in development via the `debugState` store, but we should also attach this information to our events (similar to versioning). That way we can differentiate between debugging/testing events, and production events.
- [ ] Test if we can use a blank referrer instead of the default - it probably won't always be `localhost`
- [x] More custom events, such as `libraryDelete`, `tagCreate`, `tagDelete`, `locationCreate`, `locationDelete`
- [ ] Better path sanitizing which will hopefully make it easier to implement tracking on devices where different routing is in place (mainly mobile)
    - [ ] Replace Regex matching with a custom function, that either matches on the URL components or by receiving the dynamic data from the router
- [ ] Consider setting all telemetry settings to `false` at the start of onboarding, as even removing the entire Spacedrive directory will not remove valtio/localStorage entries. This applies for `debugState.shareTelemetry`, `debugState.telemetryLogging` and `telemetryState.shareTelemetry`. I'm not sure if anything else would benefit from being reset either (maybe if no libraries are present but valtio still has node settings, we could ask if the user wants to keep/reset?)
- [ ] Fix multiple events (of the same type) not firing consecutively due to them being regarded as duplicate events. We can use a relevant UUID, or something similar, to differentiate between both duplicate events and events of the same *type* which act upon a different object/tag/location/etc. No UUIDs will ever be sent to Plausible - they'll only be used for event differentiation and will be dropped beforehand.
- [ ] ~~We could possibly proxy our app tracking through our website too, but then we'd be reliant on our website/Vercel being up for app analytics to be recorded.~~

## Privacy considerations

<sup>1</sup>: Here is what a typical analytics submission looks like:

```json
{
    "n": "pageview",
    "u": "/overview",
    "d": "app.spacedrive.com",
    "r": "http://localhost:8001/",
    "w": 2560,
    "h": 0,
    "p": "{\"platform\":\"desktop\",\"version\":\"0.1.0\",\"debug\":true}"
}
```

It contains:
* the event type
* the URL/path, relative to the `domain` (for the `pageview` event this is equivalent to the router's **sanitized** path - otherwise it's `window.location.href`)
* the domain - it's not where the app is actually hosted, but a central subdomain where we can receive analytics
* the referrer, which is just `window.location.href`
* the screen width
* `h` seems to mark if the URL is in "hash mode" ([source](https://github.com/plausible/plausible-tracker/blob/c0b87d997d839938c23023d35bac0d6683635bbc/src/lib/request.ts#L67)), which should always be false (as we don't explicitly enable it)
* additional props:
    * platform type (mobile/desktop/web)
    * the core's version
    * whether or not the app is in debug mode

`window.location.href` for Tauri seems to be `http://localhost:8001/` and I can't imagine it's too dissimilar in most cases, although I'm still going to look at possibly removing the referrer as I can see some potential issues regarding to self hosting.

Along with this, the user agent is also sent to Plausible. That looks something like: `Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/16.0 Safari/605.1.15` (on my Gentoo Linux machine, using Tauri).

## Conclusion

It may take quite a lot of work to get the above missing tasks implemented, and this PR has been around for a while already - I don't mind getting this merged and then opening Linear issues to track further improvements.


brxken128:
This PR is very nearly good to go! Just need to move the telemetry sharing to node settings (instead of library settings).

Brendonovich (approved):

PR #585, opened 2023-03-02
Fix unique constraint on file path table

fogodev:
We have a unique constraint at `file_path` table that wasn't being respected due to `extension` column buing nullable, allowing for unwanted duplicated files


brxken128 (approved):
Looks good to me!

PR #588, opened 2023-03-03
Eng 388 remove pre commit hooks

Brendonovich:
They're annoying and we can run `pnpm lint` manually


brxken128 (approved):
Super happy about this 🙏

oscartbeaumont:
@Brendonovich can we please add `lint-staged` into the CI workflow so we don't end up with a mega lint commit again? Those big lint commits mess up the Git blame/history so much!

PR #589, opened 2023-03-03
Colocate native code + reorganise

Brendonovich:
All Swift code is now located inside a crate with Rust bindings to it, rather than the Rust bindings being somewhere else. This adds consistency since all rust can be found in `core` or any `crates/` folder, and `swift-rs` is only used inside crates that are macOS-only.
This should probably be accounted for in #527.


oscartbeaumont (approved):

brxken128 (approved):
I can't see anything particularly problematic, but a lot of it is Swift/mobile which I don't have experience with.

If MacOS has been tested with these changes and you're confident mobile won't break, I'd say you're good to go 🙏

PR #599, opened 2023-03-09
[ENG-371] Subpath shallow re indexing

fogodev:
Introducing shallow versions of Indexer, File Identifier and Thumbnailer jobs and also make them work with a given sub path instead of always being applied to the entire location. Also making Indexer Job reentrant, now we don't need to delete all `file_path`s on a full rescan of a location. 

Introducing the call of rspc mutation `quickRescan` to be called whenever the `Explorer` `path` or `location_id` changes in the frontend, this way whenever we open a directory, a light scan of this directory is performed, running shallow versions of indexer, file identifier and thumbnailer jobs. 

Making improvements in the `file_path`s IDs bookkeeping. Now it handles better multiple locations and libraries.

A bunch of other minor improvements and refactors.


oscartbeaumont (approved):

brxken128 (approved):
Looks good!

jamiepine (approved):
amazing

PR #600, opened 2023-03-09
[ENG-408] Dynamic Object icons

jamiepine:
Our icons should be glob imported where the name corresponds to {ObjectKind}_{Extension}.png

Screenshot 2023-03-03 at 4 39 24 PM


interface/app/$libraryId/Explorer/File/Thumb.tsx
```diff
@@ -1,18 +1,19 @@
...
-import Video from '@sd/assets/images/Video.png';
 import clsx from 'clsx';
 import { CSSProperties } from 'react';
 import { ExplorerItem } from '@sd/client';
-import { Folder } from '@sd/ui';
 import { usePlatform } from '~/util/Platform';
 import { getExplorerItemData } from '../util';
+import classes from './Thumb.module.scss';
+
+const icons = import.meta.glob('../../../../../packages/assets/icons/*.png', { eager: true });
```
Brendonovich, line 8: is `eager: true` a good idea? i'm not entirely sure what it means for an image to be eagerly imported (possibly all the URLs are loaded?) but usually you don't want to eagerly import so many files.

jamiepine: the promise stuff was giving me hell, this just works

Brendonovich: it works but idk if it's idea 😅 a bit of lazy and suspense should do the trick unless u don't care

jamiepine: I don't care just yet honestly, cause it doesn't appear to be a performance hit at all

Brendonovich: ehh it probs won't for desktop since loading doesn't take long at all, but it'll probably baloon bundle size which will be more noticeable on web.

Brendonovich (approved):

PR #601, opened 2023-03-09
Basic Spacedrop working with rspc!

oscartbeaumont:
It ain't pretty but it's working.

image

image

Note: The current system can only send files up to 255 bytes and they must be fully UTF-8. These limitations are because I haven't quite got the protocol down-packed for sending the file and wanted to get rspc hooked up. They will be easy limits to remove tomorrow or early next week.

Testing this system:
 - `echo "Hello World" > ~/Desktop/demo.txt`
 - `cargo run -p server` + `pnpm web dev` (or can use `pnpm desktop dev`)
 - In another terminal run: `export PORT=42069 && export DATA_DIR=./sdserver_data2 && cargo run -p server` - this creates the client we are gonna send the file to
 - Go to Spacedrop in the UI and use the ugly-looking UI I threw together with a path to the file you created in step one.

### DIFF
--- a/apps/desktop/package.json
+++ b/apps/desktop/package.json
@@ -21,7 +21,8 @@
 		"@tanstack/react-query": "^4.24.4",
 		"@tauri-apps/api": "1.2.0",
 		"react": "^18.2.0",
-		"react-dom": "^18.2.0"
+		"react-dom": "^18.2.0",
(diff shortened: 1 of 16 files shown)


Brendonovich:
w pr

PR #607, opened 2023-03-13
[ENG-429] Crypto organization/refactor

brxken128:
This PR heavily cleans up the crypto crate and the API for it. More things have been feature-gated too.

Some stuff has been re-exported in places that make more sense, and we have a new (awesome) macro to derive both the `Encryptor` and `Decryptor` types - they only have a few differences between eachother.

Things look a lot cleaner, simpler and easier to use now.


Brendonovich (dismissed):

brxken128:
Drafted as I forgot to update some of the API usage within the core.

Brendonovich (approved):

PR #619, opened 2023-03-18
[Desktop, UI] Fix library dropdown overflow & navigation

niikeec:
This PR fixes the issue with long library names in the `LibraryDropdown` menu and the `Manage Library` button navigates to `general` library settings rather than just settings. It also includes a Radix `DropdownMenu`, which uses classes and item internals from `ContextMenu`, that way changes are applied to both components. Could also work on replacing all the current Headless dropdowns with the new Radix one.

image

image

### DIFF
--- a/packages/ui/src/ContextMenu.tsx
+++ b/packages/ui/src/ContextMenu.tsx
@@ -8,21 +8,21 @@ interface Props extends RadixCM.MenuContentProps {
 	trigger: React.ReactNode;
 }
 
-const MENU_CLASSES = `
-  flex flex-col z-50
-  min-w-[8rem] px-1 py-0.5 my-2
-  text-left text-sm text-menu-ink
-  bg-menu cool-shadow
-	border border-menu-line
-  select-none cursor-default rounded-md
-`;
+export const contextMenuClasses = clsx(
+	'z-50 flex flex-col',
+	'my-2 min-w-[8rem] px-1 py-0.5',
+	'text-menu-ink text-left text-sm',
+	'bg-menu cool-shadow',
+	'border-menu-line border',
+	'cursor-default select-none rounded-md'
+);
 
-export const Root = ({ trigger, children, className, ...props }: PropsWithChildren<Props>) => {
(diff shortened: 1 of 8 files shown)


packages/ui/src/ContextMenu.tsx
```diff
@@ -1,62 +1,68 @@
...
 	);
 };
 
-export const Separator = () => (
-	<RadixCM.Separator className="border-b-menu-line pointer-events-none mx-2 border-0 border-b" />
+export const contextMenuSeparatorClassNames =
+	'border-b-menu-line pointer-events-none mx-2 my-1 border-0 border-b';
+
+const Separator = (props: { className?: string }) => (
+	<RadixCM.Separator className={clsx(contextMenuSeparatorClassNames, props.className)} />
```
Brendonovich, line 37: This can actually use the `tw` helper, something like 
```
tw(RadixCM.Separator)`...`
```

packages/ui/src/DropdownMenu.tsx
```diff
@@ -0,0 +1,162 @@
...
+import clsx from 'clsx';
+import React, { PropsWithChildren, Suspense, useCallback, useRef, useState } from 'react';
+import { Link } from 'react-router-dom';
+import {
+	ContextMenuItemProps,
+	ItemInternals,
+	contextMenuClasses,
+	contextMenuItemStyles,
+	contextMenuSeparatorClassNames,
+	contextSubMenuTriggerClassNames
```
Brendonovich, line 11: This is fine for now but I'll go through and get rid of the old dropdown eventually.

Brendonovich (changes requested):
Not a fan of all the duplication but I'll clean it up later. Make CI pass and you're good.

packages/ui/src/ContextMenu.tsx
Brendonovich: actually nvm the classes are exported so it won't work

Brendonovich (approved):

PR #627, opened 2023-03-20
Fix adding Locations that already have a `.spacedrive` file

### LINKED ISSUE #625: Error adding Location after removing the Desktop app data
### Describe the bug

After resetting the Desktop app data, if you try to add any Location you have previously added it fails with an error being thrown to the console:
image

This seems to be related to some backend logic about handling Locations that already have `.spacedrive` files in them:
https://github.com/spacedriveapp/spacedrive/blob/249d5b5534cb13a5ae0b6041760177411ea8c63d/core/src/location/mod.rs#L90-L100

I asked @fogodev about it, and he explained to me that the frontend must treat these errors and call the appropriate method, `locations.relink` and `locations.addLibrary` respectively, so the backend can successfully process these locations.

I have started to develop a solution for this issue [here](https://github.com/HeavenVolkoff/spacedrive/compare/add-location-button-indexer-rules...HeavenVolkoff:spacedrive:handle-errors-adding-locations). However, it depends on the changes in #620, so I will wait until that is merged to open a new PR.

### Reproduction

1. Follow the instructions at `CONTRIBUTING.md`
2. Execute Desktop app, create a Library and add a Location
3. Delete the app data (`${XDG_DATA_HOME}/spacedrive` on Linux)
4. Execute Desktop app again, create a Library
5. Attempt to add previous Location
6. No location is added, error is thrown in console

### Expected behavior

Location is correctly added and any issue with the current `.spacedrive` file is resolved

### Platform and versions

```Shell
7.29.3
cargo 1.68.0

$> uname -a
```

### Stack trace

```Shell
[Error] RSPCError: This location belongs to another library, must update .spacedrive file: (path: "/home/vitor/Workspace/DockerHills")
RSPCError — error.ts:154
from — error.ts:183
(anonymous function) — client.ts:183

	(anonymous function) (chunk-ELGFVIMM.js:1375)
	(anonymous function)
```

### Additional context

None

HeavenVolkoff:
Hello,

This PR addresses the issue of adding locations that already have a `.spacedrive` file, either because they belong to other Libraries or because their linked directory has been moved.

I implemented logic to handle both the `NeedRelink` and `AddLibraryToMetadata` errors raised by the backend when adding a new Location. When either of these errors occurs, a form error message is displayed to the user in `AddLocationDialog`, informing them that the location they are trying to add is either from another Library or needs to be relinked:
image
image

I decided to use the form's built-in error UI because the user's flow is similar to encountering a validation error when entering incorrect data, so I thought it would be more natural than a confirmation dialog.

If the user continues with adding the Location, the appropriate method is called on the backend: either `locations.relink` or `locations.addLibrary`. The `addLibrary` logic is working correctly, as far as I tested. However, the `relink` method has some caveats. Because I couldn't find a way to get the ID of the Location that will be relinked, it was not possible to update its indexer rules after relinking. Therefore, any rule changes made by the user in the form will be ignored during a relink. Additionally, I encountered some edge cases while testing this functionality with some unusual scenarios (such as restoring a `.spacedrive` file of a removed Location or moving a Location directory with the desktop app closed). The first resulted in an error, and the second caused the desktop app to stop opening until I restored the directory to its original location. As I'm unsure what the expected behavior is for these situations, I will open separate issues for them.

In addition to these changes, I also made the following improvements:
- Fixed a bug in `AddLocationDialog` that caused the indexer rules checkboxes not to track external changes made to the form state.
- Made minor changes to the errors raised by the backend when executing `locations.create`, giving them unique codes that the frontend can use to differentiate them.
- Abstracted the duplicated `openDirectoryPickerDialog` logic in `AddLocationButton` and `AddLocationDialog` into a function.
- Adjust `AddLocationDialog` style to match other Dialogs in the app

Closes #625.

### DIFF
--- a/interface/app/$libraryId/settings/library/locations/AddLocationDialog.tsx
+++ b/interface/app/$libraryId/settings/library/locations/AddLocationDialog.tsx
@@ -1,6 +1,6 @@
 import { RSPCError } from '@rspc/client';
-import { ChangeEvent, useEffect, useState } from 'react';
-import { Controller } from 'react-hook-form';
+import { ChangeEvent, useEffect, useRef, useState } from 'react';
+import { Controller, UseFormReturn } from 'react-hook-form';
 import { useLibraryMutation, useLibraryQuery } from '@sd/client';
 import { CheckBox, Dialog, UseDialogProps, useDialog } from '@sd/ui';
 import { Input, useZodForm, z } from '@sd/ui/src/forms';
@@ -9,9 +9,17 @@ import { Platform, usePlatform } from '~/util/Platform';
 
 const schema = z.object({ path: z.string(), indexer_rules_ids: z.array(z.number()) });
 
+type FormFieldValues<U> = U extends UseFormReturn<infer U> ? U : never;
+
 interface Props extends UseDialogProps {
 	path: string;
 }
+
+const LOCATION_ERROR_MESSAGES: Record<number, string | undefined> = {
+	// \u000A is a line break, It works with css white-space: pre-line
+	404: 'Location is already linked to another Library.\u000ADo you want to add it to this Library too?',
+	409: 'Location already present.\u000ADo you want to relink it?'
+};
 
 export const openDirectoryPickerDialog = async (platform: Platform): Promise<string> => {
 	if (!platform.openDirectoryPickerDialog) return '';
@@ -33,7 +41,6 @@ export const AddLocationDialog = (props: Props) => {
 	const relinkLocation = useLibraryMutation('locations.relink');
 	const indexerRulesList = useLibraryQuery(['locations.indexer_rules.list']);
 	const addLocationToLibrary = useLibraryMutation('locations.addLibrary');
-	const deleteLocationIndexerRule = useLibraryMutation('locations.indexer_rules.delete');
 
 	const form = useZodForm({
 		schema,
@@ -45,6 +52,7 @@ export const AddLocationDialog = (props: Props) => {
 
 	useEffect(() => {
 		const subscription = form.watch((_, { name }) => {
+			// Clear custom location error when user changes location path
 			if (name === 'path') {
 				form.clearErrors('root.serverError');
 				setExceptionCode(0);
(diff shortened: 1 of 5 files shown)


HeavenVolkoff:
While reviewing these changes, I found a bug in handling rspc errors other than 404 and 409. Currently, it does not show an alert dialog informing the user of the error and instead closes the dialog as if the location was added successfully, when in fact nothing has happened. I will send a patch to fix this later.

HeavenVolkoff:
Backend errors are now handled correctly. Any error, except those with a 500 status code, will be displayed as a form error message. Adding locations that belong to other libraries or relinking moved locations works both in the desktop app and the web version. To address issues with the old logic, I changed the error message for backend errors that require custom submit logic in the frontend, namely `NeedRelink` and `AddLibraryToMetadata`, to correctly identify them. This is not a final solution, but a temporary workaround until rspc supports sending custom data alongside error messages.

interface/app/$libraryId/settings/library/locations/AddLocationDialog.tsx
```diff
@@ -88,25 +179,37 @@ export const AddLocationDialog = (props: Props) => {
...
+												const checkBoxValue = Number.parseInt(checkBoxRef.value);
+												if (checkBoxRef.checked) {
+													field.onChange([...field.value, checkBoxValue]);
 												} else {
 													field.onChange(
-														field.value.filter((value) => value !== Number.parseInt(ref.value))
+														field.value.filter((fieldValue) => fieldValue !== checkBoxValue)
 													);
 												}
 											}}
```
Brendonovich, line 193: While we're here you may as well change this from HTML's `onChange` to Radix's `onCheckedChange`

interface/app/$libraryId/settings/library/locations/AddLocationDialog.tsx
```diff
@@ -26,59 +60,116 @@ export const AddLocationDialog = (props: Props) => {
...
+		const { watch, clearErrors } = form;
+		useEffect(() => {
+			// Clear custom remote error when user performs any change on the form
+			const subscription = watch(() => {
+				clearErrors(REMOTE_ERROR_FORM_FIELD);
+				setRemoteError(null);
+			});
+			return () => subscription.unsubscribe();
+		}, [watch, clearErrors]);
+	}
```
Brendonovich, line 75: I'd prefer if `form` wasn't destructed, since there's nothing wrong with putting it as a `useEffect` dependency. Then the block isn't needed either.

interface/app/$libraryId/settings/library/locations/AddLocationDialog.tsx
```diff
@@ -26,59 +60,116 @@ export const AddLocationDialog = (props: Props) => {
...
+
+	const onLocationSubmit = async ({ path, indexer_rules_ids }: FormFieldValues<typeof form>) => {
```
Brendonovich, line 77: Putting this in an extra variable is cool, but I'd suggest moving the `form.handleSubmit` up here too so that we can benefit from type inference for the function arguments.

Brendonovich (changes requested):
Nice work! I don't like the error UI but that can stay that way so that Jamie has incentive to fix it.

interface/app/$libraryId/settings/library/locations/AddLocationDialog.tsx
HeavenVolkoff: I see your point. My original reasoning was to try minimizing the update scope of the useEffect hook. However, looking back at it now, it seems like an unnecessary premature optimization.

Brendonovich (approved):

PR #628, opened 2023-03-20
rspc v3 syntax

oscartbeaumont:
This is an experimental branch using the alpha version of the rspc v3 syntax. It also comes with the new version of [Specta](https://github.com/oscartbeaumont/specta) which provides stupidly better error handling and a large number of bug fixes.

I spent the weekend working with @Brendonovich on this and we have got some stuff working which will greatly help with the maintainability of our rspc integration as it means we can remove all uses of private rspc APIs (of which we have a lot). Currently, it takes me 2-6 hours of updating Spacedrive when rspc does any non-minor changes which is pretty bad.

These changes also come with a shift in the mental model which should make more sense and allow us to make better use of middleware which will help to solve many challenges such as authentication as we get further into launching our product.

TODO:
 - [ ] Get frontend talking with Rust again
 - [ ] Fix RN transport

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2329,9 +2329,9 @@ dependencies = [
 
 [[package]]
 name = "futures-channel"
-version = "0.3.26"
+version = "0.3.27"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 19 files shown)


oscartbeaumont:
Closed in favor of #756 and a follow up PR with the new syntax stuff.

PR #633, opened 2023-03-26
[Desktop, UI] Input, color picker improvements

niikeec:
Changes made to Input:
- label `string` - Shows label above input field
- error `string | boolean`
	- `boolean` -  Applies "error" classes to the field
	- `string` - Applies "error" classes to the field and shows the error below the input
- icon `Icon | ReactNode` - Shows icon
- iconPosition `'left' | 'right'` `default = 'left'` - Determine the position of the icon
- right `ReactNode` - Add extra content on the right side of the input
- outerClassName `string` - Apply styles to the outer(parent) div
- labelClassName `string` - Apply styles to the label
- iconClassName `string` - Apply styles to the icon
- rightClassName `string` - Apply styles to the right content

These changes remove reusable label code and fix input buttons clashing with the input.

image
image

Changes made to the color picker:
- Changed react-colorful styles
- Added hex input
image

Changes made to useDialog:
- accepts a new `closeOnSubmit: boolean` prop `default = true` - whether to close the dialog after the form has been submitted
- returns a new `close: () => void` prop  - manually close the dialog

Changes made to create tag form:
- Prevent the form from submitting a tag with an empty name

image

### DIFF
--- a/apps/landing/src/components/DocsSidebar.tsx
+++ b/apps/landing/src/components/DocsSidebar.tsx
@@ -1,6 +1,5 @@
 import clsx from 'clsx';
-import { MagnifyingGlass } from 'phosphor-react';
-import { Input, SearchInput } from '@sd/ui';
+import { SearchInput } from '@sd/ui';
 import { DocsNavigation } from '../pages/docs/api';
 import config from '../pages/docs/docs';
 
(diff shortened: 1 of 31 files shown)


Brendonovich:
I appreciate what you've done here, but +-500 lines of assorted changes without any prior consultation isn't something I want to review or accept. We require linking to an issue to ensure that both reviewers and contributors are on the same page about what code should be written - we don't want anyone's time and effort being wasted.

For this PR in particular, there's a number of things I would have liked to discuss beforehand (not a fan of all the `className` props,  `label` + `error` should be on `FormField`) , and its scope (500 lines, 30 files) is way too large for anyone to _want_ to review.

I'm going to close this PR for the reasons I mentioned above, but it'd be great to discuss these changes individually in issues and get them merged in separate PRs!

niikeec:
Opened a discussion #640

PR #641, opened 2023-03-29
Improve the way react-devtools script is injected in dev mode

### LINKED ISSUE #624: Desktop app in dev mode fails to connect to React-Devtools
### Describe the bug

The desktop app currently includes a feature that attempts to connect to a remote React-Devtools session when it initializes in dev mode. You can see the code for it:
https://github.com/spacedriveapp/spacedrive/blob/249d5b5534cb13a5ae0b6041760177411ea8c63d/apps/desktop/src/index.tsx#L7-L12

However, in my environment, the React-Devtools hangs during initialization and I see the following errors in the app console:
image
image

This issue appears to be caused by initializing the DevTools connection after the body has already been processed and the main React library has already been loaded. According to this [issue](https://github.com/facebook/react/issues/24844#issuecomment-1177765718), the correct way to load the React-Devtools script is to inject it in the head early during the page parsing.

To avoid changing the loading order of the main app script too much, I implemented a simple Vite plugin that modifies the main index.html to inject the React-Devtools script. You can see the code for it [here](https://github.com/HeavenVolkoff/spacedrive/blob/dd603a8d328808da68d4e849e1a19f4a37f5aced/apps/desktop/vite.config.ts#L18-L31).

As far as I've tested, the plugin works fine, and the app successfully connects to the React-Devtools. I also added some logic to automatically open the DevTools together with the app in dev mode, but I'm not sure if this is a desired feature. If my solution is acceptable, I would be happy to open a PR.

### Reproduction

1. Follow the instructions at `CONTRIBUTING.md`
2. Execute the app alongside a React-Devtools instance
3. React-Devtools hangs and app shows error in console

### Expected behavior

React-Devtools successfully connects

### Platform and versions

```Shell
7.29.3
cargo 1.68.0

$> uname -a
```

### Stack trace

```Shell
[Error] TypeError: undefined is not a function (near '...hook.sub...')
	initBackend (localhost:16214)
	(anonymous function) (localhost:16909)
```

### Additional context

None

HeavenVolkoff:
Hello,

This implements the suggestion I proposed in #624. Basically, it is a simple vite plugin that injects the react-devtools script at the end of `index.html` head element when running in dev mode. ~Besides this, I also changed how the `pnpm desktop dev` command works, so it opens the react-devtools alongside the desktop app. If this behavior is undesired, please let me know, and I will promptly revert it.~

Closes #624


apps/desktop/package.json
```diff
@@ -6,7 +6,10 @@
 	"private": true,
 	"scripts": {
 		"vite": "vite",
-		"dev": "tauri dev",
+		"dev": "concurrently -kc auto 'pnpm:dev-*'",
+		"dev-vite": "vite --clearScreen=false --mode development",
+		"dev-tauri": "tauri dev",
+		"dev-react-devtools": "react-devtools",
```
Brendonovich, line 12: I don't think react devtools are used enough that they should be opened every time, `pnpm react-devtools` is small enough to write manually. That way `vite` can still stay under Tauri's control.

Brendonovich (changes requested):
thaaanks

apps/desktop/package.json
```diff
@@ -6,10 +6,9 @@
 	"private": true,
 	"scripts": {
 		"vite": "vite",
-		"dev": "concurrently -kc auto 'pnpm:dev-*'",
+		"dev": "tauri dev",
 		"dev-vite": "vite --clearScreen=false --mode development",
-		"dev-tauri": "tauri dev",
-		"dev-react-devtools": "react-devtools",
+		"react-devtools": "react-devtools",
```
Brendonovich, line 11: I don't think this is necessary - `pnpm react-devtools` should work anywhere inside the repo without being made a dedicated script (pnpm just goated like that).

CONTRIBUTING.md
```diff
@@ -51,7 +51,11 @@ To quickly run only the desktop app after `prep` you can use:
 
 - `pnpm desktop dev`
 
-To run the landing page
+To run the desktop app alongside react-devtools (MacOS and Linux):
+
+- `( pnpm desktop react-devtools & pnpm desktop dev; kill %1; )`
+
+To run the landing page:
```
Brendonovich, line 58: I think just saying something like 'React Devtools can be launched with `pnpm react-devtools`' would be better, since it lets the contributor decide how they want to launch both processes and isn't unix-specific.

Brendonovich (changes requested):

CONTRIBUTING.md
HeavenVolkoff: Fair enough, I will change it

apps/desktop/package.json
HeavenVolkoff: Oh, didn't know that, I will remove this line. This is my first time using pnpm, normally I use yarn. But using pnpm here is quite nice, it has some interesting features

CONTRIBUTING.md
HeavenVolkoff: Is this wording ok?

```
To quickly run only the desktop app after `prep` you can use:

- `pnpm desktop dev`

  If necessary, react-devtools can be launched using `pnpm react-devtools`.
  However, it must be executed before the desktop app for it to connect.
```

Brendonovich: Yea looks good

Brendonovich (approved):

PR #659, opened 2023-03-30
[Desktop] Add label & error to form field

niikeec:
Added label and error message to form fields as mentioned in #640, also replaced current labels with the new form field one. 

Example:
image

### DIFF
--- a/packages/ui/src/forms/FormField.tsx
+++ b/packages/ui/src/forms/FormField.tsx
@@ -4,18 +4,17 @@ import { useFormContext } from 'react-hook-form';
 export interface UseFormFieldProps extends PropsWithChildren {
 	name: string;
 	label?: string;
-	required?: boolean;
 	className?: string;
 }
 
 export const useFormField = <P extends UseFormFieldProps>(props: P) => {
-	const { name, label, required, className, ...otherProps } = props;
+	const { name, label, className, ...otherProps } = props;
 	const { formState, getFieldState } = useFormContext();
 	const state = getFieldState(props.name, formState);
 	const id = useId();
 
 	return {
-		formFieldProps: { id, name, label, required, className, error: state.error?.message },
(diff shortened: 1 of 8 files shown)


packages/ui/src/forms/FormField.tsx
```diff
@@ -3,26 +3,41 @@ import { useFormContext } from 'react-hook-form';
...
-	const _ = ctx.getFieldState(name);
+	const state = ctx.getFieldState(props.name);
 
-	return <>{children}</>;
+	return (
+		<div className={props.className}>
+			{props.label && (
+				<label htmlFor={props.id} className="mb-1 flex text-sm font-medium">
+					{props.label}
+					{props.required && <span className="ml-1 text-red-500">*</span>}
```
Brendonovich, line 34: I don't think this is necessary. Having `required` in the props declaration is also kinda yucky.

Brendonovich (changes requested):

packages/ui/src/forms/FormField.tsx
niikeec: That's true, but it is nice to tell the user which fields are required and which aren't before validation.

Thoughts on moving field state into useFormField hook, that way you don't have to call getFieldState for each field you want to have access to the error.

```ts
export const useFormField = <P extends UseFormFieldProps>(props: P) => {
	const { name, label, required, className, ...otherProps } = props;
	const { formState, getFieldState } = useFormContext();
	const state = getFieldState(props.name, formState);
	const id = useId();

	return {
		formFieldProps: { id, name, label, required, className, error: state.error?.message },
		childProps: { ...otherProps, id, name }
	};
};
```

Brendonovich: Moving the error state into the hook sounds good, but I stand by not using `*` for required fields. We can look at adding it another time, but I don't think it fits our design style right now.

Brendonovich (approved):

PR #660, opened 2023-03-30
added title icons for slots

vloe:
**what:**
added the ability to have title icons in slots. the icons I chose are of course not written in stone and can easily be changed.

**how it works:**
as far as I know are there three different slot types (note, info, warning). either a warning icon, or an info icon, is added to the slot title based of the slot type. if the slot type is not note, info or warning (which it probably shouldn't be), an info icon is still added.

**before imgs:**
bilde
bilde
bilde

**after imgs:**
bilde
bilde
bilde

**closes discussion:** 
https://github.com/spacedriveapp/spacedrive/discussions/654

### DIFF
--- a/apps/landing/src/style.scss
+++ b/apps/landing/src/style.scss
@@ -186,6 +186,12 @@ html {
 .slot-block-content {
 	@apply my-1 mx-0 mb-0 text-white;
 }
+.slot-block-title-icon {
+	@apply my-0 h-5 w-5 text-white;
+}
+.slot-block-title-container {
+	@apply flex flex-row items-center gap-x-1;
+}
 
 .doc-sidebar-button:hover,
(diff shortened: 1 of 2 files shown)


utkubakir:
First of all, thank you for your contribution!

Here are the few issues I noticed at a glance;
- Formatting (Use our prettier config please)
- Using svgs instead of phosphor icons. (It's best to use the icon library, whenever possible.)

vloe:
**update:**
- slot icons are now from phosphor. 
- changed the markdownParse file from ts to tsx, to be able to use the phosphor icons (imports are still correct so don't think this effect anything).
- using `renderToString` function from react to render the icons inside the string.

**slots now look like this:**

bilde
bilde
bilde

utkubakir (approved):

PR #675, opened 2023-04-04
[Desktop] Renamable files

### LINKED ISSUE #674: Missing file rename functionality
### Describe the bug

Currently, there's no way to rename a file from the explorer.

### Reproduction

_No response_

### Expected behavior

_No response_

### Platform and versions

```Shell
7.29.0
cargo 1.68.2 (6feb7c9cf 2023-03-26)
rustc 1.68.2 (9eb3afe9e 2023-03-27)
```

### Stack trace

_No response_

### Additional context

_No response_

niikeec:
This PR adds file renaming functionality, similar to how it works on macOS. With this, you can now rename files directly from within the app, without having to leave and rename them from the file manager.

The following keybindings have been added to enable the file renaming feature:
- Pressing the `Enter/Tab` key or losing focus will rename the file if the new name isn't empty or equal to its original.
- Pressing the `Escape` key will cancel the renaming process and revert the file name back to its original name.
- Pressing `ctrl/cmd + z` will revert the file name back to its original name and re-highlight the file name.

https://user-images.githubusercontent.com/43032218/229827352-6201b2d2-57b1-4b0f-8fc3-2d17131e8a79.mp4

Closes #674


core/src/api/files.rs
```diff
@@ -179,4 +180,32 @@ pub(crate) fn mount() -> RouterBuilder {
...
+					location.path.clone() + &args.new_file_name,
+				)
+				.await
+				{
+					Ok(_) => Ok(()),
+					Err(e) => Err(rspc::Error::new(
+						ErrorCode::Conflict,
+						format!("Failed to rename file: {}", e),
+					)),
+				}
```
fogodev, line 208: Please change this part to 
```rust
let location = find_location(&library, args.location_id)
    .select(location::select!({ path }))
    .exec()
    .await?
    .ok_or(LocationError::IdNotFound(args.location_id))?;

let location_path = Path::new(&location.path);
fs::rename(
    location_path.join(&args.file_name),
    location_path.join(&args.new_file_name),
)
.await
.map_err(|e| {
    rspc::Error::new(ErrorCode::Conflict, format!("Failed to rename file: {e}"))
})?;

invalidate_query!(library, "tags.getExplorerData");

Ok(())
```
This way, you just fetch the location path from database instead of the entire object, you avoid unnecessary string allocation and reallocation of `.clone()` and `+` operator, unnecessary match because the `Ok` case did nothing. And also refresh the explorer with the new file name.

fogodev (changes requested):
Just reviewed the Rust part, can't say much about TS, it isn't my strong suit lol

core/src/api/files.rs
niikeec: Thanks, this looks much cleaner. I'm not the strongest in Rust.

jamiepine: So I'm going to change this even further, I think we should supply just the `file_path_id` and the new name to the backend, I've been struggling to get Ericson's code above to work, at least on macOS, just getting path not found errors. Tons of debugging and the path is indeed correct, at least I think. Will play with it and commit a fix shortly.

fogodev: Yeah, there was a mistake on my part, I didn't note that `materialized_path`s were being passed in this route. We currently represent materialized paths on a specific struct `MaterializedPath` with all its inner logic, like removing a leading `/` before joining in a `Path`. But aside from that, I was thinking that this rename feature should use the same route for cutting and pasting, as cutting and pasting is just a rename nonetheless, and no data is actually moved in disk. Using https://github.com/spacedriveapp/spacedrive/blob/main/core/src/api/files.rs#L174 which internally dispatch a job renaming the file or directory.

niikeec:
Added an option so you can start renaming with Enter and disabled arrow selection while editing.

core/src/api/files.rs
niikeec: Pushed a quick fix. I wanted to use the file paths id initially, but I saw that it's not unique.

PR #683, opened 2023-04-07
ENG-449 - Keyboard support for Search & Component Extraction

ameer2468:
- TopBar.tsx had multiple components, so I extracted them into their own files.
- Searchbar did not support cmd/ctrl + F, so I added that + escape to unfocus the input.
- Removed PageLayout.DragChildren wrapper around SearchBar in spacedrop.tsx, this was causing a bug where the searchbar would disappear sometimes when you switch routes.Another reason is because PageLayout.Dragchildren isn't necessary and does not affect functionality.

Proposal down the line:

- We should implement a topbar that changes its tools/options based on context, instead of having to render SearchBar multiple times through the app, we can have it all in one place that is dynamic and would change based on the page the user is on (in reference to the tools/options provided)

Video of keyboard support:

https://user-images.githubusercontent.com/33054370/230666121-349ff252-4743-4608-a758-7b2dc34c3d84.mp4

Closes ENG-449

### DIFF
--- a/interface/app/$libraryId/Explorer/SearchBar.tsx
+++ b/interface/app/$libraryId/Explorer/SearchBar.tsx
@@ -1,0 +1,74 @@
+import clsx from 'clsx';
+import { ComponentProps, forwardRef, useEffect } from 'react';
+import { useForm } from 'react-hook-form';
+import { Input, Shortcut } from '@sd/ui';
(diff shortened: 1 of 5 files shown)


jamiepine (approved):
Good work, I like the proposal for the future, it should exist on all the main screens for sure.

jamiepine (approved):

PR #686, opened 2023-04-10
Add `MetadataManager` into sd-p2p

oscartbeaumont:
This adds the ability for Spacedrive to inform the P2P code about changes to the `PeerMetadata`. Once informed of an update a new Mdns broadcast will be queued ensuring that other clients on the local network are aware of the change.

A thing to note is that a restart is ALWAYS going to be required if the P2P keypair is updated but I think this is fair.

I haven't connected this to the rspc API's for updating peer metadata because they are part of #671 and not on the main yet. The `update_metadata` will need to be called when the config is changed, as I don't think integrating it `NodeConfigManager` is smart as that creates a circular dependency that will get messy quickly.

Usage:
 ```rs
let node = NodeContext { .. }; // Get this from context
node.p2p.update_metadata(&node.config).await;
```


oscartbeaumont:
@Brendonovich This will also fix the bug (or intentional design idk) where all the incoming sync events are sent to the user's first library.

PR #689, opened 2023-04-11
[ENG-291] Location settings

HeavenVolkoff:
This PR partially implements the `IndexerRuleEditor` component (still lacks the ability to create indexer rules, will be addressed in ENG-384), which allows for editing a location's indexer rules from its settings page. Additionally, I replaced the checkbox list of indexer rules in `AddLocationDialog` with `IndexerRuleEditor` for consistency. I also performed some minor clean-up in the location settings page and improved the error handling for some of its logic.

[indexer_rules.webm](https://user-images.githubusercontent.com/9082460/231084236-3dc40514-5050-437d-a2a5-cfa2fc4aed22.webm)

### DIFF
--- a/interface/app/$libraryId/settings/library/locations/$id.tsx
+++ b/interface/app/$libraryId/settings/library/locations/$id.tsx
@@ -57,25 +57,29 @@ export const Component = () => {
 	// fallback to 0 (which should always be an invalid location) when parsing fails
 	const locationId = (params.id ? Number(params.id) : 1) || 0;
 	useLibraryQuery(['locations.getById', locationId], {
-		onSettled: (data, error) => {
-			if (isFirstLoad && (!data || error)) {
-				navigate(-1);
+		onSettled: (data, error: Error | null) => {
+			if (isFirstLoad) {
+				if (!data && error == null) error = new Error('Failed to load location settings');
+
+				// Return to previous page when no data is available at first load
+				if (error) navigate(-1);
+				else setIsFirstLoad(false);
+			}
+
+			if (error) {
 				showAlertDialog({
 					title: 'Error',
 					value: 'Failed to load location settings'
 				});
-			} else {
-				setIsFirstLoad(false);
-				if (data && !isDirty) {
-					form.reset({
-						path: data.path,
-						name: data.name,
-						hidden: data.hidden,
-						indexerRulesIds: data.indexer_rules.map((i) => i.indexer_rule.id),
-						syncPreviewMedia: data.sync_preview_media,
-						generatePreviewMedia: data.generate_preview_media
-					});
-				}
+			} else if (data && (isFirstLoad || !isDirty)) {
+				form.reset({
+					path: data.path,
+					name: data.name,
+					hidden: data.hidden,
+					indexerRulesIds: data.indexer_rules.map((i) => i.indexer_rule.id),
+					syncPreviewMedia: data.sync_preview_media,
+					generatePreviewMedia: data.generate_preview_media
+				});
 			}
 		}
 	});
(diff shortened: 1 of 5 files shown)


interface/app/$libraryId/settings/library/locations/$id.tsx
```diff
@@ -55,20 +43,48 @@ export const Component = () => {
...
-			id: Number(id),
-			name: data.displayName,
-			sync_preview_media: data.syncPreviewMedia,
-			generate_preview_media: data.generatePreviewMedia,
-			hidden: data.hidden,
-			indexer_rules_ids: []
-		})
-	);
+	const { isDirty } = form.formState;
+	const locationId = Number.parseInt(params.id ?? '');
```
brxken128, line 47: Is there perhaps a better way to handle this?

interface/app/$libraryId/settings/library/locations/$id.tsx
```diff
@@ -1,52 +1,40 @@
...
-	displayName: z.string(),
-	localPath: z.string(),
-	indexer_rules_ids: z.array(z.string()),
-	generatePreviewMedia: z.boolean(),
-	syncPreviewMedia: z.boolean(),
-	hidden: z.boolean()
+	name: z.string(),
+	path: z.string(),
+	hidden: z.boolean(),
+	indexer_rules_ids: z.array(z.number()),
```
brxken128, line 23: I think TS names should be camel case for consistent formatting

brxken128 (commented):
Looking great! I'm not a TS expert but I left a few comments, feel free to ignore those if needed - I could very well be wrong about a few things.

Nice work!

interface/app/$libraryId/settings/library/locations/$id.tsx
```diff
@@ -55,20 +43,48 @@ export const Component = () => {
...
-	);
+	const { isDirty } = form.formState;
+	const locationId = Number.parseInt(params.id ?? '');
+	useLibraryQuery(['locations.getById', locationId], {
+		onError: (e) => {
+			showAlertDialog({
+				title: 'Error',
+				value: 'Failed to update location settings'
+			});
+			console.error({ e });
```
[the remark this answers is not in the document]
brxken128: as far as I know, there's very few errors that are logged to the browser console directly (especially if they're also shown to the user)

i also think RSPC shows the errors in the terminal window (if the error came from rust). i'm not sure on the criteria/specific scenarios that this can happen in, though

may be best to double check with another member of the team, i could be entirely wrong 🙏

interface/app/$libraryId/settings/library/locations/$id.tsx
brxken128: Rust does indeed use snake case, but the naming is being used within a form schema within TS so it should follow TS' rules imo

i can't quite remember off the top of my head, i do think we use camel case everywhere in TS but let me double check quickly

interface/app/$libraryId/settings/library/locations/$id.tsx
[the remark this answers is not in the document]
oscartbeaumont: rspc has the ability to a global error handling which is definitely something we will have to set up at some point.

In general, I think we should keep log messages exactly like this out of `main` because you can enable the rspc console logger (using the hidden menu when you click the version number in the bottom left) to get the same log but with even more information. If the log message has some special case or extra metadata they are definitely fine though.

interface/app/$libraryId/settings/library/locations/$id.tsx
utkubakir: We do have some inconsistencies but using camelCase in TS/React is the way, we changed some snake_cases before and I am pretty sure they will all be changed someday. It's basically cringe (As @Brendonovich would put it) to see snake_case on components.

brxken128 (approved):
Everything looks solid to me! Nice work 🙏

HeavenVolkoff:
Sorry @brxken128, I only realized that what I had committed wasn't working properly after requesting a new review.

jamiepine (approved):

PR #690, opened 2023-04-11
[ENG-286] Set default indexer rules for location

HeavenVolkoff:
This PR adds a new default indexer rule called `No OS protected`, which is intended to prevent the indexing of OS-protected files. To implement this rule, I had to modify the logic of the `AcceptFilesByGlob` and `RejectFilesByGlob` rules types to accept multiple globs and use a `GlobSet` for matching. The only potential issue with these changes is the `globset_from_globs` function, which creates a new GlobSet for each indexer rule during each directory/file check. While my testing didn't show any noticeable performance impact, the best approach would be to memoize it to execute only once for each indexer rule. However, I couldn't find a simple way to do this in Rust. If anyone has a solution, it would be greatly appreciated. Otherwise, I'll create a low-priority issue to address this later.

TODO: Still needs to test on macOS ~and Windows~

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -54,3 +54,15 @@ indent_size = 4
 indent_size = 4
 insert_final_newline = false
 trim_trailing_whitespace = true
+
+# SQL
+# https://www.sqlstyle.guide/
+[*.sql]
+indent_size = 4
+indent_style = space
+
+# Prisma
+# https://www.prisma.io/docs/reference/tools-and-interfaces/prisma-schema/data-model#formatting
+[*.prisma]
+indent_size = 4
+indent_style = space
--- a/interface/app/$libraryId/settings/library/locations/AddLocationDialog.tsx
+++ b/interface/app/$libraryId/settings/library/locations/AddLocationDialog.tsx
@@ -52,25 +52,17 @@ export const AddLocationDialog = ({ path, ...dialogProps }: Props) => {
 	const [remoteError, setRemoteError] = useState<null | RemoteErrorFormMessage>(null);
 
 	// This is required because indexRules is undefined on first render
-	const indexerRules = listIndexerRules.data;
-	const defaultValue = useMemo(
-		() => ({
-			path: path,
-			indexerRulesIds:
-				indexerRules?.filter((rule) => rule.default).map((rule) => rule.id) ?? []
-		}),
-		[indexerRules, path]
+	const indexerRulesIds = useMemo(
+		() => listIndexerRules.data?.filter((rule) => rule.default).map((rule) => rule.id) ?? [],
+		[listIndexerRules.data]
 	);
 
-	const form = useZodForm({
-		schema,
-		defaultValues: defaultValue
-	});
+	const form = useZodForm({ schema, defaultValues: { path, indexerRulesIds } });
 
 	useEffect(() => {
 		// Update form values when default value changes and the user hasn't made any changes
-		if (!form.formState.isDirty) form.reset(defaultValue);
-	}, [form, defaultValue]);
+		if (!form.formState.isDirty) form.reset({ path, indexerRulesIds });
+	}, [form, path, indexerRulesIds]);
 
 	useEffect(() => {
 		// TODO: Instead of clearing the error on every change, we should just validate with backend again
--- a/interface/app/$libraryId/settings/library/locations/AddLocationDialog.tsx
+++ b/interface/app/$libraryId/settings/library/locations/AddLocationDialog.tsx
@@ -1,9 +1,9 @@
 import { ErrorMessage } from '@hookform/error-message';
 import { RSPCError } from '@rspc/client';
 import { useQueryClient } from '@tanstack/react-query';
-import { useEffect, useState } from 'react';
+import { useEffect, useMemo, useState } from 'react';
 import { Controller } from 'react-hook-form';
-import { useLibraryMutation } from '@sd/client';
+import { useLibraryMutation, useLibraryQuery } from '@sd/client';
 import { Dialog, UseDialogProps, useDialog } from '@sd/ui';
 import { Input, useZodForm, z } from '@sd/ui/src/forms';
 import { showAlertDialog } from '~/components/AlertDialog';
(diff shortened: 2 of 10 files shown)


brxken128 (approved):
I'm uncertain regarding many of the actual rules themselves, but the Rust here is solid! Great job again haha

HeavenVolkoff:
Tested on Windows and Linux, by adding `root`/`C:` as locations, seems to be working fine. OS files were ignored.

interface/app/$libraryId/settings/library/locations/AddLocationDialog.tsx
```diff
@@ -41,25 +41,37 @@ export const openDirectoryPickerDialog = async (platform: Platform): Promise<nul
...
+		[indexRules, path]
+	);
+
 	const form = useZodForm({
 		schema,
-		defaultValues: {
-			path: props.path,
-			indexerRulesIds: []
-		}
+		defaultValues: useMemo(() => defaultValue, [defaultValue])
```
Brendonovich, line 65: this `useMemo` doesn't do anything - it'll update each time `defaultValue` changes

interface/app/$libraryId/settings/library/locations/AddLocationDialog.tsx
```diff
@@ -41,25 +41,37 @@ export const openDirectoryPickerDialog = async (platform: Platform): Promise<nul
...
 
-export const AddLocationDialog = (props: Props) => {
-	const dialog = useDialog(props);
+export const AddLocationDialog = ({ path, ...dialogProps }: Props) => {
+	const dialog = useDialog(dialogProps);
 	const platform = usePlatform();
 	const queryClient = useQueryClient();
 	const createLocation = useLibraryMutation('locations.create');
 	const relinkLocation = useLibraryMutation('locations.relink');
+	const { data: indexRules } = useLibraryQuery(['locations.indexer_rules.list'], {});
```
Brendonovich, line 50: We prefer to not destructure `useQuery` calls, it's not worth it especially once you do multiple queries in the same component.

interface/app/$libraryId/settings/library/locations/AddLocationDialog.tsx
```diff
@@ -41,25 +41,37 @@ export const openDirectoryPickerDialog = async (platform: Platform): Promise<nul
...
 	const [remoteError, setRemoteError] = useState<null | RemoteErrorFormMessage>(null);
 
+	// This is required because indexRules is undefined on first render
+	const defaultValue = useMemo(
+		() => ({
+			path: path,
+			indexerRulesIds: indexRules?.filter((rule) => rule.default).map((rule) => rule.id) ?? []
+		}),
+		[indexRules, path]
+	);
```
Brendonovich, line 61: I think `useLibraryQuery`'s `initialData` argument can be used instead of doing this

interface/app/$libraryId/settings/library/locations/AddLocationDialog.tsx
```diff
@@ -41,25 +41,37 @@ export const openDirectoryPickerDialog = async (platform: Platform): Promise<nul
...
-			indexerRulesIds: []
-		}
+		defaultValues: useMemo(() => defaultValue, [defaultValue])
 	});
 
 	useEffect(() => {
-		// Clear custom remote error when user performs any change on the form
+		// Update form values when default value changes and the user hasn't made any changes
+		if (!form.formState.isDirty) form.reset(defaultValue);
+	}, [form, defaultValue]);
```
Brendonovich, line 71: i really don't like this but we may be able to keep it

interface/app/$libraryId/settings/library/locations/IndexerRuleEditor.tsx
```diff
@@ -22,12 +22,12 @@ export function IndexerRuleEditor<T extends FieldType>({
 	field,
 	editable
 }: IndexerRuleEditorProps<T>) {
-	const listIndexerRules = useLibraryQuery(['locations.indexer_rules.list'], {});
+	const { data: indexRules } = useLibraryQuery(['locations.indexer_rules.list'], {});
```
Brendonovich, line 25: no destructure plz

Brendonovich (changes requested):

.editorconfig
```diff
@@ -54,3 +54,15 @@ indent_size = 4
...
+# https://www.sqlstyle.guide/
+[*.sql]
+indent_size = 4
+indent_style = space
+
+# Prisma
+# https://www.prisma.io/docs/reference/tools-and-interfaces/prisma-schema/data-model#formatting
+[*.prisma]
+indent_size = 4
+indent_style = space
```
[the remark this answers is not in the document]
HeavenVolkoff: The formatting, yes, but not the editor options. If you are also using the [EditorConfig VSCode Extension](https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig), which does change the editor options, you may experience inconsistent behavior between editing and formatting Prisma files in VSCode without this setting, as it will use the default `.editorconfig` settings (Tabs with size 2) for editing and the Prisma Extension settings for formatting.

interface/app/$libraryId/settings/library/locations/AddLocationDialog.tsx
HeavenVolkoff: You're right, I missed that. This is a left-over from a previous attempt that I made, will remove it.

interface/app/$libraryId/settings/library/locations/AddLocationDialog.tsx
HeavenVolkoff: This is because just changing the `defaultValue` doesn't update the current state of the form. So, without this, on the very first use of this dialog in the app, the component would load without any default indexer rule set.

interface/app/$libraryId/settings/library/locations/AddLocationDialog.tsx
HeavenVolkoff: I don't see how. The problem isn't that `useLibraryQuery` isn't fetching the data, but rather that `react-hook-form` ignores changes to the default value unless it's wrapped in a `useMemo`, and that it doesn't update the form state after its initial setup. Since this component is likely the first place in the app where a query for `locations.indexer_rules.list` will be requested, the value received during instantiation will be `undefined`, causing the form to be set up with an empty list for the default indexer rules. Later, when the corrected value is queried from the backend and the component is updated with the populated indexer rule list (probably during component mount), the form needs to be coaxed into updating its default state through the `useMemo` and `useEffect` hooks.

It's true that subsequent uses of this component won't have this issue since the value will already be cached. So, a potential solution would be to have a list of properties to be requested during app load so that their value is already cached before first use. However, I'm not sure if that's a better solution than simply handling this edge case here. What do you think?

Brendonovich: Ah I can understand memoising the `filter().map()` for efficiency, but I think memoising the entire `defaultValue` is overdoing it a little. I guess it just depends on whether we want `path` as a dependency in `useMemo` or the `useEffect`, where if we don't then we have to memoise the object separately but if we do then we can construct the object inline in both `defaultValues` and `reset`. Either is fine I guess.

HeavenVolkoff: That is fair, I will change it, so it only affects `defaultValues` instead of the whole object

Brendonovich (approved):

PR #691, opened 2023-04-11
Give FilePath a pub_id

Brendonovich:
I forgot to do this earlier so here it is. File paths are now properly shared resources, so they can be modified from anywhere. This decreases efficiency sadly, but allows for more possibilities.
I've added the `parent` and `chlidren` relations back to `FilePath` to allow fetching both local and pub ids for parent file paths, I'm not sure if this is a good idea though as it supposedly makes deletion very slow. @fogodev if the indexer logic can be changed so that the parent relation isn't required then it may be good to remove it.


brxken128 (approved):
Looks good!

core/prisma/schema.prisma
```diff
@@ -153,23 +155,21 @@ model FilePath {
     date_indexed  DateTime @default(now())
 
     // NOTE: this self relation for the file tree was causing SQLite to go to forever bed, disabling until workaround
```
fogodev, line 157: Is this problem solved? LMAO

fogodev:
Removed some pending stuff, like the `LastFilePathIdManager` as it isn't needed anymore. Also tried using uuids with `Uuid` type as much as I could, just converting to `Vec<u8>` in the last possible step before sync system or PCR, to avoid needless clones from heap, as internally `Uuid` uses `[u8; 16]` (which is just `u128` or `(u64, u64)`). Also some small stuff, like renaming the cursor at `file_identifier_job` to just account for the `file_path.id`, as `location.id` isn't needed anymore.

jamiepine (approved):
Hoping this isn't going to cause pain down the line, but for now to make things work I approve.

core/prisma/schema.prisma
Brendonovich: it is not but it is neccessary for now. if we can remove usage of `parent` from the indexer code then we can get rid of the relation again.

PR #696, opened 2023-04-12
ENG-460 Location creation fails for some writable directories

HeavenVolkoff:
This PR removes the check that verifies if a location's path is read-only during creation. This check caused false positives in cases where the users actually has write permissions to the path (I experienced it on both Windows and Linux). Instead, the creation of the `.spacedrive` metadata file can act as a better indicator of write access. Additionally, I implemented a rollback mechanism for location creation when either the `.spacedrive` file fails to be written or the location cannot be added to the library location manager.


brxken128 (approved):
Looks okay to me!

PR #700, opened 2023-04-12
[🏎️] : Full turbo 

thebinarymutant:
Goals of this PR:
1. Improve DX by taking advantage of Turborepo (Hopefully no changes in workflow will be required)
2. (Personal) To get an understanding of spacedrive packages and how they entwine

Closes #(issue)


Brendonovich:
We don't build each package to a `dist` folder by design. Instead, we use an approach where compilation is done by each app and each package is imported as source code rather than compiled code, kinda like [this blog post from Turborepo](https://turbo.build/blog/you-might-not-need-typescript-project-references).
The main reason we do this is for better hot reloading support, but it also means that we don't need turbo to spin up a build process per package while we work.
We could look into doing some optimisation of asset bundling since that tends to take a while, but for now I don't think it's necessary - building our Rust code takes wayyy longer anyway.

thebinarymutant:
Completely missed the [Internal Packages](https://turbo.build/repo/docs/handbook/sharing-code/internal-packages) part. I'm just going to take some time to get a better understanding and get back to you during the weekend. Thank you, @Brendonovich.

Brendonovich:
If you'd like to learn more about how the code is structured I'd suggest asking in our [Discord's development channel](https://discord.gg/PZxpa4vB), we're happy to answer questions and would rather do that than have to deny PRs. Hope to see you there!

PR #693, opened 2023-04-11
ENG-450: TopBar revamp and restructure

ameer2468:
**Edit: Initially converted to draft**

- Based on my convos with Utku, we'll be shifting to this layout: < > ----- Search ------ Icons, 
- To address responsiveness, the tools/options, they will be collapsed as a burger menu (next PR)

**This is a restructure of the Topbar component, it is a now a component that is global across the app. With this new change here is what we got**:

- Modify the tools/options based on routes, some routes may require different options
- Tools/options can be grouped, upon grouping, within the UI there's a separator.
- TopBar.tsx is now a much cleaner file, previously it was filled with repetitive code.
- I installed framer-motion in preparation for minimal animations within the topbar for some search bar related animations. It's a great library for UI movement.

**Minor adjustments not relevant to this PR**:

- ScreenHeading component adjusted slightly and is now being used within all pages, as some pages were not using them. Overview didn't have a title, so I added it in there too.

**Whats next**:

- For responsiveness, the tools/options need to collapse.

**A video representing these changes:**

https://user-images.githubusercontent.com/33054370/231208329-7ef63378-e772-417a-aa14-86452d12dfab.mp4

PR #701, opened 2023-04-12
[ENG-450] TopBar revamp and restructure

ameer2468:
Branch broke locally - so I had to recreate.

References: https://github.com/spacedriveapp/spacedrive/pull/693


interface/app/$libraryId/Explorer/SearchBar.tsx
```diff
@@ -4,72 +4,78 @@ import { useForm } from 'react-hook-form';
...
-				if (!dirtyFields.searchField) reset();
-			}
-		});
+export default forwardRef<HTMLInputElement, Props>((props, forwardedRef) => {
+	const {
+		register,
+		handleSubmit,
+		reset,
+		formState: { dirtyFields }
+	} = useForm();
```
Brendonovich, line 17: Ik it wasn't done in this PR but I don't think using RHF just for one search bar is necessary - a regular `useState` would be fine. Bringing in RHF should be done with our custom `useForm` and `<Form>` anyway.

jamiepine (approved):

PR #705, opened 2023-04-13
[ENG-458] - Top Bar Responsiveness

ameer2468:
This PR makes the top bar responsive.

**Notes**:

- Used side effects to get the width of the topbar for the sake of collapsing at the right time, rather than using media queries from tailwind for a better user experience.
- There's only 2 cases when it needs to collapse, when the tool count is >= 8 or  6 and at certain widths. No need to go further because thats the maximum amount you can resize the window to.
- Added check icon as an optional prop to top button component, that is the best way to apply this.
- Used the Popover component, rather than DropDown because it's styling is completely off and different, don't want to make un-necessary edits.

**Other notes**:

- Individual buttons are now supported for the Top bar, this allows them to have their own spacing and are away from each other + are rounded from both sides.

**Keep in mind another iteration within the code will take place in the next PR for API improvement.**

**Whats next**:

- Fix the popover component from going to the left corner when you resize the window, decided to do that in another PR so I don't go too far away from the focus of this change.

**Video below**:

https://user-images.githubusercontent.com/33054370/231892378-da10a143-d919-46f3-9fea-7140ca624d52.mp4

### DIFF
--- a/interface/app/$libraryId/Explorer/SearchBar.tsx
+++ b/interface/app/$libraryId/Explorer/SearchBar.tsx
@@ -32,11 +32,7 @@ export default forwardRef<HTMLInputElement, Props>((props, forwardedRef) => {
 	}, [forwardedRef]);
 
 	return (
-		<form
-			data-tauri-drag-region
-			onSubmit={() => {}}
-			className={`relative flex h-7 ${props.formClassName}`}
-		>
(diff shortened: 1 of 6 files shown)


HeavenVolkoff (changes requested):

jamiepine (approved):
Amazing, works great

PR #706, opened 2023-04-14
[ENG-469] Make Prettier and ESLint work together

HeavenVolkoff:
This PR resolves the conflicts between Prettier and ESLint, which were causing issues with indentation and Tailwind rules order in various parts of the project. Previously, either Prettier or ESLint were selectively disabled in certain places to avoid linting errors. With these changes, the rules have been standardized, enabling both tools to work together throughout the entire project. However, this change has resulted in the linter reporting errors in approximately 150 files. ~It's worth noting that this PR does not reformat the entire project. It only includes the necessary changes to allow the tools to function correctly together. I think the best approach to address the formatting issues is to gradually make the required changes as we modify the files in future PRs.~ (edit: It now includes the formatting changes, which were made by running `pnpm format` and `pnpm lint:fix`. This ensures that the CI remains functional for future PRs.)

Changes:

 - Added ESLint plugins for EditorConfig and Prettier.
 - ~Ensured that each linted subproject has a `.prettierrc.js` and `.eslintrc.js`~.
 - ~Pointed `.prettierrc.js` and `.eslintrc.js` of subprojects that use Tailwind to their corresponding `tailwind.config.js`~ (edit: Configured ESLint's 'tailwindcss/classnames-order' rule to the absolute path of the tailwind config).
 - Added a `.editorconfig` file to standardize basic formatting options across tools and editors.
 - Added `.gitattributes` to hide `pnpm-lock.yaml` in git diff output (very helpful for Git users in terminal).
 - Included EditorConfig in the recommended extensions for VSCode.
 - Replaced some instances of `pnpm exec <command>` with `pnpm <command>`.
 - Removed superfluous Tauri config for Linux.
 - Added a global ESLint config in `./package.json` to allow running ESLint from the project root (required for some advanced features of VSCode's ESLint extension).
 - Updated dependencies related to formatting and linting.
 - Fix a syntax error in `interface/types/declarations.d.ts`
 - Remove unused `markdown-link-check` dependency


apps/mobile/package.json
```diff
@@ -10,7 +10,7 @@
 		"ios": "expo run:ios",
 		"xcode": "open ios/spacedrive.xcworkspace",
 		"android-studio": "open -a '/Applications/Android Studio.app' ./android",
-		"lint": "eslint src",
+		"lint": "eslint . --cache",
```
utkubakir, line 13: This should be `src`, we don't need to lint anything outside of it?

package.json
```diff
@@ -52,5 +51,6 @@
 		"npm": "pnpm",
 		"yarn": "pnpm",
 		"node": ">=18.0.0"
-	}
+	},
+	"eslintConfig": {}
```
utkubakir, line 55: Do we need this?

interface/package.json
```diff
@@ -73,7 +73,7 @@
 		"@types/react-dom": "^18.0.6",
 		"@types/react-router-dom": "^5.3.3",
 		"@vitejs/plugin-react": "^1.3.1",
-		"prettier": "^2.7.1",
+		"prettier": "^2.8.7",
```
utkubakir, line 76: prettier can be moved to root package.json or `@sd/ui/package.json`

package.json
HeavenVolkoff: VSCode's ESLint extension requires a global ESLint config to run its task to scan the whole project. If there isn't one, it errors out...

apps/mobile/package.json
HeavenVolkoff: Some other subproject had files outside `src` that, IMHO, should be checked. So, I changed this for consistency, but I can revert it no problem.

HeavenVolkoff:
The mobile CI failed, but I'm not sure why. There shouldn't be any changes here that could have affected the app logic.

apps/mobile/package.json
utkubakir: Yeah, let's change this and the other instances where the src folder is present. I think the only big exception should be `interface`

utkubakir (dismissed):
Not sure why our ```printWidth``` is not being recognized but other than that, it looks good. I also fixed the eslint errors and ran `pnpm format`

HeavenVolkoff:
`printWidth` should be working correctly. However, it's worth noting that `.editorconfig` now standardizes `tabWidth` across tools and IDEs (requires an extension in VSCode). Currently, it's set to 4, which I based on the project's appearance from a clean clone. If this value isn't correct, I can change it easily.

fogodev (approved):

PR #709, opened 2023-04-14
[ENG-467]: Topbar improvements + show_at control

ameer2468:
**This is a great update to the topbar.**

**The changes**:

- Removed groupKeys and refactored to not needing them. 
- You can now control at what resolution to display certain tools.
- Whatever tool that doesn't have 'sm:flex' will be automatically displayed in the hidden/dropdown menu.
- We know now that at the smallest window resize you could have up to 4 tools showing, the rest is within the dropdown menu, hence why I refactored the side effect (useLayoutEffect) and used window sizing instead.

**Video**:

https://user-images.githubusercontent.com/33054370/232164117-430ecbbf-bc42-4bf0-9ea1-247d29ada207.mp4


interface/hooks/useToolBarOptions.tsx
```diff
@@ -26,18 +26,27 @@ export type RoutePaths =
...
+export type show_at_resolution = 'sm:flex' | 'md:flex' | 'lg:flex' | 'xl:flex' | '2xl:flex';
 
+export interface ToolOption {
+	icon: JSX.Element;
+	onClick?: () => void;
+	individual?: boolean;
+	toolTipLabel: string;
+	topBarActive?: boolean;
+	popOverComponent?: JSX.Element;
+	show_at_resolution: show_at_resolution;
```
Brendonovich, line 45: `camelCase` for properties plz

interface/hooks/useToolBarOptions.tsx
```diff
@@ -26,18 +26,27 @@ export type RoutePaths =
...
 
-export type groupKeys = 'groupOne' | 'groupTwo' | 'groupThree' | 'groupFour' | 'groupFive';
+// string concatination is incorrect usage of Tailwind - the complete and correct way is to use the full string
+// this is why I added :flex to the end of each string
+//sm is  640px
+//md is 768px
+//lg is 1024px
+//xl is 1280px
+//2xl is 1536px
+export type show_at_resolution = 'sm:flex' | 'md:flex' | 'lg:flex' | 'xl:flex' | '2xl:flex';
```
Brendonovich, line 36: `PascalCase` for types plz

interface/hooks/useToolBarOptions.tsx
```diff
@@ -46,110 +55,115 @@ export const TOP_BAR_ICON_STYLE = 'm-0.5 w-5 h-5 text-ink-dull';
 export const useToolBarRouteOptions = () => {
 	const store = useExplorerStore();
 
-	const toolBarRouteOptions: Record<RoutePaths, ToolOptions> = {
+	const toolBarRouteOptions: Record<RoutePaths, { options: ToolOption[][] }> = {
```
Brendonovich, line 58: I wonder if it would be better for each route/layout to define its own toolbar options and provide them via a combination of layout components and context, rather than defining them all here where the string names of the routes have to match. Not important for this PR but might be a good idea going forward.

interface/app/$libraryId/Explorer/TopBar.tsx
```diff
@@ -16,44 +16,35 @@ export default () => {
 	const navigate = useNavigate();
 	const topBarRef = useRef<HTMLDivElement>(null);
 	const searchRef = useRef<HTMLInputElement>(null);
+	const toolsRef = useRef<HTMLDivElement>(null);
```
Brendonovich, line 19: I'm pretty sure all of these are unnecessary now

interface/app/$libraryId/Explorer/TopBar.tsx
```diff
@@ -69,85 +60,84 @@ export default () => {
...
+									: index === 0
+									? 'left'
+									: index === group.length - 1
+									? 'right'
+									: 'none';
+								return (
+									<div
+										data-tauri-drag-region
+										key={toolTipLabel}
+										className={`hidden ${show_at_resolution} items-center ${
```
Brendonovich, line 93: We prefer to use `clsx` to concatenate classes instead of template literals

Brendonovich (changes requested):
Seeing as `TopBar` is now a top-level component it should probably be moved to `app/$libraryId` in a future PR, along with moving `useToolbarOptions` into `TopBar.tsx` and defining route-specific top bar info alongside the routes rather than at a top level.

Also cmd+shift+5 is a great tool for screen recording on Mac that doesn't require firing up OBS 😄

interface/hooks/useToolBarOptions.tsx
ameer2468: It is a good idea, child components setup in a certain way would provide the TopBar (parent) with what it would need in order to function.

Brendonovich (approved):

PR #710, opened 2023-04-14
[ENG-461] Location creation fails for '/' (root) and 'C:' ([A-Z] drives)

HeavenVolkoff:
This PR fixes a possible error when adding a location that could occur when the core fails to extract a name from its accompanying path. Additionally, I improved the location's name extraction logic by using [`localize_name`](https://docs.rs/normpath/1.1.1/normpath/trait.PathExt.html#tymethod.localize_name) and added an `Unknown` value as the default when all else fails. Besides this, I fixed a possible panic that could occur when the Indexer job fails to retrieve the created/modified date of a path. To solve this, I implemented a `MetadataExt` trait with `created_or_now`/`modified_or_now` that returns the current time when the path time isn't available.

### DIFF
--- a/core/src/location/mod.rs
+++ b/core/src/location/mod.rs
@@ -17,11 +17,11 @@ use crate::{
 
 use std::{
 	collections::HashSet,
-	ffi::OsStr,
-	path::{Path, PathBuf},
+	path::{Component, Path, PathBuf},
 };
 
 use futures::future::TryFutureExt;
+use normpath::PathExt;
 use prisma_client_rust::QueryError;
 use rspc::Type;
 use serde::Deserialize;
@@ -454,11 +454,27 @@ async fn create_location(
 
 	let location_path = location_path.as_ref();
 
-	let name = location_path
-		.file_name()
-		.and_then(OsStr::to_str)
-		.map(str::to_string)
-		.unwrap();
+	let name = location_path.normalize().map_or_else(
+		|_| {
+			location_path
+				.components()
+				.next()
+				.and_then(|component| match component {
+					Component::Prefix(component) => {
+						let prefix = component.as_os_str().to_string_lossy().to_string();
+						Some(if prefix == "/" {
+							"Root".to_string()
+						} else {
+							prefix
+						})
+					}
+					_ => None,
+				})
+				.get_or_insert("Unknown".to_string())
+				.to_owned()
+		},
+		|p| p.localize_name().to_string_lossy().to_string(),
+	);
 
 	let path = location_path
 		.to_str()
--- a/core/src/location/mod.rs
+++ b/core/src/location/mod.rs
@@ -452,34 +452,62 @@ async fn create_location(
 ) -> Result<location_with_indexer_rules::Data, LocationError> {
 	let Library { db, sync, .. } = &library;
 
-	let location_path = location_path.as_ref();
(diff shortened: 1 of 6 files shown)


HeavenVolkoff:
Creating a location in `root` or `C:` is still not possible due to permission errors that cause the indexer job to panic. However, when paired with #690, it can work as the paths with permission issues are ignored. Therefore, I believe this is ready for review since it solves the original problem of creating a root location, which was the failure to name the location. The indexer panics should be addressed in another issue/PR.

core/src/location/mod.rs
```diff
@@ -493,11 +493,27 @@ async fn create_location(
...
+				})
+				.get_or_insert("Unknown".to_string())
+				.to_owned()
+		},
+		|p| p.localize_name().to_string_lossy().to_string(),
+	);
+
+	if name == "/" {
+		name = "Root".to_string()
+	}
```
Brendonovich, line 516: Maybe `name` does end up as `/` on Windows, but would it be better to use `"Root"` if `location_path` has a certain length for cross-platform compatibility?

core/src/location/mod.rs
```diff
@@ -493,11 +493,27 @@ async fn create_location(
...
+					Component::Prefix(component) => {
+						Some(component.as_os_str().to_string_lossy().to_string())
+					}
+					_ => None,
+				})
+				.get_or_insert("Unknown".to_string())
+				.to_owned()
+		},
+		|p| p.localize_name().to_string_lossy().to_string(),
+	);
```
Brendonovich, line 512: Is this necessary over just doing `unwrap_or_else(|| "Unknown".to_string())`?

Brendonovich (changes requested):

core/src/location/mod.rs
HeavenVolkoff: You're right, I didn't consider that. In my opinion, I don't think `Root` is a good name for Windows because Windows users aren't accustomed to the notion of a root drive. Windows treats individual drives as their own root and assigns letters to them. Even for macOS and Linux users who have a Windows node with Spacedrive running on it, `Root` is not a good alternative because they will also expect that the `C:` drive will always show up as `C:`. I think the solution here is to check if the path given to `create_location` has a `/` prefix on Windows, and replace it with the same path under `C:`. This would also simplify the `No OS protected indexer` rule in #690 as I would not need to handle the '/' cases for Windows. What do you think?

HeavenVolkoff:
I had a discussion with @fogodev about how to handle Windows paths for new locations. We have both agreed that normalizing the path during its creation would be the best approach. Currently, the file watcher logic already assumes that each file has a path that follows the OS-specific format. So, it would be inconvenient that any future changes, or any other area that may deal with paths, to be required to account for Unix paths on Windows. Even more so, considering no user input for file paths, excluding manual entries, will return Unix paths on Windows. Therefore, I have included the logic to normalize paths on Windows in my latest commit. The normalization is done by indirectly calling `GetFullPathNameW`, which converts Unix paths to their Windows counterparts, adds the drive letter, and resolves any relative component of the path. Please note that this method only deals with the path and not the file system, and it doesn't resolve symlinks, which is consistent with our current Unix behavior.

Brendonovich (approved):

PR #711, opened 2023-04-17
add updater pub key

Brendonovich:
I have the private key atm, will add to gh actions as `TAURI_PRIVATE_KEY`


brxken128 (dismissed):
Looks good. Are we going the statically hosted JSON route?

PR #713, opened 2023-04-17
[ENG-468]: Toggle Popover off on window resize

ameer2468:
**This PR toggles the Popover off when the user resizes window, this should be the normal behaviour from a UI point of view.** 

- Also fixes a bug where the popover would be moving to left the corner, because it's parent element would be hidden for different screen sizes.
- Used 2 refs, for the button itself and the popover to guarantee the functionality of this expected behaviour.

### DIFF
--- a/packages/ui/src/Popover.tsx
+++ b/packages/ui/src/Popover.tsx
@@ -31,7 +31,7 @@ export const Popover = ({ trigger, children, disabled, className, ...props }: Pr
 			window.removeEventListener('click', handleClickOutside);
 			window.removeEventListener('touchstart', handleClickOutside);
 		};
-	}, [open]);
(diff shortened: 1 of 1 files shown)


Brendonovich:
Should this be normal behaviour? Looking at other apps i'm not sure - Discord, Slack and Linear's popovers don't seem to hide when you resize the window. If you've got counter-examples I'd be happy to see them tho.

Brendonovich:
That's fair, but they also use native UIs and not web UIs - keep in mind Spacedrive can an also be a website.
I'm not a UX guy so I'm honestly not sure whether we should honour native or web behaviour. Might be a question for @jamiepine

utkubakir (approved):

PR #715, opened 2023-04-17
[ENG-475] Fix list view scroller

niikeec:
This PR fixes the scroller behavior in the explorer list view


Brendonovich (approved):
i'd like to not use absolute positioning but if we must then so be it 😞

HeavenVolkoff (approved):

PR #724, opened 2023-04-19
[ENG]-487: Toolbar now uses Context API & Portal

ameer2468:
The topbar/toolbar has now been updated to use Context API & Portals.


interface/app/$libraryId/TopBar/ToolBarProvider.tsx
```diff
@@ -0,0 +1,48 @@
...
+	icon: JSX.Element;
+	onClick?: () => void;
+	individual?: boolean;
+	toolTipLabel: string;
+	topBarActive?: boolean;
+	popOverComponent?: JSX.Element;
+	showAtResolution: ShowAtResolution;
+}
+
+export type ShowAtResolution = 'sm:flex' | 'md:flex' | 'lg:flex' | 'xl:flex' | '2xl:flex';
```
utkubakir, line 20: Can we remove `:flex` from this?

interface/app/$libraryId/media.tsx
```diff
@@ -1,5 +1,11 @@
 import { ScreenHeading } from '@sd/ui';
+import { ToolOption } from './TopBar/ToolBarProvider';
+import { useToolBar } from './TopBar/useToolBar';
 
 export const Component = () => {
+	const toolBarOptions: { options: ToolOption[][] } = {
+		options: [[]]
+	};
+	useToolBar(toolBarOptions);
```
utkubakir, line 9: Why not inline these? It would still be typed and 1 less import is good.

interface/app/$libraryId/TopBar/ToolBarProvider.tsx
ameer2468: Tailwind would have compilation issues - you cannot do something like `${resolutionhere}:flex` - it needs to be the complete class.

interface/app/$libraryId/media.tsx
ameer2468: Valid point, ill do this.

interface/app/$libraryId/Layout/index.tsx
```diff
@@ -50,10 +54,13 @@ const Layout = () => {
 			<div className="relative flex w-full overflow-hidden">
 				{library ? (
 					<LibraryContextProvider library={library}>
-						<Suspense fallback={<div className="h-screen w-screen bg-app" />}>
-							<Outlet />
-						</Suspense>
-						<QuickPreview libraryUuid={library.uuid} />
+						<TopBarContext.Provider value={{ topBarChildrenRef }}>
+							<TopBar ref={topBarChildrenRef} />
```
Brendonovich, line 58: Don't put `TopBar` here just yet, I need to rework some of the routing first to make it work

interface/app/$libraryId/TopBar/index.tsx
```diff
@@ -1,45 +1,35 @@
...
 
-	useLayoutEffect(() => {
-		const handleResize = () => {
-			setWindowSize(window.innerWidth);
-		};
-		window.addEventListener('resize', handleResize);
-		handleResize();
-		return () => window.removeEventListener('resize', handleResize);
-	}, []);
+const TopBar = forwardRef<HTMLDivElement, Props>((props, ref) => {
```
Brendonovich, line 26: I think you're meant to use `ref` from `forwardRef` here rather than from `props`

interface/app/$libraryId/media.tsx
```diff
@@ -1,5 +1,11 @@
 import { ScreenHeading } from '@sd/ui';
+import TopBarChildren from './TopBar/TopBarChildren';
 
 export const Component = () => {
-	return <ScreenHeading>Media</ScreenHeading>;
+	return (
+		<>
+			<TopBarChildren toolOptions={[[]]} />
```
Brendonovich, line 7: Ideally `TopBarChildren` wouldn't be needed if no options need to be rendered

Brendonovich (changes requested):
Great work on this!

Brendonovich (approved):

PR #725, opened 2023-04-19
Misc UI fixes

HeavenVolkoff:
This is a PR with simple fixes for some UI problems I found while working on #723:
 - Fix for the unused title prop in `AlertDialog` (allowing it to now show the given title).
 - ~Fix for a runtime React warning about a `forwardRef` not using the given ref.~
 - Remove `console.log` from `OptionsPanel.tsx`
 - Replace `invalidateQuery` with refresh for `locations.list` in `AddLocationDialog`
 - Lint:fix Select.tsx
 - Update turbo to fix https://github.com/vercel/turbo/issues/3598

### DIFF
--- a/interface/app/$libraryId/Explorer/SearchBar.tsx
+++ b/interface/app/$libraryId/Explorer/SearchBar.tsx
@@ -1,5 +1,5 @@
 import clsx from 'clsx';
-import { ComponentPropsWithRef, forwardRef, useEffect, useImperativeHandle } from 'react';
+import { ComponentPropsWithRef, forwardRef, useEffect } from 'react';
 import { useRef, useState } from 'react';
 import { Input, Shortcut } from '@sd/ui';
 import { useOperatingSystem } from '~/hooks/useOperatingSystem';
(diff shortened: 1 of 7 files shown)


interface/app/$libraryId/Explorer/SearchBar.tsx
```diff
@@ -8,12 +8,14 @@ interface Props extends ComponentPropsWithRef<'input'> {
 	formClassName?: string;
 }
 
-export default forwardRef<HTMLInputElement, Props>((props) => {
+export default forwardRef<HTMLInputElement, Props>((props, ref) => {
```
Brendonovich, line 11: we don't actually need this ref, just remove it entirely

HeavenVolkoff: I know, but React keeps throwing warnings in the console due to it not being assigned, and if someone, for some reason, decides to do `<SearchBar ref={...}>` in the future it wouldn't work without this.
image

Brendonovich: If someone needs the ref in the future they can add it themselves, for now I'm confident that we're only ever going to be using the SearchBar component once and we won't need its ref.

Brendonovich (approved):

PR #726, opened 2023-04-19
[Eng 495] Better job report fields

fogodev:
This PR introduces the `action` field on jobs to gather related jobs, like on a scan location, that we have to run Indexer > File Identifier > Thumbnailer jobs, in this specific order. 

I also had to change how jobs hierarchy were represented, now when you queue multiple jobs, all queued jobs will be children of the first job, this way we have a better representation for the `action` info. 

This `action` column have the format `{action string received on jobs dispatch}(-{order})*`, where the string part is something like `"scan_location"`, and the optional order part appears just on children jobs, being in the range `1..`, representing its position on queue after the parent job.

Aside from the action stuff, this PR also introduces the notion of background jobs, which is useful to now show on UI the shallow versions of indexer, file identifier and thumbnailer jobs for example. This fields doesn't exist in the database, it's totally derived from each job, as it's just some static data when the job is implemented. It was implemented as a const already defined as `false` in the `StatefulJob` trait, and can be override to `true` on demand, when a new job implement this trait.


core/prisma/schema.prisma
```diff
@@ -368,14 +368,16 @@ model ObjectInSpace {
...
 
 model Job {
-    id       Bytes  @id
-    name     String
-    node_id  Int
-    action   Int
+    id      Bytes   @id
+    name    String
+    node_id Int
+    action  String? // Will be composed of "{action_description}_{uuid_of_parent_job}"
```
Brendonovich, line 374: We already have the parent job id - why does it need to be stored here?

Brendonovich (changes requested):

core/prisma/schema.prisma
fogodev: I introduced the `parent_id` embedded here to enable the frontend to differentiate between related jobs for each action in a easier way. But I agree that just just using `"{action}(_{order})*"` is enough, has in the frontend you could fetch actions without order to get the parent and then use its `id` to get the children jobs in the same action.

Brendonovich (approved):

PR #731, opened 2023-04-21
[ENG-489] FileThumb has incorrect dimensions for some thumbnails

HeavenVolkoff:
This PR is a rework of the `FileThumb` component, with multiple bug fixes:

 - Fixed tall and long images being displayed with incorrect dimensions (zoomed)
 - Fixed application crashes when selecting any non-image file while the inspector was open (at least on Linux)
 - Implemented video rendering when requiring a `FileThumb` with `loadOriginal`
 - Fixed the inspector's container cropping the `FileThumb` when selecting tall images
 - Correctly implemented `size=0` and allowed for `size=null` in `FileThumb`
 - Fixed the video extension pill's incorrect position with portrait video (now it will be correctly positioned regardless of the video's aspect ratio)
 - Internalized the `mediaAspectSquare` style introduced by `MediaView` (can be enabled with `cover=true`)
 - Changed `ObjectKind` to an enum for improved type checking.

### DIFF
--- a/interface/app/$libraryId/Explorer/File/Thumb.tsx
+++ b/interface/app/$libraryId/Explorer/File/Thumb.tsx
@@ -1,111 +1,146 @@
 import * as icons from '@sd/assets/icons';
 import clsx from 'clsx';
-import { CSSProperties, useEffect, useState } from 'react';
-import { ExplorerItem, useLibraryContext } from '@sd/client';
+import { useRef, useState } from 'react';
+import { ExplorerItem, isKeyOf, useLibraryContext } from '@sd/client';
 import { useExplorerStore } from '~/hooks/useExplorerStore';
 import { useIsDark, usePlatform } from '~/util/Platform';
 import { getExplorerItemData } from '../util';
 import classes from './Thumb.module.scss';
 
+export const getIcon = (
+	isDir: boolean,
+	isDark: boolean,
+	kind: string,
+	extension?: string | null
+) => {
+	if (isDir) return icons[isDark ? 'Folder' : 'Folder_Light'];
+
+	let document: Extract<keyof typeof icons, 'Document' | 'Document_Light'> = 'Document';
+	if (extension) extension = `${kind}_${extension.toLowerCase()}`;
+	if (!isDark) {
+		kind = kind + '_Light';
+		document = 'Document_Light';
+		if (extension) extension = extension + '_Light';
+	}
+
+	return icons[
+		extension && isKeyOf(icons, extension) ? extension : isKeyOf(icons, kind) ? kind : document
+	];
+};
+
 interface Props {
 	data: ExplorerItem;
-	size: number;
-	loadOriginal?: boolean;
+	size: null | number;
 	className?: string;
+	loadOriginal?: boolean;
 	forceShowExtension?: boolean;
-	extensionClassName?: string;
 }
 
 export default function Thumb(props: Props) {
-	const { cas_id, isDir, kind, hasThumbnail, extension } = getExplorerItemData(props.data);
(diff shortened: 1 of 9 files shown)


packages/client/src/utils/objectKind.ts
```diff
@@ -1,26 +1,28 @@
...
+	Widget,
+	Album,
+	Collection,
+	Font,
+	Mesh,
+	Code,
+	Database
+}
+
+export type ObjectKinds = keyof typeof ObjectKind;
```
Brendonovich, line 28: I think this would be better called `ObjectKindKey` - it shouldn't be plural since union types are multiple options for a single type, and `ObjectKind` is already taken

packages/client/src/utils/index.ts
```diff
@@ -19,3 +19,7 @@ export function arraysEqual<T>(a: T[], b: T[]) {
 
 	return a.every((n, i) => b[i] === n);
 }
+
+export function isKeyOf<T extends object>(obj: T, key: PropertyKey): key is keyof T {
+	return key in obj;
+}
```
Brendonovich, line 25: this seems redundant

interface/app/$libraryId/Explorer/QuickPreview.tsx
```diff
@@ -53,7 +53,7 @@ const pdfViewerEnabled = () => {
 
 function FilePreview({ explorerItem, kind, src, onError }: FilePreviewProps) {
 	const className = clsx('relative inset-y-2/4 max-h-full max-w-full translate-y-[-50%]');
-	const fileThumb = <FileThumb size={1} data={explorerItem} className={className} />;
+	const fileThumb = <FileThumb size={0} data={explorerItem} cover={true} className={className} />;
```
Brendonovich, line 56:
```suggestion
	const fileThumb = <FileThumb size={0} data={explorerItem} cover className={className} />;
```

Brendonovich (changes requested):

packages/client/src/utils/index.ts
HeavenVolkoff: I agree XD. I just couldn't get the TS typchecker to reduce a variable type to be the `keyof` an object after some `in` checks, so I came up with this, but if you think it is too much I can remove it and place some `as keyof typeof ...` in those places, no problem

jamiepine (dismissed):

interface/app/$libraryId/Explorer/File/Thumb.tsx
```diff
@@ -1,111 +1,184 @@
...
 	className?: string;
-	forceShowExtension?: boolean;
-	extensionClassName?: string;
+	loadOriginal?: boolean;
 }
 
-export default function Thumb(props: Props) {
-	const { cas_id, isDir, kind, hasThumbnail, extension } = getExplorerItemData(props.data);
-	const store = useExplorerStore();
+function Thumb({ size, cover, ...props }: ThumbProps) {
```
Brendonovich, line 44: Probably worth throwing `memo` on here just in case, we're gonna be rendering a lot of thumbs

Brendonovich (changes requested):

interface/app/$libraryId/Explorer/File/Thumb.tsx
HeavenVolkoff: You talking about exporting the Thumb component wrapped in memo? If so, I am already doing that at the end of the file.

Brendonovich: ah didn't see that, my bad

Brendonovich (approved):

PR #735, opened 2023-04-21
[ENG-506] Disable links

niikeec:
This PR disables "disabled links" as they currently aren't.

### DIFF
--- a/interface/app/$libraryId/Layout/Sidebar/Link.tsx
+++ b/interface/app/$libraryId/Layout/Sidebar/Link.tsx
@@ -26,7 +26,7 @@ export default (props: PropsWithChildren<NavLinkProps & { disabled?: boolean }>)
 	return (
 		<NavLink
 			{...props}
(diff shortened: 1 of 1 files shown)


interface/app/$libraryId/Layout/Sidebar/Link.tsx
```diff
@@ -26,6 +26,7 @@ export default (props: PropsWithChildren<NavLinkProps & { disabled?: boolean }>)
 	return (
 		<NavLink
 			{...props}
+			onClick={(e) => props.disabled && e.preventDefault()}
```
utkubakir, line 29: Wait, what if it's not disabled? Don't you need to do `props.onClick(e)`?

utkubakir (changes requested):

interface/app/$libraryId/Layout/Sidebar/Link.tsx
niikeec: You're right

PR #737, opened 2023-04-21
[ENG-508] Filter locations

niikeec:
This PR adds location filtering in the locations settings.

### DIFF
--- a/interface/app/$libraryId/settings/library/locations/index.tsx
+++ b/interface/app/$libraryId/settings/library/locations/index.tsx
@@ -1,3 +1,5 @@
+import { useMemo, useState } from 'react';
+import { useDebouncedCallback } from 'use-debounce';
 import { useLibraryQuery } from '@sd/client';
 import { SearchInput } from '@sd/ui';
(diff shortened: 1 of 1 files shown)


interface/app/$libraryId/settings/library/locations/index.tsx
```diff
@@ -7,20 +9,35 @@ import ListItem from './ListItem';
 export const Component = () => {
 	const locations = useLibraryQuery(['locations.list']);
 
+	const [search, setSearch] = useState('');
+	const updateSearch = useDebouncedCallback(setSearch, 750);
```
Brendonovich, line 13: `useDebounce` is probably more appropriate for something like this

niikeec: I'd say both work in this case as performance isn't a concern here, otherwise, I would lean towards the callback as useDebounce causes more rerenders.

utkubakir (approved):
I agree with @niikeec performance is meh here.

PR #740, opened 2023-04-21
[ENG-490] Add support for late key manager setup

brxken128:
This PR allows users to set up the key manager after onboarding, as shown in the screenshot.

It also removes key manager setup from onboarding, as it was decided that their should be a dedicated key manager configuration stage.

This removes the secure temp keystore also.

Mind the UI, this has been mentioned in Slack.
key manager setup screen

The UI has been fixed thanks to @HeavenVolkoff, it looks pretty awesome now!


Brendonovich (approved):

utkubakir:
I made your changes to mobile so we should be good on the ts side. Let's make clippy happy and merge this!

packages/ui/src/Popover.tsx
```diff
@@ -15,7 +15,10 @@ export const Popover = ({ trigger, children, disabled, className, ...props }: Pr
 		if (popOverRef.current && triggerRef.current) {
 			if (
 				!popOverRef.current.contains(event.target as Node) &&
-				!triggerRef.current.contains(event.target as Node)
+				!triggerRef.current.contains(event.target as Node) &&
+				// FIX-ME: WORKAROUND for Portal elements inside Popover
+				event.target !== document.body &&
+				event.target !== document.body.parentElement
```
HeavenVolkoff, line 21: This is not a good solution, but it should work for now. Based on my understanding, it shouldn't interfere with any other functionality as the popover itself is in a Portal, and we don't have anything on the body except for portalled components. Another possible solution would be to use a Context that wraps the Popover children and shares the popover itself to be used as the container for all Portals created inside it. However, this approach requires adjusting all components that use Portals to take this into account. Therefore, I didn't include it in this PR and will open a separate issue for it.

HeavenVolkoff:
UI now:

[Screencast from 2023-04-22 00-59-20.webm](https://user-images.githubusercontent.com/9082460/233761275-c8e649f5-e7a4-4fb6-af8f-23586a7814c1.webm)

Brendonovich (dismissed):

utkubakir (approved):

PR #750, opened 2023-04-22
[ENG-471, ENG-472] Add `dry_run` parameter to some API routes and related UI changes

HeavenVolkoff:
This PR adds a `dry_run` parameter to the following API routes:

 - `location.create`
 - `location.addLibrary`
 - `location.indexer_rules.create`
 
The `dry_run` parameter enables the frontend to validate data with backend validation without committing it to the DB. 

Additionally, this PR includes the following changes:

 - `addLocationDialog` and `indexerRuleEditor` have been modified to allow automatic data validation without user interaction.
 - `method` is now part of the `addLocationDialog` form schema, and a hidden input
 - `useCallbackToWatchForm` hook has been created to simplify watching form changes with an asynchronous function that also executes during component mount.
 - An `index.ts` has been added for hooks and components.
 - `invalidate_query` has been added to any create/edit API methods of the `location` route.

Fixes ENG-471, ENG-472.


core/src/location/mod.rs
```diff
@@ -143,30 +165,50 @@ impl LocationCreateArgs {
...
-		Ok(location)
+			info!(
+				"Added library (library_id = {}) to location: {location:?}",
+				library.id
+			);
+
+			Ok(Some(location))
+		} else {
+			Err(LocationError::DryRunError)
+		}
```
fogodev, line 211: This `LocationError::DryRunError` here never happens, as you checked at `if self.dry_run` and `create_location` function can only return `None` if `dry_run` is `true`. So you can remove this error, remove this first if and put an `Ok(None)` in the `else` branch of the `if let`

fogodev (changes requested):

core/src/api/locations.rs
```diff
@@ -279,7 +285,11 @@ fn mount_indexer_rule_routes() -> AlphaRouter<Ctx> {
 		.procedure("create", {
 			R.with2(library())
 				.mutation(|(_, library), args: IndexerRuleCreateArgs| async move {
-					args.create(&library).await.map_err(Into::into)
+					if args.create(&library).await?.is_some() {
+						invalidate_query!(library, "locations.indexer_rules.list");
+					}
+
+					Ok(())
```
fogodev, line 292: You changed the behavior of this route, now it doesn't return the newly created indexer rule anymore, the front-end isn't expecting it?

HeavenVolkoff: No, this change doesn't affect the front-end since it's not using the returned indexer rule. Additionally, the `locations.create` endpoint also doesn't return the created location, so I think this change brings the `locations.indexer_rule.create` endpoint in line with it.

fogodev (approved):

Brendonovich (approved):

PR #753, opened 2023-04-22
[ENG-524] Dismissible notices

jamiepine:
WIP: needs to actually be dismissable + not affect the virtualized list rendering (try scrolling on media view, items unload because of the offset created by the notice)

also needs to be on most screens

### DIFF
--- a/interface/app/$libraryId/Explorer/DismissibleNotice.tsx
+++ b/interface/app/$libraryId/Explorer/DismissibleNotice.tsx
@@ -1,6 +1,9 @@
 import { Collection, Image, Video } from '@sd/assets/icons';
 import clsx from 'clsx';
+import { ReactNode } from 'react';
 import DismissibleNotice from '~/components/DismissibleNotice';
+import { dismissibleNoticeStore } from '~/hooks/useDismissibleNoticeStore';
+import { ExplorerLayoutMode, useExplorerStore } from '~/hooks/useExplorerStore';
 
 const MediaViewIcon = () => (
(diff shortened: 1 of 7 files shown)


interface/app/$libraryId/Explorer/View.tsx
```diff
@@ -94,9 +95,24 @@ export default memo((props: Props) => {
...
+					<>
+						<DismissibleNotice notice="listView" />
+						<ListView />
+					</>
+				)}
+				{layoutMode === 'media' && (
+					<>
+						<DismissibleNotice notice="mediaView" />
+						<MediaView />
+					</>
```
utkubakir, line 114: `DismissibleNotice` can take `layoutMode` to clean up some code.

interface/app/$libraryId/Explorer/DismissibleNotice.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+export default (props: Props) => {
+	const notice = notices[props.notice];
+
+	return (
+		<DismissibleNotice
+			title={
+				<>
+					<span className="font-normal">Meet</span> {notice.title}
+				</>
```
utkubakir, line 50: Why is this wrapped with `<>`?

utkubakir (changes requested):

interface/app/$libraryId/Explorer/View.tsx
niikeec: Pushed a change

utkubakir (approved):

PR #754, opened 2023-04-24
Give file paths pub_id and use it primarily

Brendonovich:
Makes the indexer compatible with sync, yay

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6411,7 +6411,7 @@ dependencies = [
 [[package]]
 name = "rspc"
 version = "0.1.4"
-source = "git+https://github.com/oscartbeaumont/rspc?rev=745f0b210f6d268a4cef5d43b595e88d99ac269f#745f0b210f6d268a4cef5d43b595e88d99ac269f"
+source = "git+https://github.com/oscartbeaumont/rspc?rev=3e2608e9658b54240a3ba076fb0f7c4307e094fc#3e2608e9658b54240a3ba076fb0f7c4307e094fc"
 dependencies = [
  "futures",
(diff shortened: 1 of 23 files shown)


core/src/object/file_identifier/shallow_file_identifier_job.rs
```diff
@@ -136,18 +135,17 @@ impl StatefulJob for ShallowFileIdentifierJob {
...
+			// .order_by(file_path::id::order(Direction::Asc))
 			.select(file_path::select!({ id }))
 			.exec()
 			.await?
-			.map(|d| d.id)
 			.unwrap(); // SAFETY: We already validated before that there are orphans `file_path`s
 
 		// SAFETY: We just initialized `state.data` above
-		state.data.as_mut().unwrap().cursor.file_path_id = first_path_id;
+		state.data.as_mut().unwrap().cursor = first_path.id;
```
fogodev, line 148: It's guaranteed that `file_path`s will come ordered by id?

Brendonovich: Kind of? Instead of `id` being generated by the whole atomic system that used to exist before, it'll just be assigned whenever the file paths are inserted. They'll still be somewhat ordered by id but maybe not in exactly the same way.

fogodev: Yeah, I got this doubt because the following logic got identifier with the cursor stuff assumes that we got the very first `file_path.id` that passes our filtering. If this applies, then it's alright to me.

fogodev (approved):

PR #756, opened 2023-04-24
rspc 0.1.3 prep

oscartbeaumont:
Changes:
 - Change `rspc::Type` to `specta::Type` ready for the upstream change over
 - Upgrade rspc
 - Upgrade Specta
 - Upgrade PCR
 - Library subscriptions that don't suck finally!!!!!
 - Also removed all custom `impl Type`'s which is gonna be nice for easy Specta upgrades even over breaking changes.
 - The library query system's ease of maintenance has regressed in this PR but it's gonna be redone soon anyway.

### DIFF
--- a/core/src/api/sync.rs
+++ b/core/src/api/sync.rs
@@ -5,12 +5,9 @@ use super::{utils::LibraryRequest, RouterBuilder};
 pub fn mount() -> RouterBuilder {
 	RouterBuilder::new()
 		.library_subscription("newMessage", |t| {
-			t(|ctx, _: (), library| {
+			t(|_, _: (), library| {
 				async_stream::stream! {
-					let Some(lib) = ctx.library_manager.get_ctx(library.id).await else {
-						return
-					};
-					let mut rx = lib.sync.tx.subscribe();
+					let mut rx = library.sync.tx.subscribe();
 					while let Ok(msg) = rx.recv().await {
 						let op = match msg {
 							SyncMessage::Ingested(op) => op,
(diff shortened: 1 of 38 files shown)


oscartbeaumont:
This PR also fixes this which is an edge case I didn't consider in the error logging. Everything is still working but these logs are stupid annoying.
image

core/src/api/sync.rs
```diff
@@ -5,9 +5,9 @@ use super::{utils::LibraryRequest, RouterBuilder};
 pub fn mount() -> RouterBuilder {
 	RouterBuilder::new()
 		.library_subscription("newMessage", |t| {
-			t(|ctx, _: (), library_id| {
+			t(|ctx, _: (), library| {
 				async_stream::stream! {
-					let Some(lib) = ctx.library_manager.get_ctx(library_id).await else {
+					let Some(lib) = ctx.library_manager.get_ctx(library.id).await else {
```
Brendonovich, line 10: Isn't this unnecessary since we already have `library`?

crates/crypto/Cargo.toml
```diff
@@ -8,7 +8,8 @@ edition = "2021"
 rust-version = "1.67.0"
 
 [features]
-rspc = ["dep:rspc"]
+rspc = ["dep:rspc", "dep:specta"]
```
Brendonovich, line 11: i think only the `specta` feature is needed now

oscartbeaumont: Nope. rspc is for `rspc::Error`. They can probs be one feature but I am keeping the previous semantics.

Brendonovich: Ah ok cool

Brendonovich (approved):

PR #757, opened 2023-04-24
[ENG-499] Search improvements

niikeec:
This PR adds explorer top bar options to search, a pre-search & no results state, and dir navigation from search.

### DIFF
--- a/interface/hooks/useTopBarOptions.tsx
+++ b/interface/hooks/useTopBarOptions.tsx
@@ -11,7 +11,7 @@ import OptionsPanel from '~/app/$libraryId/Explorer/OptionsPanel';
 import { TOP_BAR_ICON_STYLE, ToolOption } from '~/app/$libraryId/TopBar';
 import { getExplorerStore, useExplorerStore } from './useExplorerStore';
 
-export const useToolOptions = () => {
+export const useTopBarOptions = () => {
 	const explorerStore = useExplorerStore();
 
 	const explorerViewOptions: ToolOption[] = [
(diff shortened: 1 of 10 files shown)


interface/hooks/useTopBarOptions.tsx
```diff
@@ -0,0 +1,72 @@
...
+	Rows,
+	SidebarSimple,
+	SlidersHorizontal,
+	SquaresFour
+} from 'phosphor-react';
+import OptionsPanel from '~/app/$libraryId/Explorer/OptionsPanel';
+import { TOP_BAR_ICON_STYLE, ToolOption } from '~/app/$libraryId/TopBar';
+import { getExplorerStore, useExplorerStore } from './useExplorerStore';
+
+export const useTopBarOptions = () => {
```
utkubakir, line 14: Why do we have this hook? I thought @ameer2468 worked on making it definable for every screen independently so we can have different options.

niikeec: Search and Explorer use these 2 top bar options. Why define on 2 different ends.

ameer2468: This hook was deleted, we don't need a hook anymore.

utkubakir: He does that @ ameer2468, It is just something he created to save duplicate code which is fair, It might be helpful to add a comment to the hook so we don't get confused.

utkubakir (approved):

utkubakir (approved):

PR #760, opened 2023-04-26
Update deps

oscartbeaumont:
Mobile and Crypto have been left out of these updates.


crates/crypto/Cargo.toml
```diff
@@ -57,11 +57,11 @@ hex = "0.4.3"
 
 # linux OS keyring - using `v2.0.2` as newer versions broke a lot due to async/lifetimes
 [target.'cfg(target_os = "linux")'.dependencies]
-secret-service = { version = "2.0.2", optional = true }
+secret-service = { version = "3.0.1", optional = true }
```
brxken128, line 60: can you undo this entire file please? this change alone breaks everything keyring-related as it forces you to use an async runtime, which in turn means the keyring trait needs to be unnecessarily async.

all of the crypto deps will be updated in my new PR - if any of the stuff in here _needs_ changing for the app to build with thisnPR, then go for it

brxken128 (changes requested):

crates/crypto/Cargo.toml
oscartbeaumont: Done. I hadn't finished testing yet but thanks for letting me know to save me to time of working that out.

oscartbeaumont:
It hurts me that this PR reformats a bunch of Tailwind classes in other files, which is why I closed it.

@brxken128  made a good point that this PR decreases the number of known vulnerabilities in Spacedrive (as reported by Cargo Audit) from 18 to 3 so it's probs worth doing.

PR #761, opened 2023-04-26
Spacedrop

oscartbeaumont:


Brendonovich (changes requested):

core/src/p2p/p2p_manager.rs
```diff
@@ -30,23 +40,25 @@ pub enum P2PEvent {
...
+		peer_id: PeerId,
+		name: String,
 	},
+	// TODO: Expire peer + connection/disconnect
 }
 
 pub struct P2PManager {
 	pub events: broadcast::Sender<P2PEvent>,
 	pub manager: Arc<Manager<PeerMetadata>>,
+	spacedrop_pairing_reqs: Arc<Mutex<HashMap<Uuid, oneshot::Sender<Option<String>>>>>,
```
fogodev, line 54: Makes sense to use a `DashMap` here?

brxken128: I'm inclined to agree also, dashmap would fit perfectly

oscartbeaumont: I’m inclined to not agree because we don’t want eventual consistency we want strong consistency. The map also has very little contention. I could use the ‘evmap’ crate given I think it has the ability to synchronise readers but with low contention I don’t think it’s going to be of any performance gain.

I have and generally will avoid ‘DashMap’ for most P2P stuff because I have experienced problems with its eventually consistent nature before in this application of it. I suspect this is more of a dev only thing but still. With two nodes running on the same computer they can connect on the scale of milliseconds instead of tens or hundreds of milliseconds like it would between physical machines.

fogodev: If it has little contention, I don't oppose to`Mutex<HashMap<K, V>>` then, but `evmap` actually has eventual consistency. I haven't checked `DashMap` source code, but the author says that it uses shards and each shard is behind a `RwLock` internally, but I can't guarantee anything about its consistency model anyway.

oscartbeaumont: `evmap` and `DashMap` are a bit different in implementation by my understanding which is why I bring it up. `evmap` has a `.refresh` call which swaps the read/write pointers allowing you to control a switch from eventual consistency to strong consistency by my understanding.

fogodev: Oh now I got your point then

fogodev (approved):
I wish it had less `unwraps` but I understand that it's an initial implementation of the feature that will still be refined, so LGTM for now

Brendonovich (approved):

PR #762, opened 2023-04-26
File Opening

Brendonovich:


apps/desktop/src-tauri/src/main.rs
```diff
@@ -17,18 +17,46 @@ mod app_linux;
...
+	if let Some(library) = node.library_manager.get_library(library).await {
+		let path = library.get_file_path(id).await.unwrap().unwrap();
+
+		dbg!(&path);
+
+		opener::open(path).ok();
+	}
+
+	Ok(())
+}
```
fogodev, line 43: Can this have proper error handling, please? lol to avoid technical debt. Also, was this `dbg!` forgotten here?

fogodev: Maybe we should introduce `anyhow` or `eyre` in the desktop app

core/src/library/library.rs
```diff
@@ -97,4 +98,30 @@ impl Library {
...
+			}))
+			.exec()
+			.await?
+			.map(|record| {
+				// I hate this but everything starts with `/` so we can't PathBuf::join
+				PathBuf::from(format!(
+					"{}{}",
+					record.location.path, record.materialized_path
+				))
+			}))
```
fogodev, line 125: You can join with the `MaterializedPath` it will handle any necessary logic regarding materialized paths.

```rust
Path::new(&record.location.path).join(&MaterializedPath::from((record.location_id, &record.materialized_path));
```

fogodev (changes requested):

fogodev (approved):

PR #769, opened 2023-04-27
[ENG-533] Update navigation

niikeec:


interface/app/$libraryId/Layout/Sidebar/Link.tsx
```diff
@@ -20,20 +20,25 @@ const styles = cva(
...
+	...props
+}: PropsWithChildren<NavLinkProps & { disabled?: boolean }>) => {
 	const os = useOperatingSystem();
+	const location = useLocation();
 
 	return (
 		<NavLink
-			{...props}
 			onClick={(e) => (props.disabled ? e.preventDefault() : props.onClick?.(e))}
+			replace={!!location.pathname.match(/^\/[\w\d-]+\/settings\//)}
```
Brendonovich, line 33: Leveraging `useMatch` here would probably be nicer

Brendonovich (dismissed):
This whole system seems rather complex when we already have a router - could a combination of `replace: true` and checking `history` do the job?

interface/app/$libraryId/TopBar/SearchBar.tsx
```diff
@@ -39,7 +39,7 @@ export default () => {
 	}, 300);
 
 	useEffect(() => {
-		updateParams(value);
+		if (searchPath.pathname === location.pathname) updateParams(value);
 	}, [value]);
```
utkubakir, line 43: I think you need to add `searchPath` & `location`

interface/app/onboarding/Layout.tsx
```diff
@@ -19,20 +25,26 @@ export const Component = () => {
 	const debugState = useDebugState();
 	const navigate = useNavigate();
 
+	const libraries = useCachedLibraries();
+	const library =
```
utkubakir, line 29: Don't we have a way to get the currentLibrary? defaulting the first one seems weird

niikeec: Only defaults to first if it can't find the current one.

utkubakir (approved):

PR #771, opened 2023-04-28
[ENG-551] - Jobs grouping and UI improvements

ameer2468:
**This is an upgrade to the Job manager UI, jobs are now being grouped.** 

- Jobs with assosicated parents are grouped.
- Individual running jobs and individual jobs are displayed alone.
- If any grouped jobs have running jobs, and their children are removed, the running job is displayed alone.

**Videos**:

https://user-images.githubusercontent.com/33054370/235203181-f63a7651-dbb6-4524-a334-2e31462f1aef.mov

https://user-images.githubusercontent.com/33054370/235203560-f980241c-4843-47f2-9899-3d2b5729c8f6.mov

### DIFF
--- a/interface/app/$libraryId/Layout/Sidebar/GroupedJob.tsx
+++ b/interface/app/$libraryId/Layout/Sidebar/GroupedJob.tsx
@@ -1,7 +1,7 @@
 import clsx from 'clsx';
 import dayjs from 'dayjs';
 import { Folder, X } from 'phosphor-react';
-import { MutableRefObject, memo, useEffect, useRef, useState } from 'react';
+import { MutableRefObject, memo, useEffect, useState } from 'react';
 import { JobReport } from '@sd/client';
 import { Button, ProgressBar, Tooltip } from '@sd/ui';
 import { IGroupedJobs } from './GroupedJobs';


interface/app/$libraryId/Layout/Sidebar/GroupedJob.tsx
```diff
@@ -0,0 +1,191 @@
...
+							onClick={() => clearAJob?.(data.id as string)}
+							size="icon"
+						>
+							<Tooltip label="Remove">
+								<X className="h-4 w-4 cursor-pointer" />
+							</Tooltip>
+						</Button>
+					)}
+					<div
+						onClick={() => setToggleJobs(!toggleJobs)}
```
utkubakir, line 50: `setToggleJobs((v) = > !v))`

interface/app/$libraryId/Layout/Sidebar/Job.tsx
```diff
@@ -23,52 +23,66 @@ import { Button, ProgressBar, Tooltip } from '@sd/ui';
...
+		icon: Copy,
+		filesDiscovered: ''
 	},
 	file_cutter: {
 		name: `Moved ${numberWithCommas(job.task_count)} ${filesTextCondition(job)}`,
-		icon: Scissors
+		icon: Scissors,
+		filesDiscovered: ''
 	}
 });
```
utkubakir, line 87: Why do we have `filesDiscovered` here? Maybe change it to data or some other name because if we follow this syntax, it will get ugly pretty quickly.

interface/app/$libraryId/Layout/Sidebar/GroupedJob.tsx
```diff
@@ -0,0 +1,191 @@
...
+								/>
+							))}
+							{data.childJobs.map((job) => (
+								<Job
+									isGroup={true}
+									className={clsx(
+										`border-none pl-10`,
+										toggleJobs && 'darker-app-bg'
+									)}
+									clearAJob={(arg) => clearJobHandler(arg)}
```
utkubakir, line 114: `clearAJob={clearJobHandler}` also renaming `clearAJob` to `clearJob` might be better here.

interface/app/$libraryId/Layout/Sidebar/GroupedJob.tsx
```diff
@@ -0,0 +1,191 @@
...
+					</div>
+					{toggleJobs && (
+						<>
+							{data.runningJobs.map((job) => (
+								<Job
+									className={clsx(
+										`border-none pl-10`,
+										toggleJobs && 'darker-app-bg'
+									)}
+									isGroup={true}
```
utkubakir, line 102: replace i`sGroup={true}` with `isGroup`

interface/app/$libraryId/Layout/Sidebar/GroupedJob.tsx
```diff
@@ -0,0 +1,191 @@
...
+											)} */}
+										</div>
+									</div>
+									<div className="grow" />
+								</div>
+								{!allJobsCompleted && !toggleJobs && (
+									<div className="mt-[6px] w-full">
+										<ProgressBar
+											value={getTotalTasks(data.childJobs).completed}
+											total={getTotalTasks(data.childJobs).total}
```
utkubakir, line 87: Maybe make this `getTotalTasks(data.childJobs)` into a constant, you seem to use it a lot and loops are bad 🤔

interface/app/$libraryId/Layout/Sidebar/GroupedJob.tsx
```diff
@@ -0,0 +1,191 @@
...
+import { Folder, X } from 'phosphor-react';
+import { MutableRefObject, memo, useEffect, useRef, useState } from 'react';
+import { JobReport } from '@sd/client';
+import { Button, ProgressBar, Tooltip } from '@sd/ui';
+import { IGroupedJobs } from './GroupedJobs';
+import Job from './Job';
+
+interface GroupJobProps {
+	data: IGroupedJobs;
+	clearAJob?: (arg: string) => void;
```
utkubakir, line 12: How can this be undefined?

ameer2468: > How and why can this be undefined?

It's not because of it being undefined, the intention is to make the prop optional.

When we map the running jobs, we don't need the clearAJob handler function. Also wouldn't make sense to make a new component just to skip the idea of not making the prop optional.

interface/app/$libraryId/Layout/Sidebar/GroupedJob.tsx
ameer2468: > replace `isGroup={true}` with `isGroup`

This wouldn't be a good idea from a readability perspective, the component has multiple props, therefore it would be better to keep it like this for consistency.

If it was just one prop, id put isGroup alone.

interface/app/$libraryId/Layout/Sidebar/Job.tsx
ameer2468: > Why do we have `filesDiscovered` here? Maybe change it to data or some other name because if we follow this syntax, it will get ugly pretty quickly.

Renaming it to Data implies that there are multiple variables or data, it doesn't show it's a single value.

Meanwhile, filesDiscovered is clearer naming and I would know there is only one value here.

interface/app/$libraryId/Layout/Sidebar/GroupedJob.tsx
ameer2468: Will apply this to guarantee state updates, although this doesn't really make a difference in this context.

interface/app/$libraryId/Layout/Sidebar/GroupedJob.tsx
ameer2468: You're right, ill do this when I wake up tomorrow.

interface/app/$libraryId/Layout/Sidebar/GroupedJob.tsx
ameer2468: This is correct, no need to write an arrow function, can just put it in there.

interface/app/$libraryId/Layout/Sidebar/GroupedJob.tsx
ameer2468: Un-intended, this is a mistake actually. Shouldn't be an optional value nor it can be undefined.

interface/app/$libraryId/Layout/Sidebar/GroupedJob.tsx
Brendonovich: I'm with Utku on this one, `isGroup` looks nicer

interface/app/$libraryId/Layout/Sidebar/GroupedJob.tsx
utkubakir: it's also cleaner imo

interface/app/$libraryId/Layout/Sidebar/GroupedJob.tsx
```diff
@@ -0,0 +1,192 @@
...
+	};
+
+	return (
+		<>
+			{data.childJobs.length === 0 ? null : (
+				<ul className={clsx(`relative overflow-hidden`, toggleJobs && 'groupjobul')}>
+					{allJobsCompleted && !checkForJobsRunning && (
+						<Button
+							className="absolute right-[10px] top-[30px] cursor-pointer"
+							onClick={() => clearJob?.(data.id as string)}
```
utkubakir, line 42: `?.` can be removed

interface/app/$libraryId/Layout/Sidebar/Job.tsx
```diff
@@ -81,30 +95,41 @@ const StatusColors: Record<JobReport['status'], string> = {
 	Paused: 'text-gray-500'
 };
 
-function Job({ job, clearAJob }: { job: JobReport; clearAJob?: (arg: string) => void }) {
-	const niceData = getNiceData(job)[job.name] || {
+interface JobProps {
+	job: JobReport;
+	clearAJob?: (arg: string) => void;
```
utkubakir, line 100: How about this? why is this undefined and can also be renamed to "clearJob"

interface/app/style.scss
```diff
@@ -213,3 +217,58 @@ body {
...
+		position: absolute;
+		z-index: 10;
+		height: 100%;
+		top: 0;
+		width: 10px;
+		border-left: 1px solid rgba(50, 51, 67, 1);
+		content: '';
+		left: 25px;
+	}
+}
```
utkubakir, line 274: Can we do without these? If not, let's move it to where are they used as they are too specific

interface/app/style.scss
```diff
@@ -19,6 +19,10 @@ body {
 	}
 }
 
+.darker-app-bg {
+	background: #20202b;
+}
+
```
utkubakir, line 25: Instead of hardcoding, adding it to `colors` and the tailwind config is better (for theming)

interface/app/$libraryId/Layout/Sidebar/JobManager.tsx
```diff
@@ -21,7 +28,7 @@ export function JobsManager() {
 			queryClient.invalidateQueries(['jobs.getHistory']);
 		}
 	});
-	const { mutate: clearAJob } = useLibraryMutation(['jobs.clear'], {
+	const clearAJob = useLibraryMutation(['jobs.clear'], {
```
utkubakir, line 31: clearJob

utkubakir (commented):
The code looks good imo, it's just nitpicking the code style atm. Great job man!

interface/app/$libraryId/Layout/Sidebar/GroupedJob.tsx
```diff
@@ -0,0 +1,192 @@
...
+function GroupedJob({ data, clearJob }: GroupJobProps) {
+	const [toggleJobs, setToggleJobs] = useState<MutableRefObject<boolean> | boolean>(false);
+	const toggleRef = useRef(toggleJobs);
+	const checkForJobsRunning = data.childJobs?.some((job) => job.status === 'Running');
+	const allJobsCompleted = data.childJobs?.every((job) => job.status === 'Completed');
+	const getTasks = getTotalTasks(data.childJobs);
+
+	useEffect(() => {
+		setToggleJobs(toggleRef.current); //this is to keep the toggled group open on re-renders
+	}, []);
```
Brendonovich, line 24: Why is there both `useState` and `useRef`? `toggleRef` is never updated and will always be `false`, and the `useEffect` doesn't do anything since `toggleJobs` will always be initialised as `false`

interface/app/style.scss
```diff
@@ -213,3 +217,58 @@ body {
 	width: 8px !important;
 	border-radius: 3px !important;
 }
+
+.removelistdot {
+	list-style: none;
+}
+
+.groupjobul {
```
Brendonovich, line 225: idk if it's possible but it'd be great if this was implemented in tailwind with our theme's colours and `rem` sizing rather than `px`

Brendonovich (changes requested):

interface/app/$libraryId/Layout/Sidebar/Job.tsx
ameer2468: This is needed, when we map the running jobs we don't want to pass a clearJob function.

interface/app/$libraryId/Layout/Sidebar/GroupedJob.tsx
ameer2468: I forgot to remove this, in a previous commit this was needed.

utkubakir (approved):

utkubakir (approved):

Brendonovich (approved):

PR #775, opened 2023-05-01
Upgrade rspc TS

oscartbeaumont:
Will allow the development and usage of:
 - `useInfiteQuery` for explorer
 - Typesafe errors


oscartbeaumont:
Blocked on Metro not supporting "package.json" "exports" so it can't understand `@rspc/client/v2` is valid.

utkubakir (approved):
Let's wait for Mobile CI, other than that LGTM

packages/client/src/rspc.tsx
```diff
@@ -0,0 +1,118 @@
...
+export const useLibrarySubscription = libraryHooks.useSubscription;
+
+export function useInvalidateQuery() {
+	const context = nonLibraryHooks.useContext();
+	useBridgeSubscription(['invalidation.listen'], {
+		onData: (ops) => {
+			for (const op of ops) {
+				const key = [op.key];
+				if (op.arg !== null) {
+					key.concat(op.arg);
```
HeavenVolkoff, line 101: I'm not entirely clear on the logic behind the invalidation system, but I believe this line isn't behaving as expected. This is because `concat` doesn't mutate either of the existing arrays[¹], but instead returns a new merged array that includes the original source and its arguments. This means that the `key` array won't ever contain the value of `op.arg`. If the intention was to include `op.arg` in `key`, a working alternative would be to use `key.push(...op.arg)`  or `key = key.concat(op.arg)`.

[¹]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/concat

PR #779, opened 2023-05-03
[ENG-581, ENG-597, ENG-600] Overview categories

jamiepine:
Screenshot 2023-05-17 at 4 09 39 AM

### Introducing Overview Categories

Categories finally give the Overview screen some functionality, a place to quickly discover the contents of your library.

Mostly based on ObjectKind, with additional support for recents, favorites, downloads and more.

- Added "Double click action" setting to explorer view options [ENG-597]. 
- Added new icons: Favorite, Game, Book and Trash.
- Added persisted store `useExplorerConfigStore` for saving double click preference.
- Added support for Books to the `file_ext` crate and the `Book` ObjectKind.
- Added empty state to the Inspector [ENG-600]
- Accidentally committed my stashed changed to the key manager UI (don't let it stop you reviewing, will fix in another PR)
- Added current node to the sidebar, greyed out for now.
- Refactored `overview.tsx` into a directory with smaller components.

**Bugs and issues:** 
- Due to how the category selection is rendered as a child of the explorer, a re-render flicker sometimes occurs when changing categories (doesn't happen in prod?)
- For some reason the first time a category is loaded, only some of the results render. Clicking away and then back again fixes it.
- Sometimes the top bar options on the Overview disappear until a re-render.
- Categories are scrollable horizontally, but there is no indication of this, a subtle fade and a circular right caret button might be necessary.
- List view headings are obstructed by the category bar


interface/hooks/useExplorerConfigStore.ts
```diff
@@ -0,0 +1,14 @@
...
+	openOnDoubleClick: true
+});
+
+export function useExplorerConfigStore() {
+	return useSnapshot(explorerConfigStore);
+}
+
+export function getExplorerConfigStore() {
+	return explorerConfigStore;
+}
```
utkubakir, line 14: Do we need another store? I feel like this can be in the `explorerStore`

interface/app/$libraryId/settings/library/locations/$id.tsx
```diff
@@ -22,6 +22,7 @@ const schema = z.object({
 	path: z.string(),
 	hidden: z.boolean(),
 	indexerRulesIds: z.array(z.number()),
+	location_type: z.string(),
```
utkubakir, line 25: `locationType` to follow the name syntax

utkubakir (changes requested):
Gotta make the CI happy, other than that TS part looks great!

interface/hooks/useExplorerConfigStore.ts
jamiepine: I don't think you can persist half the store with Valtio? Cause not everything in explorerStore should be persisted

interface/app/$libraryId/settings/library/locations/$id.tsx
jamiepine: oops

interface/hooks/useExplorerConfigStore.ts
utkubakir: oh right, that's true. We gotta move stuff here to persist them then (like `layoutMode`, `gridItemSize`)

fogodev:
The core part LGTM, feel free to merge when front-end is ready

interface/hooks/useExplorerConfigStore.ts
jamiepine: Lets tackle that later, I want to explore how we can maybe persist these differently per location / folder + override them depending on context. Will explain later.

utkubakir (approved):

PR #781, opened 2023-05-04
[ENG-536] Improve indexer for huge locations

fogodev:
Introduced a bunch of changes regarding the indexing pipeline on core:

First, I changed the `file_path` table, totally removing the `parent_id` column. This column was to represent a tree in SQL using the adjacency list model, as we also have the `materialized_path` column that is used to represent a tree in SQL using the path enumeration model. I removed the `parent_id` as adjacency list is a weaker abstraction. Following this removal, I changed how we used materialized paths entirely, now it just stores the path to the parent directory, with a trailing `/`. This way we don't store twice the name and extension, as before they appeared on the `materialized_path`, `name` and `extension` columns respectively. With these changes, we optimized a little of how much data we store on the database.

With the `parent_id` removal, now the walker and the indexer jobs could be optimized. First I separated the walking part from the `init` method and introduced 2 kinds of step for the full indexer job, the first kind is just to save found files on db, like before. The second kind is to keep walking on a new directory that wasn't walked before. This way we eliminated the problem that for really huge locations we only save progress after the `init` walked through every possible directory, which could take really long. For now, I introduced a limit of 50k found files to stop the walker and store the “to be walked” directories on new steps, and this limit is easily adjustable as a parameter.

Aside from that. I also introduce a new kind of error called `FileIOError` that should be used for error handling instead of `#[from] io::Error` which never tells us which file had an error. It's a bit verbose having to use a bunch of `.map_err(|e| FileIOError::from((path, e)))` but such is life... I tried to have a better error handling on a bunch of places already, but I don't think that I covered all bases.

These changes basically broke half of the core, so I had to fix everything, and now this PR is enormous LMAO, sorry about that.


apps/mobile/src/components/explorer/Explorer.tsx
```diff
@@ -34,7 +34,7 @@ const Explorer = ({ data }: ExplorerProps) => {
 		if (isPath(data) && data.item.is_dir) {
 			navigation.push('Location', {
 				id: data.item.location_id,
-				path: data.item.materialized_path
+				path: `${data.item.materialized_path}${data.item.name}/`
```
Brendonovich, line 37: I wonder if we should treat paths as arrays of segments rather than as actual strings. Joining them on the frontend like this just feels really wrong.

Brendonovich (approved):
Seems to work great, would like to get a review from @oscartbeaumont before merging

apps/mobile/src/components/explorer/Explorer.tsx
fogodev: Yeah, I agree on this, I didn't like this either. Would passing the `materialized_path` and the `name` as separated parameters in the front and joining them in the back using `IsolatedFilePathData` facilities a good tradeoff?

Brendonovich: Something like that sounds fair

fogodev: Actually, this one is being used on `location.getExplorerData` route, which is being phased out at #812. So we have to discuss how these 2 PRs will be handled.

Brendonovich: I'm gonna merge #812 and then get this pr fixed up and merged

Brendonovich (approved):

PR #782, opened 2023-05-04
Hide Notice and some tweaks

utkubakir:
- Removed prettier/recommended from eslint
- Optimized imports
- Added auto-generated file to prettier ignore list


interface/app/$libraryId/Explorer/DismissibleNotice.tsx
```diff
@@ -52,9 +53,15 @@ const notices = {
...
 export default () => {
 	const { layoutMode } = useExplorerStore();
+	const location = useLocation();
+
 	const notice = notices[layoutMode];
 
 	if (!notice) return null;
+
+	// Hide notice on overview page
+	if (location.pathname.endsWith('/overview')) return null;
```
Brendonovich, line 63: I feel like conditionally rendering in the parent based on the result of a `useMatch` would be a bit nicer than `DismissableNotice` being aware of its route

Brendonovich (changes requested):

PR #783, opened 2023-05-04
Job Manager improvement & refactor

ameer2468:
These are improvements to the previous Job manager work that I did.

- Jamie reported that he couldn't see the running job of Generating thumbnails, this fixes it.
- Improved text conditional wording, if there's multiple task(s) or file(s) + a string should be passed to add an S to the end.
- If a job is queued, the 'X' button does not render + the text 'Queued' shows if a job is queued.


interface/app/$libraryId/Layout/Sidebar/Job.tsx
```diff
@@ -199,8 +219,9 @@ function JobTimeText({ job }: { job: JobReport }) {
 	}
 }
 
-function filesTextCondition(job: JobReport) {
-	return job?.task_count > 1 || job?.task_count === 0 ? 'files' : 'file';
+function JobCountTextCondition(job: JobReport, word: string) {
+	const addStoEnd = job?.task_count > 1 || job?.task_count === 0 ? `${word}s` : `${word}`;
+	return addStoEnd;
```
utkubakir, line 224: This looks weird, if the job can be undefined, how can you compare it with a number? without doing like this `(job?.task_count || 0)`. I guess if ts is okay with this, I am too lol

interface/app/$libraryId/Layout/Sidebar/Job.tsx
```diff
@@ -214,15 +235,25 @@ export function AllRunningJobsWithoutChildren({
...
-	const mapJobsForIds = jobs?.map((job) => job.id);
-	const checkIfJobHasChildren = filterRunning?.filter((job) => !mapJobsForIds?.includes(job.id));
+	const runningJobsNoChildren = () => {
+		const singleRunningJobs = [];
+		for (const job of jobs) {
+			for (const runningJob of runningJobs) {
+				if (
+					job.parent_id !== runningJob.id &&
+					job.id !== runningJob.id &&
+					job.id !== job.id
```
utkubakir, line 245: `job.id !== job.id` seems unnecessary

interface/app/$libraryId/Layout/Sidebar/Job.tsx
ameer2468: The ? isn't needed actually, not sure why it was there in the first place O.o

interface/app/$libraryId/Layout/Sidebar/Job.tsx
ameer2468: Without `job.id !== job.id` here is the result of duplication

Screenshot 2023-05-04 at 11 51 35 PM

jamiepine (approved):

PR #785, opened 2023-05-04
[ENG-575] Infinite scroll

niikeec:


core/src/api/locations.rs
```diff
@@ -152,7 +155,11 @@ pub(crate) fn mount() -> AlphaRouter<Ctx> {
 							.exec()
 							.await?;
 
-						let cursor = results.pop().map(|r| r.pub_id);
+						let cursor = if results.len() as i32 > limit {
+							results.pop().map(|r| r.pub_id)
+						} else {
+							None
+						};
```
utkubakir, line 162: Are we losing any items to this? We are taking limit + 1 and seem to be popping that so I think it would be wise to double-check this.

utkubakir (changes requested):

utkubakir (approved):

utkubakir (approved):

PR #786, opened 2023-05-04
source `has_thumbnail` correctly for the recent file query

brxken128:
For whatever reason, the `has_thumbnail` column in the `Object` table is never actually used, and as such it is not a good metric for checking whether or not a thumbnail actually exists. This PR changes that, and instead uses the `library.has_thumnail()` function.

I have not tested whether or not this fixes thumbnails in the recent files view.

I think a good solution would be to allow the thumbnailer and indexers to check whether or not each object has a thumbail, and update the database accordingly. This means that even if the user deletes a specific thumbnail, the database can stay up to date. It would also mean that we don't need to rely on a separate function just for checking if a thumbnail exists or not, we can just source it from the database.

### DIFF
--- a/core/prisma/migrations/old20230504181124_remove_has_thumbnail/migration.sql
+++ b/core/prisma/migrations/old20230504181124_remove_has_thumbnail/migration.sql
@@ -1,0 +1,30 @@
+/*
+  Warnings:
+
+  - You are about to drop the column `has_thumbnail` on the `object` table. All the data in the column will be lost.
+
+*/
+-- RedefineTables
+PRAGMA foreign_keys=OFF;
+CREATE TABLE "new_object" (
+    "id" INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT,
+    "pub_id" BLOB NOT NULL,
+    "kind" INTEGER NOT NULL DEFAULT 0,
+    "key_id" INTEGER,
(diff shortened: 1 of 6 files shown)


jamiepine:
Ah, I see where the confusion came from, but they are supposed to be different, and we will be changing the names. When we query, we check the filesystem for the thumbnail, because we need to know for _sure_ the system has access to that image, otherwise we render a question mark (default browser behaviour), but the database column (which should be renamed to has_generated_thumbnail (...and moved to the Object model)) tells us if one has been generated at all so we can fetch it from that node if it doesn't exist on the current one, this will be implemented later. 

With that said, I can't merge this, sorry Jake!

brxken128:
> Ah, I see where the confusion came from, but they are supposed to be different, and we will be changing the names. When we query, we check the filesystem for the thumbnail, because we need to know for _sure_ the system has access to that image, otherwise we render a question mark (default browser behaviour), but the database column (which should be renamed to has_generated_thumbnail (...and moved to the Object model)) tells us if one has been generated at all so we can fetch it from that node if it doesn't exist on the current one, this will be implemented later.
> 
> With that said, I can't merge this, sorry Jake!

This also fixed thumbnails being displayed in the recent files view, as it sources the `has_thumbnail` value for the `ExplorerItem` in exactly the same way as the explorer query.

PR #789, opened 2023-05-04
ban useParams hook

utkubakir:


Brendonovich:
can we ban `useSearchParams` too plz :)

Brendonovich (approved):

PR #800, opened 2023-05-06
[ENG-592] Prevent libraries with blank names/no names from being created

brxken128:
This PR adds validation to prevent libraries with either: entirely empty names, or names that consist of purely whitespaces, from being created. It adds validation in both Rust and TS, and they both work flawlessly.

### DIFF
--- a/interface/app/onboarding/new-library.tsx
+++ b/interface/app/onboarding/new-library.tsx
@@ -71,7 +71,12 @@ export default function OnboardingNewLibrary() {
 						/>
 						<div className="flex grow" />
 						<div className="mt-7 space-x-2">
-							<Button type="submit" variant="accent" size="sm">
+							<Button
+								type="submit"
+								variant="accent"
+								disabled={!form.formState.isValid}
+								size="sm"
+							>
 								New library
 							</Button>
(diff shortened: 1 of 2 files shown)


interface/app/onboarding/new-library.tsx
```diff
@@ -13,7 +13,7 @@ import {
 import { useUnlockOnboardingScreen } from './Progress';
 
 const schema = z.object({
-	name: z.string().min(1, 'Name is required')
+	name: z.string().min(1, 'Name is required').regex(/[\S]/g)
```
utkubakir, line 16: Why not `trim()` here too? We don't wanna allow `libName   `. If that regex also does that, maybe adding a comment above might be helpful.

brxken128: In one of my commits I did actually add trim, but we don't trim anything else (I think) so I decided against adding it. Maybe we need a separate Linear issue to add trim everywhere (tag names being another major place).

I will add a comment about what the regex does - it only checks to see if the entire string is whitespace or not.

utkubakir (approved):

PR #802, opened 2023-05-08
[ENG-546] Improve error handling

oscartbeaumont:
Changes:
 - [x] Log to disk
 - [x] Add some more Clippy lints to enforce error handling.
 - [x] Button to export logs in the UI
 - [x] Fix `ErrorBoundary`
 - [x] Removing `unwrap` and `panic` calls


fogodev:
Just a quick note in this regard. But I had to change a bunch of error handling stuff on #781, like introducing a new `FileIOError` instead of plain `io::Error` for errors handling files, as using `?` on `io::Error` we have no data on the file that got the error.  Which is something crucial for my new walker implementation, given that instead of stopping on error, I collect every error that happen while indexing. I also changed many places to use this new error.

oscartbeaumont:
@fogodev How long until that other PR is going to be ready/merged? If it can be done as soon as possible that would be great and I will put this PR on hold until that then so we aren't modifying the same code.

fogodev:
Finally got #781 merged

utkubakir:
Are we going to use sentry to catch rust errors? I think we talked about it in Slack but not sure what you decided on

brxken128 (dismissed):
Everything looks awesome to me! Super happy about these changes, better error handling (and additional safety) is always welcome.

core/src/api/files.rs
```diff
@@ -109,7 +109,9 @@ pub(crate) fn mount() -> AlphaRouter<Ctx> {
 						.update(
 							object::id::equals(id),
 							vec![object::date_accessed::set(Some(
-								Utc::now().with_timezone(&FixedOffset::east_opt(0).unwrap()),
+								Utc::now().with_timezone(&FixedOffset::east_opt(0).expect(
+									"'0' can't possibly overflow the 'FixedOffset' bounds",
+								)),
```
Brendonovich, line 114:
```suggestion
								Utc::now().into()
```

Brendonovich (changes requested):

Brendonovich (dismissed):

Brendonovich (approved):

PR #803, opened 2023-05-08
Eng 510 open with in context menu

Brendonovich:
Implements Open With functionality on macOS, with support for Linux and Windows TBD.


brxken128 (dismissed):
Nice job! I can quite possibly give the Linux implementation a shot.

We may also need some mechanism to set a default app for an extension/file type, so the regular "Open" will use this app.

brxken128 (approved):

PR #805, opened 2023-05-09
[ENG-537] Notification system

oscartbeaumont:


oscartbeaumont:
This is ready to be merged but if both it and #1062 are merged at the same time we will brick the Prisma migrations.

Can we merge #1062 first!

Brendonovich (approved):

PR #807, opened 2023-05-09
remove unnecessary select styles and configure storybook

Brendonovich:


utkubakir:
Gotta make Eslint happy! Can we also make Clippy happy too? I know it didn't fail on this PR but looks like a easy fix

utkubakir (approved):

PR #812, opened 2023-05-12
Replace locations.getExplorerData with search.paths

Brendonovich:
As identified by @jamiepine, `getExplorerData` is just a subset of our existing `search` routes, so we may as well retire `locations.getExplorerData`


jamiepine:
Accidentally marked this as ready for review, noticed tags have not yet been upgraded

interface/app/$libraryId/location/$id.tsx
```diff
@@ -110,33 +111,38 @@ const useToolBarOptions = () => {
...
 			{
 				library_id: library.uuid,
-				arg: {
-					location_id,
-					path: explorerState.layoutMode === 'media' ? null : path,
-					limit,
-					kind: explorerState.layoutMode === 'media' ? [5, 7] : null
-				}
+				arg
 			}
```
utkubakir, line 136: why not inline arg? it would look cleaner

Brendonovich: didn't wanna use ternaries but ehh

oscartbeaumont (approved):

PR #815, opened 2023-05-13
[ENG-528] QuickPreview isn't correctly handling errors for video/audio playback

HeavenVolkoff:
- [X] Centralize the file preview logic in `Thumb.tsx`
 - [X] Rename `Thumb.tsx` main component to `FileThumb` to match mobile naming
 - [X] Fix `QuickPreview` not handling some file errors by offloading the error handling to `FileThumb`
 - [X] Improve `FileThumb`'s error handling
 - [X] Minor improvements to how `FileThumb` works in Inspector
 - [X] Change video borders to follow the video aspect-ratio, just like in Finder
 - [X] Add an internal `Thumbnail` component in `Thumb.tsx`
 - [X] Move the `getIcon` utility function to `assets/icons`
 - [X] Add a new `useCallbackToWatchResize` hook
 - [X] Replace `ResizeObserver` with the new resize hook in `FileThumb`
 - [X] Simplify and improve the `useIsDark` hook by replacing `react-responsive` with direct usage of WebAPI `matchMedia`
 - [X] Fix PDF preview completely breaking the application
   - [X] Allow Linux to access both the `spacedrive://` custom protocol and the workaround webserver
   - [X] On Linux, only use the webserver for audio and video, which don't work when requested through a custom protocol
   - [X] Configure tauri IPC to allow API access to the `spacedrive://localhost` domain to avoid PDF previews from breaking the security scope and rendering the application unusable
   - [X] Configure CSP to allow the PDF plugin custom protocol used by WebKit


interface/app/$libraryId/Explorer/File/Thumb.tsx
```diff
@@ -49,20 +51,26 @@ function Thumb({ size, cover, ...props }: ThumbProps) {
...
+	// Allows returning to thumbnails if the orignal fails to load
+	const [loadOriginal, setLoadOriginal] = useState<boolean>(false);
+
+	useEffect(() => {
+		setLoadOriginal(props.loadOriginal || false);
+	}, [props.loadOriginal, props.data]);
+
 	// Allows disabling thumbnails when they fail to load
-	const [useThumb, setUseThumb] = useState<boolean>(hasThumbnail);
+	const [loadThumb, setLoadThumb] = useState<boolean>(false);
```
Brendonovich, line 67: Could these and some of the above be merged into one `useState`? Would be much nicer to have a single union type than 4 separate values that could get into invalid states. Idk exactly how it all works so fairo if not.

Brendonovich (changes requested):

interface/app/$libraryId/Explorer/File/Thumb.tsx
```diff
@@ -39,156 +111,192 @@ export interface ThumbProps {
...
+		kind,
+		props.data.item.id,
+		isDir,
+		isDark,
+		cas_id,
+		library.uuid,
+		platform,
+		thumbType,
+		extension,
+		locationId
```
Brendonovich, line 169: Destructure `getExplorerItemData` inside the effect and you should be able to reduce 4 of these to 1 right?

Brendonovich (changes requested):

interface/app/$libraryId/Explorer/File/Thumb.tsx
HeavenVolkoff: `kind`, `extension`, `hasThumbnail` are used in other parts of the code. But if you don't mind repeating the call for `getExplorerItemData` I can do something like this:

```diff
diff --git a/interface/app/$libraryId/Explorer/File/Thumb.tsx b/interface/app/$libraryId/Explorer/File/Thumb.tsx
index 7283df86..f464a20e 100644
--- a/interface/app/$libraryId/Explorer/File/Thumb.tsx
+++ b/interface/app/$libraryId/Explorer/File/Thumb.tsx
@@ -121,10 +121,7 @@ function FileThumb({ size, cover, ...props }: ThumbProps) {
 	const [src, setSrc] = useState<string>('#');
 	const [thumbType, setThumbType] = useState(ThumbType.Icon);
 	const { locationId, newThumbnails } = useExplorerStore();
-	const { kind, extension, hasThumbnail, cas_id, isDir } = getExplorerItemData(
-		props.data,
-		newThumbnails
-	);
+	const { kind, extension, hasThumbnail } = getExplorerItemData(props.data, newThumbnails);
 
 	useEffect(() => {
 		if (props.loadOriginal) {
@@ -137,6 +134,8 @@ function FileThumb({ size, cover, ...props }: ThumbProps) {
 	}, [props.loadOriginal, props.data, hasThumbnail]);
 
 	useEffect(() => {
+		const { kind, extension, cas_id, isDir } = getExplorerItemData(props.data, newThumbnails);
+
 		switch (thumbType) {
 			case ThumbType.Original:
 				if (locationId) {
@@ -156,18 +155,7 @@ function FileThumb({ size, cover, ...props }: ThumbProps) {
 				setSrc(getIcon(kind, isDir, isDark, extension));
 				break;
 		}
-	}, [
-		kind,
-		props.data.item.id,
-		isDir,
-		isDark,
-		cas_id,
-		library.uuid,
-		platform,
-		thumbType,
-		extension,
-		locationId
-	]);
+	}, [props.data, newThumbnails, isDark, library.uuid, platform, thumbType, locationId]);
 
 	const childClassName = 'max-h-full max-w-full object-contain';
 	return (
```

utkubakir: 7 instead of 10, doesn't matter that much and calling the same function twice just seems unnecessary. You could probably combine them all if you didn't destructure them but meh

utkubakir (dismissed):

interface/app/$libraryId/Explorer/File/Thumb.tsx
HeavenVolkoff: I don't think not destructuring it would work as intended, considering `useExplorerData` always returns a new object in every call, which, if I am not mistaken, would result in react running the hook on every render, due to the object changing

Brendonovich: If it were me I'd change `getExplorerItemData` to `useExplorerItemData` and do a `useMemo` internally to avoid running the effect on every render but ehh it's not necessary

HeavenVolkoff: Did something like that in my last commit, but I didn't change `getExplorerItemData` directly, just used a `useMemo` on it inside the component.

Brendonovich (approved):

PR #822, opened 2023-05-15
[ENG-577 + ENG 608] - Index Rules UI Editor + Job Manager small tweaks

ameer2468:
This PR applies the new design and update of Index rules UI editor.

For the changes, please watch the video below:

**Notes**:

- The previous code was all in one file, mostly, so I also broke down the work into separate components.
- We now have a IndexerRuleEditor directory with the different components, cleaner structure.
- My goal was to also make this implementation easier to understand than the previous.

https://github.com/spacedriveapp/spacedrive/assets/33054370/660d6e81-b1c4-48cb-bd7c-1fbc68c10d6b

**Edit**: Updated as of latest commit

Screenshot 2023-05-16 at 1 57 44 PM

Screenshot 2023-05-16 at 2 27 55 PM


interface/app/$libraryId/settings/library/locations/IndexerRuleEditor/index.tsx
```diff
@@ -0,0 +1,51 @@
+import clsx from 'clsx';
```
utkubakir, line 1: Rename this file to `IndexerRuleEditor`. Index styling should be for pages not for components. (@Brendonovich can you confirm?)

interface/app/$libraryId/settings/library/locations/IndexerRuleEditor/RuleButton.tsx
```diff
@@ -0,0 +1,102 @@
...
+					(error) =>
+						showAlertDialog({
+							title: 'Error',
+							value: String(error) || 'Failed to add location'
+						})
+				)
+				.finally(() => {
+					setWillDelete(false);
+					setIsDeleting(false);
+				});
```
utkubakir, line 48: Can we do this properly? (with await and try-catch)

interface/app/$libraryId/settings/library/locations/IndexerRuleEditor/RulesForm.tsx
```diff
@@ -0,0 +1,326 @@
...
+			case 'Extension': {
+				const extensionRegex = /^\.[^.\s]+$/;
+				return {
+					value: extensionRegex.test(value),
+					message: value ? 'Invalid extension' : 'Value required'
+				};
+			}
+			case 'Name': {
+				return {
+					value: value ? true : false,
```
utkubakir, line 90: Maybe replace `value ? true : false` with `Boolean(value)` here?

utkubakir (changes requested):

interface/app/$libraryId/settings/library/locations/IndexerRuleEditor/index.tsx
ameer2468: If you take JobManager as an example, theres an index file there. I used that as reference.

interface/app/$libraryId/settings/library/locations/IndexerRuleEditor/RuleButton.tsx
ameer2468: I was going to do this, but looking at previous examples theres no async and await. Not sure why tho.

interface/app/$libraryId/settings/library/locations/IndexerRuleEditor/RulesForm.tsx
```diff
@@ -0,0 +1,326 @@
...
+							return `**/${value}`;
+						case 'Extension':
+							// .tar should work for .tar.gz, .tar.bz2, etc.
+							return [`**/*${value}`, `**/*${value}.*`];
+						default:
+							return value;
+					}
+				})
+			} as IndexerRuleCreateArgs;
+			createIndexerRules.mutateAsync(formatData);
```
HeavenVolkoff, line 144: This promise should be either awaited or returned, as it is right now backend errors will be ignored by the frontend

interface/app/$libraryId/settings/library/locations/IndexerRuleEditor/RulesForm.tsx
```diff
@@ -0,0 +1,326 @@
...
+			case 'Advanced': {
+				return {
+					value: value ? true : false,
+					message: 'Value required'
+				};
+			}
+			default:
+				return undefined;
+		}
+	};
```
HeavenVolkoff, line 109: If we're going to use our own custom validation logic, I think it would be better to migrate all the HTML native pattern validation to this function and remove it from RuleInput. Additionally, it might be better to export this function from RuleInput since it specifically handles validation for it.

interface/app/$libraryId/settings/library/locations/IndexerRuleEditor/RulesForm.tsx
```diff
@@ -0,0 +1,326 @@
...
+			} as IndexerRuleCreateArgs;
+			createIndexerRules.mutateAsync(formatData);
+			setToggleNewRule && setToggleNewRule(false);
+		},
+		[createIndexerRules, setToggleNewRule]
+	);
+
+	return (
+		<>
+			{createPortal(
```
HeavenVolkoff, line 152: I think there should be a comment here explaining why the portal is needed, just like in the previous version:
```
// The portal is required for Form because this component can be nested inside another form element
```

interface/app/$libraryId/settings/library/locations/IndexerRuleEditor.tsx
```diff
@@ -1,617 +0,0 @@
...
-									>
-										<Plus />
-									</Button>
-								</div>
-
-								<ErrorMessage
-									name={REMOTE_ERROR_FORM_FIELD}
-									variant="large"
-									className="mt-2"
-								/>
```
HeavenVolkoff, line 609: Same as above, this showed backend errors

interface/app/$libraryId/settings/library/locations/IndexerRuleEditor/RulesForm.tsx
```diff
@@ -0,0 +1,311 @@
...
+								className="mt-2"
+							/>
+						</div>
+						<div className="my-[25px] flex w-full flex-row items-center">
+							<label className="grow text-sm font-medium">
+								This group of index rules is a Whitelist{' '}
+								<Tooltip label="By default, an indexer rule acts as a deny list, causing a location to ignore any file that match its rules. Enabling this will make it act as an allow list, and the location will only display files that match its rules.">
+									<Info className="inline" />
+								</Tooltip>
+							</label>
```
HeavenVolkoff, line 268: Maybe this can be removed? I think the new toggle communicates its intent better than the previous checkbox, so this text has become quite redundant and may confuse users more than it helps.

jamiepine:
looking better for sure, can we clean up the look of the rule groups? hate the red X and the white border

interface/app/$libraryId/settings/library/locations/IndexerRuleEditor/RulesForm.tsx
ameer2468: You're right, i'll remove it.

ameer2468:
> 

I honestly don't like it either, ill see what I can do, maybe apply the design from the figma instead of this.

ameer2468:
> looking better for sure, can we clean up the look of the rule groups? hate the red X and the white border

ameer2468:
Latest @jamiepine  and everyone else. I'd like to propose this route, looks better and does whats needed. Also added a delete confirmation, because users could accidently click.

Screenshot 2023-05-16 at 1 57 36 PM

Screenshot 2023-05-16 at 1 57 44 PM

HeavenVolkoff (dismissed):

utkubakir:
- I don't like the whitelist/blacklist thing in the middle but we will soon have this moved to per rule basis so that can stay.
- Padding/spacing looks off with the save button
- `+ New` button is poorly positioned (the top of it should align with the indexer rules text)
- ^ and it's too wide
- Why do we have a permanent delete button? It can overlay when hovered.

And I think the rules look poor. You know we can make it a more significant component with an icon and with Info Pills which shows if it's a System tag and if it's enabled etc. It can also say how many rules are in it.

ameer2468:
> - I don't like the whitelist/blacklist thing in the middle but we will soon have this moved to per rule basis so that can stay.
> 
> - Padding/spacing looks off with the save button
> 
> - `+ New` button is poorly positioned (the top of it should align with the indexer rules text)
> 
> - ^ and it's too wide
> 
> - Why do we have a permanent delete button? It can overlay when hovered.
> 
> 
> 
> And I think the rules look poor. You know we can make it a more significant component with an icon and with Info Pills which shows if it's a System tag and if it's enabled etc. It can also say how many rules are in it.

The padding and spacing of the save button is correct. It is equal from the bottom and the top.

utkubakir:
> > * I don't like the whitelist/blacklist thing in the middle but we will soon have this moved to per rule basis so that can stay.
> > * Padding/spacing looks off with the save button
> > * `+ New` button is poorly positioned (the top of it should align with the indexer rules text)
> > * ^ and it's too wide
> > * Why do we have a permanent delete button? It can overlay when hovered.
> > 
> > And I think the rules look poor. You know we can make it a more significant component with an icon and with Info Pills which shows if it's a System tag and if it's enabled etc. It can also say how many rules are in it.
> 
> The padding and spacing of the save button is correct. It is equal from the bottom and the top.

image

as you can see, it's double here. Can you move the save button inside the divided box then?

jamiepine:
I think the padding is overall too much, it could comfortably be half around the outside. It doesn't need to match the padding of the location editor.

interface/app/$libraryId/settings/library/locations/IndexerRuleEditor/RuleInput.tsx
```diff
@@ -0,0 +1,164 @@
...
+									if (path) {
+										props.onChange?.(event);
+									}
+								})
+								.catch((error) =>
+									showAlertDialog({
+										title: 'Error',
+										value: String(error)
+									})
+								);
```
utkubakir, line 141: try-catch await please.

interface/app/$libraryId/settings/library/locations/IndexerRuleEditor/RulesForm.tsx
```diff
@@ -0,0 +1,294 @@
...
+				{
+					type: selectValues[0],
+					value: ''
+				}
+			]
+		}
+	});
+	const {
+		formState: { errors }
+	} = form;
```
utkubakir, line 62: This looks weird, just destructure above or do `form.formState.errors`

utkubakir (changes requested):

utkubakir (approved):

PR #824, opened 2023-05-16
[ENG-559] Spacedrop move to spaceblock

oscartbeaumont:


jamiepine (approved):
Its not impacting any active areas of the app, so this one can come right on through. LGTM.

PR #825, opened 2023-05-17
Fix Spacedrive process being `Killed` during lauch on macOS

HeavenVolkoff:
macOS (Ventura? arm?) requires all `dylibs` to be signed to allow a process to launch[¹].

This PR locally signs our ffmpeg `dylibs` with the machine's local certificate in the `setup-system` script after downloading them to work around this issue.

[¹]: https://github.com/tpoechtrager/osxcross/issues/355


brxken128 (approved):
Haven't tested but the changes seem sound!

PR #826, opened 2023-05-17
[ENG-662] Overview infinite scroll

jamiepine:
- Added infinate scroll to overview
- Fixed quick preview alignment
- Added Movie icon
- Fixed re-render bug with explorer items in the Overview


packages/assets/icons/index.ts
```diff
@@ -46,6 +46,8 @@ import Laptop from './Laptop.png';
 import Mesh from './Mesh.png';
 import Mesh_Light from './Mesh_Light.png';
 import Mobile from './Mobile.png';
+import Movie from './Movie.png';
+import Movie_white from './Movie_white.png';
```
brxken128, line 50: Should this be `Movie_Light` for consistency with the other names?

brxken128 (dismissed):

packages/assets/icons/index.ts
jamiepine: It should be, good catch.

brxken128 (approved):

PR #831, opened 2023-05-19
[ENG-624] Explorer order by

jamiepine:
added order by + fixed explorer padding situation

### DIFF
--- a/interface/hooks/useExplorerStore.tsx
+++ b/interface/hooks/useExplorerStore.tsx
@@ -1,6 +1,8 @@
 import { proxy, useSnapshot } from 'valtio';
-import { ExplorerItem } from '@sd/client';
+import { ExplorerItem, Ordering } from '@sd/client';
 import { resetStore } from '@sd/client/src/stores/util';
+
+type UnionKeys<T> = T extends any ? keyof T : never;
 
 export type ExplorerLayoutMode = 'rows' | 'grid' | 'columns' | 'media';
 
@@ -11,6 +13,10 @@ export enum ExplorerKind {
 }
 
 export type CutCopyType = 'Cut' | 'Copy';
+
+export type ExplorerOrderByKeys = UnionKeys<Ordering> | 'none';
+
+export type ExplorerDirection = 'asc' | 'desc';
 
 const state = {
 	locationId: null as number | null,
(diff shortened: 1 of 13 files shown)


utkubakir (approved):
lgtm

interface/hooks/useExplorerStore.tsx
```diff
@@ -35,7 +41,10 @@ const state = {
 	quickViewObject: null as ExplorerItem | null,
 	isRenaming: false,
 	mediaColumns: 8,
-	mediaAspectSquare: true
+	mediaAspectSquare: true,
+	orderBy: 'dateCreated' as ExplorerOrderByKeys,
```
Brendonovich, line 45: I think making this be `null` instead of `'none'` would be more appropriate

interface/hooks/useExplorerStore.tsx
```diff
@@ -1,7 +1,9 @@
 import { proxy, useSnapshot } from 'valtio';
-import { ExplorerItem } from '@sd/client';
+import { ExplorerItem, Ordering } from '@sd/client';
 import { resetStore } from '@sd/client/src/stores/util';
 
+type UnionKeys<T> = T extends any ? keyof T : never;
```
Brendonovich, line 5: Is this necessary? `keyof Ordering` should work

Brendonovich (changes requested):

jamiepine:
> I think making this be `null` instead of `'none'` would be more appropriate 

It's none so there can be a select option in the UI with "None"

interface/hooks/useExplorerStore.tsx
jamiepine: Screenshot 2023-05-18 at 10 52 22 PM

It doesn't sadly

interface/hooks/useExplorerStore.tsx
jamiepine: Screenshot 2023-05-18 at 10 52 22 PM

It doesn't sadly

Brendonovich: nvm union types are silly

interface/hooks/useExplorerStore.tsx
Brendonovich: ok fair

Brendonovich (approved):

PR #834, opened 2023-05-19
[ENG-603] Support HEIC images

jamiepine:
This PR adds thumbnail generation support for all HEIF (High Efficiency Image File Format) files.

This format is used by all modern iPhones, and the file type becomes HEIC (as they use the HEVC codec). If the AV1 codec is used, the file type becomes AVIF. This PR adds support for all 3 of these possible codecs.

HEIF containers can hold both a single stationary image, or they can store a collection/sequence of images (such as a live photo). We ignore the specific content of each file as we just use the main, "primary" image.

## Supported Extensions

This PR adds thumbnail support for:
- `heic`/`heics`<sup>1</sup>
- `heif`/`heifs`<sup>1</sup>
- `avif`/`avci`<sup>1</sup>/`avcs`<sup>1</sup>

<sup>1</sup>: untested so far due to lack of samples, but should still work.

### Unsupported Files

I have only ran into one single HEIF file that this PR can not convert, and that is due to limitations within `libheif` itself. The relevant error: `Unsupported file-type: File does not include any supported brands.`

This may be down to corruption within the file itself.

### Interleaved Only

We only support interleaved images so far - I have yet to run into/even find a file that has individual R/G/B channels, but I guarantee it will happen in production at some point.

As I can't find a sample that will work, I think the solution here will be to create one myself (using a simple Rust script which creates a HEIF image, sets the r/g/b data manually, and writes it).

I assume the re-encoding process will be similar to interleaved images, but we will be reading from separate channels instead of the same unified stream.

## The Conversion

We currently set an arbitrary file size cap of 20MiB. I have yet to see a HEIF/HEIC image surpass this, even those with very large resolutions and multiple images within the container.

For the conversion, we read the image while accounting for the `stride` of the image, in order to produce a non-stretched, correct-looking output. This is read into a sequence of `vec![r, g, b, r, g, b...]`.

Once we have the raw RGB data in the correct order, we convert it to a PNG file in memory (via the `png` crate). We use the bit depth provided by the source image, along with the appropriate width/height, in order to encode our RGB sequence into the PNG format. This is inefficient, but I have a solution in mind which should help to speed this process up<sup>2</sup>.

We then take this PNG data, and use the `image` crate to load it, as if it were any other PNG image ready to have a thumbnail generated. 

The WebP thumbnail is then generated as normal, and this is done via the usual route that all other images use.

<sup>2</sup>: we can potentially go from the RGB sequence straight to WebP, skipping both PNG and the `image` crate entirely. This does mean that we'll have two separate functions that both generate WebP thumbnails, but I feel as though the optimisation is worth it.

I have not found a suitable way to go from the RGB data directly to a type provided by the `image` crate (such as a `DynamicImage`), but I will keep looking into this.

## Testing

HEIF support is only enabled in Tauri by default, and is further restricted to MacOS and Linux only. Ensure that you re-run the setup script in order to get all required dependencies.

I've tested Apple Silicon builds in both dev and prod, and after removing all `brew` copies of `libheif` (so I'm sure the linking is working okay).


jamiepine:
Amazing work @brxken128. Tested it and it seems flawless. Let's focus on getting the libheif dep linked correctly as a priority before fixing up the smaller improvements you have planned. That way this PR is ready to merge and if the smaller issues don't make it you can make another PR.

brxken128:
> Amazing work @brxken128. Tested it and it seems flawless. Let's focus on getting the libheif dep linked correctly as a priority before fixing up the smaller improvements you have planned. That way this PR is ready to merge and if the smaller issues don't make it you can make another PR.

Fully agree! The linking will be the first task I work on tomorrow to try and get this ready for merge - things work well enough in the current state. As per @HeavenVolkoff's suggestion in Slack, this will be feature-gated to Linux and MacOS only for now (as we'll need to setup `libheif` builds for Windows).

Brendonovich (approved):
code looks good, i assume it works from jamie's testing

PR #839, opened 2023-05-20
[ENG-623] - List view obstruction fix

ameer2468:
- This fixes the list view, so that it sticks and doesn't go through.
- Also moved the keyboard delete event keyboard function to the Explorer component so that it works, as previously it was the in $id location, which is not the right place. Otherwise it wouldn't work in the overview page.

Screenshot:

Screenshot 2023-05-20 at 11 19 31 PM


interface/app/$libraryId/Explorer/ListView.tsx
```diff
@@ -340,7 +343,8 @@ export default () => {
 				onClick={(e) => e.stopPropagation()}
 				className={clsx(
 					'sticky top-0 z-20 table-header-group',
-					isScrolled && 'top-bar-blur !bg-app/90'
+					isScrolled && 'top-bar-blur !bg-app/90',
+					props.listViewHeadersClassName
```
utkubakir, line 347: why pass it 2 components up? just put it here. Also didn't Jamie's last pr fix this issue?

packages/client/src/core.ts
```diff
@@ -234,7 +234,7 @@ export type SpacedropArgs = { peer_id: PeerId; file_path: string[] }
 
 export type OwnedOperation = { model: string; items: OwnedOperationItem[] }
 
-export type ObjectWithFilePaths = { id: number; pub_id: number[]; kind: number; key_id: number | null; hidden: boolean; favorite: boolean; important: boolean; has_thumbnail: boolean; has_thumbstrip: boolean; has_video_preview: boolean; ipfs_id: string | null; note: string | null; date_created: string; date_accessed: string | null; file_paths: FilePath[] }
```
Brendonovich, line 237: this file is autogenerated by the core, you can't just change it. `location_id` isn't available on objects.

interface/app/$libraryId/Explorer/index.tsx
```diff
@@ -55,6 +58,20 @@ export default function Explorer(props: Props) {
...
+		if (!selectedItem) return;
+
+		dialogManager.create((dp) => (
+			<DeleteDialog
+				{...dp}
+				location_id={selectedItem.item.location_id}
+				path_id={selectedItem.item.id}
+			/>
+		));
+	});
```
Brendonovich, line 73: this only works bc of the addition to `core.ts`, run `pnpm prep` and then `pnpm typecheck` and this will not be valid.
this was in `/location/$id` before and not `/Explorer` exactly because it's location-specific, maybe if we turn it into a custom hook that takes in `location_id` and reuse it in both `/location/$id` and `/overview`?

Brendonovich (changes requested):

interface/hooks/useKeyDeleteFile.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+		if (!selectedItem || !location_id) return;
+
+		dialogManager.create((dp) => (
+			<DeleteDialog {...dp} location_id={location_id} path_id={selectedItem.item.id} />
+		));
+	});
+};
+
+export default useKeyDeleteFile;
```
utkubakir, line 18: Can we make this more generalized? I think `useKeyboardShortcuts` would be nice. Depending on what's passed we can enable/disable certain shortcuts here.

Brendonovich: Are there many other shortcuts that could be generalized? Imo this is the only one that makes sense to generalise at the moment. Besides, I think importing multiple hooks and configuring them individually would be nicer than importing one big hook and passing in a big config object

ameer2468: I think if there are multiple shortcuts then perhaps generalizing would be a good idea,5+ min. But currently there isnt. I think down the road tho I do see us taking inspiration from existing products and doing something great.

utkubakir: @Brendonovich you wouldn't pass a big config object, ideally this hook would enable certain shortcuts depending on what's passed and what screen are you on. Importing (and also having) 5-10 hooks to a component just seems dumb. Anyway, we have 10 `useKey` uses, so probably not a big deal now as you guys mentioned.

PR #841, opened 2023-05-21
[END-634] Coerce file extensions to lowercase

jamiepine:
There appears to be no benifit to preserving the file extention casing, especially only in production. Everything works great in dev, but in prod on macOS uppercase extensions are not compatible with Spacedrive, most notably the thumbnailer.

After some light research it appears all modern file systems are case insensitive, though this has only been tested in macOS. 

This PR removes the debug assertion encasing the lowercase coercion.

screenshot_2023-05-20_at_5 23 55_pm
_Dev VS Prod without this fix_


brxken128 (approved):
I think this means that `.MOV` files will display as `.mov` in our app? But that's not exactly the most major issue, and it's easier than changing _everywhere_ we match extensions.

LGTM!

PR #842, opened 2023-05-21
[ENG-635] Job time estimation

jamiepine:
Yummy job time estimation 

Screenshot 2023-05-20 at 9 52 49 PM

Note: For some reason generating the migrations dropped the `parent_id` column on file_paths, this is okay because it was deprecated, but it makes this migration a breaking change and is unrelated to this PR. I'd like to reset the migrations one last time once we release the first version this week.


ameer2468:
Love how this is actually being done by Rust, the backend.

brxken128 (approved):
the math and rust looks good to me! great job - hopefully it's a lot more accurate now haha

core/prisma/schema.prisma
```diff
@@ -385,8 +385,10 @@ model Job {
 
     parent_id Bytes?
 
-    task_count           Int       @default(1)
-    completed_task_count Int       @default(0)
+    task_count                   Int @default(1)
+    completed_task_count         Int @default(0)
+    estimated_remaining_seconds  Int @default(0)// Estimated remaining time in seconds
```
Brendonovich, line 390: I really don't feel good about having these in the DB, _especially_ `estimated_remaining_seconds`. It is purely ephemeral state that only exists while the job is running, I don't think it's the type of thing that needs to be persisted or synced via the regular sync system.

Brendonovich (changes requested):

core/prisma/schema.prisma
jamiepine: The whole point is that the user can view the job manager in real-time from all devices. This change does not change the frequency events are pushed, we're just adding data to it. The Job report is updated every task regardless of this PR, we can discuss improving it in the future.

Brendonovich: I'm not saying that `estimated_remaining_seconds` shouldn't be synced between nodes, i'm saying that I don't think it's the kind of thing that belongs in the regular sync system. I'd much prefer we send the value out-of-band via a separate system and don't store it in the database, also don't store how many seconds are left but rather what the estimated ending time is so that the frontend can calculate its own countdown. I guess the OOB stuff can happen later, but I'd definitely like to see it changed to use a timestamp rather than a count.

jamiepine: what's the difference between a timestamp and a count? both achieve the exact same result.

jamiepine: I suppose if a sync message is received late the time would still be accurate if it's a timestamp.

Brendonovich: I've expanded on my thoughts on job sync in #engineering, but yeah change it to a timestamp at least.

oscartbeaumont: A timestamp's are wayyy better for anything time-related. If we use a timestamp any React rerender will render the time between now and the timestamp and can display it. Trying to make a function that fires every exactly one second on the frontend to update the second count is surprisingly hard and will have drift (also React's strict mode and stuff are a thing). I also think the frontend should be updating the displayed value to the last known estimate in case the Tauri IPC bridge is lagging behind.

Brendonovich (approved):

PR #847, opened 2023-05-23
[ENG-613] Indexer rule whitelist/blacklist per rule

fogodev:
This PR introduces a semantic change of how Indexer Rules are represented, before we had that each Indexer Rule could have just one of the available kinds. Now each Indexer Rule can have many rules of each kind, according changes were made in the walker to enable this feature.


core/src/location/indexer/walk.rs
```diff
@@ -780,11 +744,17 @@ mod tests {
 			420,
 		)
 		.await
-		.unwrap()
-		.walked
-		.collect::<HashSet<_>>();
+		.unwrap();
+
+		if !walk_result.errors.is_empty() {
+			panic!("errors: {:#?}", walk_result.errors);
```
brxken128, line 750: is `panic!`ing the best thing we can do here? seems unfitting given all of the new error handling stuff

brxken128 (commented):
looks pretty good to me, just a little curious about the `panic!`s (they may be fine though tbh) 🙏🏻

HeavenVolkoff:
Migration seems to be broken:
image

HeavenVolkoff:
Tested here and everything seems to be working fine, except for the migration issue

core/src/location/indexer/walk.rs
fogodev: These are on unit tests, so I think it's alright

fogodev:
> Migration seems to be broken: image

Looks like this one is caused by the "migrator" on dev builds, and not by the migrations files. As I actually introduced a breaking change on how indexer rules are stored on database, now storing bytes representing a vector of enums for each rule kind, it isn't backward compatible anyway, so you have to remove your database. Otherwise we would have to start handling data migrations for incompatible data

HeavenVolkoff:
> Looks like this one is caused by the "migrator" on dev builds, and not by the migrations files. As I actually introduced a breaking change on how indexer rules are stored on database, now storing bytes representing a vector of enums for each rule kind, it isn't backward compatible anyway, so you have to remove your database. Otherwise we would have to start handling data migrations for incompatible data

Hun... this is interesting. Prod would also require the user to delete the db, right? Maybe we should handle data migration then. In this specific case, maybe just dropping the old table would be enough. @Brendonovich what you think?

fogodev:
> > Looks like this one is caused by the "migrator" on dev builds, and not by the migrations files. As I actually introduced a breaking change on how indexer rules are stored on database, now storing bytes representing a vector of enums for each rule kind, it isn't backward compatible anyway, so you have to remove your database. Otherwise we would have to start handling data migrations for incompatible data
> 
> Hun... this is interesting. Prod would also require the user to delete the db, right? Maybe we should handle data migration then. In this specific case, maybe just dropping the old table would be enough. @Brendonovich what you think?

Nope, as on prod we actually use the migrations, and in the migrations the old indexer rules table is dropped.

HeavenVolkoff:
> Nope, as on prod we actually use the migrations, and in the migrations the old indexer rules table is just dropped, without trying to migrate the old incompatible data. And we also repopulate the indexer rules table with basic rules in the seeder, using the new data format.

But that would mean the user will lose his custom indexer rules, no? What would happen if there are locations using them?

core/src/location/indexer/walk.rs
brxken128: that is a great point haha

fogodev:
> > Nope, as on prod we actually use the migrations, and in the migrations the old indexer rules table is just dropped, without trying to migrate the old incompatible data. And we also repopulate the indexer rules table with basic rules in the seeder, using the new data format.
> 
> Actually, dropping isn't a good solution, as it would mean the user will lose his custom indexer rules. What would happen if there are locations using them?

Yeah, if we already had a stable release to maintain, I would totally agree that custom indexer rules are important, and we should have some kind of data migration to keep them. But it isn't the case yet.

Brendonovich:
Gonna chime in to say that right now migrations aren't important and we can mangle them as much as we want, they'll be squashed into a single file once we go stable anyway

HeavenVolkoff (approved):
Tested again after the merge with main. Everything seems to be working fine. Not auto-merging yet just to wait the mobile test to pass, and also if someone else wants to review it...

PR #853, opened 2023-05-24
[ENG-652] Improved job reports

jamiepine:


jamiepine:
A lot of the frond-end code in this PR won't merge well with my Job PR, so careful when you come back to this. Seems to be lots of conflict with main too.

brxken128:
> A lot of the frond-end code in this PR won't merge well with my Job PR, so careful when you come back to this. Seems to be lots of conflict with main too.

Thanks for the heads up! I don't mind re-doing it if it comes down to it, wasn't too difficult 🙏🏻

PR #854, opened 2023-05-24
[ENG-651] Fix PDF rendering breaking app on macOS

### LINKED ISSUE #849: [MacOS] Quick View on PDFs Crashes App.
### Describe the bug

When clicking `Space` on a PDF to initiate Quick View, the app becomes unresponsive and doesn't even let you quit with `Cmd + Q`.

https://github.com/spacedriveapp/spacedrive/assets/60551230/36188416-445f-4558-8860-e741cab45e80

### Reproduction

1. Find a PDF file within the viewer.
2. Select it
3. Press Space
4. App doesn't respond anymore.

### Expected behavior

The PDF view is shown in the quick view modal.

### Platform and versions

```Shell
`pnpm --version && cargo --version && rustc --version`:
9.6.6

OS:
macOS 12.6.2 21G320 arm64

Chip:
Apple M1 Max

Spacedrive Version:
v0.1.0 - 2325a8c
```

### Stack trace

_No response_

### Additional context

_No response_

HeavenVolkoff:
- [X] Fix macOS PDF rendering
	- [X] Fix app crashing due to PDF component receiving empty URLs
	- [X] Use `<iframe>` instead of `<embed>` or `<object>` to load PDFs on macOS
	- [X] Don't send partial ranges for PDFs
- [X] `QuickPreview` improvements
	- [X] Fix `QuickPreview` not closing with space bar
	- [X] Minor improvements to `QuickPreview` header style
	- [X] Fix Button inside Button react error in `QuickPreview`
- [X] `Thumb` improvements
	- [X] Don't render thumb data without a valid source
	- [X] Fix `Thumb` not rendering original file in `Overview`
- [X] Fix double-click simultaneously renaming and opening file
- [X] Fix `AddLocationDialog` sometimes requiring multiple confirmations on first load
- [X] Rename `ExternalObject` to `PDFViewer`
	- [X] Improve how the `link` load and error events are handled

Closes #849


interface/app/$libraryId/settings/library/locations/AddLocationDialog.tsx
```diff
@@ -76,8 +76,11 @@ export const AddLocationDialog = ({
 	useEffect(() => {
 		// Update form values when default value changes and the user hasn't made any changes
 		if (!form.formState.isDirty)
-			form.reset({ path, method, indexerRulesIds }, { keepErrors: true });
-	}, [form, path, method, indexerRulesIds]);
+			form.reset(
+				{ path, method: form.getValues().method, indexerRulesIds },
```
utkubakir, line 80: What's wrong with `method` here? Why are you getting values?

utkubakir (approved):

interface/app/$libraryId/settings/library/locations/AddLocationDialog.tsx
HeavenVolkoff: This is a fun one. This change fixes the issue where the `AddLocationDialog` sometimes requires multiple confirmations on first load. The reason for this is a race condition between these two hooks:
https://github.com/spacedriveapp/spacedrive/blob/ea46e7736a26157757e8a6e34a1a2d6a8ee9e167/interface/app/%24libraryId/settings/library/locations/AddLocationDialog.tsx#L76-L80

https://github.com/spacedriveapp/spacedrive/blob/ea46e7736a26157757e8a6e34a1a2d6a8ee9e167/interface/app/%24libraryId/settings/library/locations/AddLocationDialog.tsx#L146-L166

During the first render, both of these hooks run concurrently. If `useCallbackToWatchForm` finishes executing before the `useEffect`, the `method` form value may be changed due to a backend error received by the `dryRun` execution of `addLocation`. However, when the `useEffect` finally executes, it would reset the `method` value to the one received from the component props. Which would result in the user needing to click "add" twice in the dialog. To fix this, I changed the `useEffect` to read the current `method` value from the form to account for this.

PR #865, opened 2023-05-25
[ENG-383] Use system tags as favourites backend

Brendonovich:
- [x] Backend (bad atm)
- [ ] Frontend 

Closes ENG-375


jamiepine:
After testing I've concluded the impact to query time on the Explorer being doubled (for favorites and hidden) is not worth the improvement to query syntax of search & created timestamps.

Something still feels off about the reserved tag pub_ids. I'm not quite understanding why the ids couldn't just be hard coded such that favorites is always X, every node would know favorites is X. You wouldn't need to reserve a certain amount because X would always be taken by favorites. 

We could potentially make use of system tags, but I had planned labels to be system _assigned_ tags, they wouldn't be treated the same as tags in the UI however. They're mainly intended for future AI object classification, but we could use them to identify downloads, screenshots and other context. I bring this up because maybe some of the work on this PR can be ported to labels. But it's not a priority.

As for search, we can add a two more boolean parameters to the query `favorite` and `include_hidden`, and keep the existing fields on Object. It's the cleanest way to achieve it, and the timestamp an Object was favorited or hidden isn't that important in retrospect.

Brendonovich:
> I'm not quite understanding why the ids couldn't just be hard coded such that favorites is always X

They are

```rust
pub const FAVORITES_TAG: SystemTag = SystemTag {
  pub_id: 0,
  name: "Favorites",
  color: "FAC607",
};
```

> You wouldn't need to reserve a certain amount because X would always be taken by favorites

Yes, but what if a user-generated tag is generated with the same ID. Would rather that just never happens than to have a unique constraint violation error.

jamiepine:
Labels will not be queried by the explorer, whereas the favorite bool is always returned on every Explorer Item.
 
They are effectively what system tags would be, but I don't want to use them for favorite/hidden.

We can close this PR but there are surely some parts that could be salvaged.

PR #872, opened 2023-05-26
[ENG-650] - Blacklist/whitelist each rule

ameer2468:
With the recent index rules change, we can now blacklist/whitelist each rule individually.

**This PR updates the UI to do just that**:

edit: After PR Tweaks:

Screenshot 2023-05-26 at 11 46 07 PM


utkubakir (dismissed):

Brendonovich:
> I think dropdowns are ideal when you have more than 2 options, however, in this case you're switching between just 2 values. Hence why I think switch works well here.

I guess, but it feels like more UI elements than necessary. IMO switches should be strictly for booleans, not for enums.

> From a user perspective, what does 'Kind' even mean?

Maybe 'Kind' isn't descriptive enough, but 'Mode' can be. Could make the same argument for 'Allow' though, and you've already got a help widget for that.

> The user can't see whats in the backend

Not necessarily - the data that's being provided to this form is exactly what's going into the backend. While it's good to keep things simple, it's not unreasonable to expect someone writing custom indexer rules to learn the difference between the reject and allow kinds. We already require some initial learning when you start using the app with the distinction between Locations, Nodes and Libraries, complexity can be ok.

utkubakir:
I like `Mode` and Block/Allow for the switch. `Kind` feels a completely different thing to me, but that might be the language barrier.

A dropdown would work here but the switch does make this a bit fancier.

ameer2468:
> I like `Mode` and Block/Allow (or Reject/Allow) for the switch. `Kind` feels a completely different thing to me, but that might be the language barrier.
> 
> A dropdown would work here but the switch does make this a bit fancier. It's already a boring sad table, why not add something to flick here and there

I do have to agree that a switch makes things more visually interesting/appealing. There's already a dropdown here, adding another one would work, but from a visual perspective I don't think it's great.

Brendonovich (approved):

PR #886, opened 2023-05-30
[ENG-662] Don't use job system for shallow operations

Brendonovich:
Creating dedicated jobs just for shallow operations is not necessary since they don't need to be persisted or resumed


fogodev (approved):
LGTM

PR #893, opened 2023-05-31
[ENG-677] Error button tweak for reset lib

ameer2468:
Quick styling change

**Screenshot**:

Screenshot 2023-05-31 at 12 25 21 PM


Brendonovich (approved):
we _really_ need to start managing this stuff inside our component library rather than tacking on styles after the fact, good enough for now tho

PR #896, opened 2023-05-31
[ENG-534] Themes UI & appearance

ameer2468:
Took design from mobile and applied to the appearance UI aswell.

- Enabled appearance page
- Theme cards
- Adjusted File icon size so it looks correct
- A few elements required using colors from the theme.
- Hue slider, changes the hue of theme.

**Tweaks not related to this PR, but minor**:

- Feedback dialog, emoji de-selection since it's optional.

Screenshot 2023-06-05 at 9 44 18 PM

Screenshot 2023-05-31 at 6 31 27 PM


interface/app/$libraryId/settings/client/appearance.tsx
```diff
@@ -25,31 +73,185 @@ export const Component = () => {
...
+		<>
+			<Form form={form} onSubmit={onSubmit}>
+				<Heading
+					title="Appearance"
+					description="Change the look of your client."
+					rightArea={
+						<div>
+							<Button
+								disabled={
+									themeStore.theme === 'dark' && themeStore.hueValue[0] === 235
```
utkubakir, line 110: Why the 0 here? `themeStore.hueValue[0]`

packages/client/src/stores/themeStore.ts
```diff
@@ -4,9 +4,9 @@ import { valtioPersist } from './util';
 export type Themes = 'vanilla' | 'dark';
 
 const themeStore = valtioPersist('sd-theme', {
-	theme: 'vanilla' as Themes,
+	theme: 'dark' as Themes,
 	syncThemeWithSystem: false,
-	hueValue: null as number | null
+	hueValue: [235],
```
utkubakir, line 9: again why is this an array?

utkubakir (changes requested):

packages/client/src/stores/themeStore.ts
ameer2468: The slider component expects a number within an array.

interface/app/$libraryId/settings/client/appearance.tsx
ameer2468: So we can take the first value within the array, in this case 235.

brxken128 (approved):
I'm happy with it and all changes work great, even changing system appearance works instantly. Also tested the file icon size changes, and the new feedback stuff is awesome.

Great work, as usual!

utkubakir (approved):

PR #917, opened 2023-06-06
[ENG-685] Sqlite Multithreaded

oscartbeaumont:
IDK if this is actually gonna help but it can't hurt.


fogodev:
We already enabled the WAL pragma on SQLite?

oscartbeaumont:
I tried using pragma's in a raw execute statement but Prisma doesn't like it due to something about its return type. From the upstream issues I am fairly sure the param I have set here will make it so each query waits instead of just erroring if it can't acquire the DB lock.

PR #925, opened 2023-06-07
[ENG-708] Thumbnail sharding

jamiepine:
This is a simple PR to improve the way we store thumbnails in the data folder. After a nice chat with ChatGPT I discovered why most apps do this, I noticed even Apple apps store files internally in hex-coded folders to prevent directory sizes from growing too large. Since my thumbnail folder is well over 85,000 items and takes a minute to open in Finder it could probably become an issue faster than we'd expect. This is an industry standard format for storing large collections of cached files. 

`calc_shard_hex` takes a cas_id as input, computes a blake3 hash of the filename, and returns the first two characters of the hash as the directory name. Because we're using the first two characters of a blake3 hash, this will give us 256 (16*16) possible directories, named 00 to ff.

This PR also introduces the `version_manager` util which allows us to easily migrate app data structures with a rust version enum and a `version.txt` that contains a single int32. I wrote a migration for the thumbnail folder that moves existing `webp` files into their respective shard directory.

Screenshot 2023-06-07 at 10 05 07 PM

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -69,7 +69,7 @@ version = "0.7.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9e8b47f52ea9bae42228d07ec09eb676433d7c4ed1ebdf0f1d1c29ed446f1ab8"
 dependencies = [
- "cfg-if",
+ "cfg-if 1.0.0",
  "cipher 0.3.0",
  "cpufeatures",
  "opaque-debug",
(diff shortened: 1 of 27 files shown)


oscartbeaumont:
What is the purpose of hashing the cas_id, isn't the cas_id already a hash?

jamiepine:
> What is the purpose of hashing the cas_id, isn't the cas_id already a hash?

You are so correct, it is indeed already hexadecimal. Fixed!

brxken128 (approved):
sweet work boss

PR #927, opened 2023-06-07
[ENG-381] Update context menufs actions to support multiple source

fogodev:
This PR introduces the feature to pass multiple file at same time to the backend for multiple purposes, like opening, copying, moving, renaming, encrypting and decrypt files. Still needs some frontend support.


apps/desktop/src/commands.ts
```diff
@@ -2,37 +2,37 @@
...
 }
 
 export function getFilePathOpenWithApps(library: string, id: number) {
-    return invoke()<OpenWithApplication[]>("get_file_path_open_with_apps", { library,id })
+	return invoke()<OpenWithApplication[]>("get_file_path_open_with_apps", { library, ids: [id] })
 }
 
 export function openFilePathWith(library: string, id: number, withUrl: string) {
-    return invoke()<null>("open_file_path_with", { library,id,withUrl })
+	return invoke()<null>("open_file_path_with", { library, fileIdsAndUrls: [[id, withUrl]] })
```
Brendonovich, line 34: was this done manually? i'm pretty sure tauri-specta isn't smart enough to do this, and this file shouldn't be edited

fogodev: Yeah, I did this change manually, haven't noted the comment in the file's beginning LMAO. Sorry about that, will fix ASAP. Which command I have to run to generate this file again, and if tauri-specta couldn't handle the types, how I can have proper typing in this file?

Brendonovich: All good, just run the desktop app and it'll generate new types. If `tauri-specta` can't handle the types then Specta needs more work haha.

utkubakir:
This will most likely break the app and as far as I know, multi-select isn't fully ready yet (right @niikeec ?) so maybe we should postpone this to after alpha release.

core/src/library/library.rs
```diff
@@ -100,27 +101,38 @@ impl Library {
...
+				out.insert(
+					file_path.id,
+					file_path.location.path.as_ref().map(|location_path| {
+						Path::new(&location_path).join(IsolatedFilePathData::from((
+							file_path.location.id,
+							&file_path,
+						)))
+					}),
+				);
+			});
```
Brendonovich, line 134: would be cool to `map` the results of the query and `chain` onto the iterator that makes the `HashMap`

Brendonovich (changes requested):
not a fan of `= { workspace = true }` but ehh

fogodev:
> not a fan of `= { workspace = true }` but ehh

My linter here gets really mad at `.workspace = true` lol

core/src/library/library.rs
fogodev: No can do. The `ids` vec stays borrowed if you don't consume the first iterator so `ids` can be moved into the query, otherwise we have to clone the vec which is even worse

Brendonovich: I think cloning the vec is fine but if you really don't want to do that then `out.extend` combined with using `map` instead of `for_each` works too

HeavenVolkoff:
> > not a fan of `= { workspace = true }` but ehh
> 
> My linter here gets really mad at `.workspace = true` lol

We could use this form instead:
```
[package.license]
workspace = true

[package.repository]
workspace = true

[package.edition]
workspace = true
```

Should work the same, and be more in line with the YAML syntax

PR #929, opened 2023-06-08
Fix default system rules

Brendonovich:
`default` is now properly set, plus the rules seeder is ran on library load

### DIFF
--- a/core/src/library/manager.rs
+++ b/core/src/library/manager.rs
@@ -399,6 +399,8 @@ impl LibraryManager {
 		let key_manager = Arc::new(KeyManager::new(vec![]).await?);
 		seed_keymanager(&db, &key_manager).await?;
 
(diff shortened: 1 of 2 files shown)


core/src/library/manager.rs
```diff
@@ -399,6 +399,8 @@ impl LibraryManager {
 		let key_manager = Arc::new(KeyManager::new(vec![]).await?);
 		seed_keymanager(&db, &key_manager).await?;
 
+		rules::seeder(&db).await?;
```
utkubakir, line 402: Why do we need to run a seeder on every library load? Seems unecessarry

Brendonovich: https://linear.app/spacedriveapp/issue/ENG-501/seeder-should-update-default-indexer-rules

utkubakir (approved):

PR #930, opened 2023-06-08
[ENG 709] Make all location fields optional

Brendonovich:
This makes all fields of `Location` except `id` and `pub_id` optional, in preparation to make everything optional. 
Reasoning as detailed in Slack
> Had a chat with Jordan (VapeJuiceJordan) yesterday about sync and managing data. I think it'd be a good idea for us to go through the Prisma schema and make everything optional except for IDs.
At Netflix there are entire rules just for enforcing that no data is made required, and with sync just existing I think it'd be a good move for us even though we're not Netflix. We use `@default` a bunch anyway which I would say is worse than making things optional - at least with everything being optional we can choose what to display as fallback data, `@default` means we can't do that.

> Making everything optional would allow for some sync messages to be ingested, but leave others until the app is updated, allowing for some cross-version compatability

Would appreciate @utkubakir and @jamiepine looking at this one, former to make sure mobile changes have been done correctly, latter to be aware of this move bc idk if he saw the Slack convo.


utkubakir:
I am gonna leave this up to you Jamie. I understand the need for having optional fields for future stuff we add but I can't wrap my head around changing everything and removing all the @default values. For example why date_created is optional and not set to now by default?

jamiepine:
I approve this, will merge if it has been thoroughly tested end-to-end against the latest main, PRs like this are prime culprits for new bugs.

oscartbeaumont (approved):

PR #931, opened 2023-06-09
Feat: add search using SearchMate

JosueRhea:
I have added a search feature in the documentation using [Search Mate](https://searchmate.app/).

To use this feature, you will need to provide an `appId`, which I have included in the environment variables. But for this app i wanna give the appId `0a85eacc-dafe-4381-ac25-78765241c348`. It can be public so don't worry :).

image

You can find the preview deployment [here](https://spacedrive-landing-eta.vercel.app/docs)


utkubakir:
Thank you for contributing, I am guessing this is your service so that's cool. 

I couldn't see any logic to upload our docs and their content programmatically and that would be a requirement for the search in docs. We are also not looking to pay as Agolia (or flexsearch) is free for open source and looks like SearchMate isn't.

JosueRhea:
Yeah, it indexes the documentation from the GitHub repo, rebuilds the docs based on commits and all that stuff.

For this project, I can make it free. But yes, it still makes sense to use Algolia, as you said

PR #932, opened 2023-06-10
[ENG-691] Comment/remove crypto code from `main` 

brxken128:
This PR removes all crypto code from the core and frontend, so that when #610 and their subsequent issues get merged there won't be any breaking changes.

A lot of these changes are just comments (as they'll likely be re-used), or some functions were removed entirely due to redundancy and replacement within the new refactor.

I have not tested this on mobile (as I don't have the setup), but I have made an attempt to keep things good there too.


interface/app/$libraryId/settings/Sidebar.tsx
```diff
@@ -80,10 +80,10 @@ export default () => {
 						<Icon component={TagSimple} />
 						Tags
 					</SidebarLink>
-					<SidebarLink to="library/keys">
+					{/* <SidebarLink to="library/keys">
 						<Icon component={Key} />
 						Keys
-					</SidebarLink>
+					</SidebarLink> */}
```
utkubakir, line 86: Instead of commenting this out, let's add `disabled` instead

brxken128: Good call!

utkubakir (approved):
LGTM

PR #934, opened 2023-06-12
[ENG-698] Stop leaking pk

oscartbeaumont:


core/src/node/config.rs
```diff
@@ -31,6 +31,32 @@ pub struct NodeConfig {
 	pub p2p_img_url: Option<String>,
 }
 
+// A version of [NodeConfig] that is safe to share with the frontend
+#[derive(Debug, Serialize, Deserialize, Clone, Type)]
+pub struct SanitisedNodeConfig {
```
Brendonovich, line 36: sucks but good enough for now

Brendonovich (approved):

PR #935, opened 2023-06-12
[ENG-714] 'Open with' not on folders

ameer2468:
Quick change that removes 'Open with' option from folders

Screenshot 2023-06-12 at 11 57 16 AM


interface/app/$libraryId/Explorer/File/ContextMenu/OpenWith.tsx
```diff
@@ -8,19 +8,22 @@ export default (props: { filePath: FilePath }) => {
...
+							actions={{
+								getFilePathOpenWithApps,
+								openFilePathWith
+							}}
+						/>
+					</Suspense>
+				</ContextMenu.SubMenu>
+			)}
+		</>
 	);
```
utkubakir, line 27: Before the return, do `if props.filePath.is_dir return null. Would be cleaner.

utkubakir (approved):

PR #945, opened 2023-06-14
[ENG-594] `Open With` Windows + fixes

HeavenVolkoff:
- [X] Windows `Open With`
   - [X] Implement logic to list applications capable of opening a certain file, based on its extension, using [`SHAssocEnumHandlers`](https://learn.microsoft.com/en-us/windows/win32/api/shobjidl_core/nf-shobjidl_core-shassocenumhandlers)
   - [x] Implement logic to execute a given application to open a file
   - [X] Fix `Open With` behavior when no apps are available
- [X] Add update step for VisualStudio Installer in windows setup-script.sh to fix it failing to execute due to outdated installations
- [X] Fix windows setup-script.sh being rate limited in CI runs
- [X] ~Fix App UI crashing on launch in Windows due to Edge Webview being stricter about CSS values~ (Fixed in main already)
- [X] Fix Windows, Linux and macOS `Open With` for compatibility with #927
- [X] Sort Linux file handlers entries to ensure a consistent app order in the `Open With` submenu
- [X] Make `get_file_path_open_with_apps` lenient with underline OS Api errors to allow a best-effort at returning something in the app list

### DIFF
--- a/apps/desktop/src-tauri/src/file.rs
+++ b/apps/desktop/src-tauri/src/file.rs
@@ -50,23 +50,17 @@ pub async fn open_file_path(
 }
 
 #[derive(Serialize, Type)]
-#[serde(tag = "t", content = "c")]
-#[allow(dead_code)]
-pub enum OpenWithApplication {
-	File {
-		id: i32,
-		name: String,
-		#[cfg(target_os = "linux")]
-		url: std::path::PathBuf,
-		#[cfg(not(target_os = "linux"))]
-		url: String,
-	},
-	Error(i32, String),
+pub struct OpenWithApplication {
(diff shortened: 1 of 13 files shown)


apps/desktop/src-tauri/src/file.rs
fogodev, line 1: I see that you're not using the `OpenWithApplication::Error` variant anymore, just logging errors in place and returning empty vecs in case of errors. This way you can drop this variant and change `OpenWithApplication` to be a struct.

fogodev (changes requested):

fogodev (approved):

PR #946, opened 2023-06-14
[ENG-734] Hide releases

utkubakir:


Brendonovich:
shouldn't we just get rid of the workflow outright if we only want to run it privately?

Brendonovich (approved):

PR #947, opened 2023-06-14
Optional filepath + object fields

Brendonovich:
optional all the things

plz test and check default values


core/src/location/file_path_helper/isolated_file_path_data.rs
```diff
@@ -24,12 +24,12 @@ static FORBIDDEN_FILE_NAMES: OnceLock<RegexSet> = OnceLock::new();
...
-	pub(in crate::location) is_dir: bool,
-	pub(in crate::location) name: Cow<'a, str>,
-	pub(in crate::location) extension: Cow<'a, str>,
-	pub(in crate::location) relative_path: Cow<'a, str>,
+	pub location_id: location::id::Type,
+	pub materialized_path: Cow<'a, str>,
+	pub is_dir: bool,
+	pub name: Cow<'a, str>,
+	pub extension: Cow<'a, str>,
+	pub relative_path: Cow<'a, str>,
```
fogodev, line 32: You really should take caution on this, these fields were private to the rest of core because they're built following some specific logic here on location module. The accessors methods are to be read only. Using as public fields and writing new values without following the current logic for materialized paths will introduce hard to debug errors down the line.

core/src/object/mod.rs
```diff
@@ -16,7 +16,7 @@ pub mod validation;
 // Objects are what can be added to Spaces
 
 // Object selectables!
-object::select!(object_just_id_has_thumbnail { id has_thumbnail });
+object::select!(object_just_id_has_thumbnail { id });
```
fogodev, line 19: Change name plz

fogodev (changes requested):

fogodev (approved):

PR #959, opened 2023-06-16
[ENG-715] Fix file renaming

niikeec:
Fixes: ENG-720, ENG-715


HeavenVolkoff:
There is still some weird behavior when renaming in the ListView. After renaming and unselecting a Row the file revert to the old name, but if you do a full re-scan it updates to the new name.

HeavenVolkoff (approved):

PR #966, opened 2023-06-18
feat: Updated the background on the landing page to use react-three-f…

MarquesCoding:
This PR is to change the background of the landing page.

I'm unsure how you guys import images, so you'll have to tweak it + the fading animation is buggy so you'll have to change that too, but this PR is just a concept for you to look at :)

Thanks

Ps.  Hopefully I've imported the packages correct, you have your project setup different from my work's one 😆

### DIFF
--- a/apps/landing/package.json
+++ b/apps/landing/package.json
@@ -13,6 +13,8 @@
 		"@auth/core": "^0.7.1",
 		"@aws-sdk/client-ses": "^3.337.0",
 		"@planetscale/database": "^1.7.0",
+		"@react-three/drei": "^9.76.0",
+		"@react-three/fiber": "^8.13.3",
 		"@sd/assets": "workspace:*",
 		"@sd/ui": "workspace:*",
 		"@t3-oss/env-nextjs": "^0.3.1",
@@ -40,6 +42,7 @@
 		"remark-gfm": "^3.0.1",
 		"remark-math": "^5.1.1",
 		"sharp": "^0.32.1",
+		"three": "^0.153.0",
 		"tsparticles": "^2.9.3",
 		"uuid": "^9.0.0",
 		"zod": "^3.21.4"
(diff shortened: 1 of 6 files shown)


utkubakir:
I fixed the main repo issues and the code looks alright. However, when the web app is mounted (not the image), the container looks off.

image

utkubakir:
> I did throw a DM but presume it wasn't read 🤷
> 
> "I just noticed that you might have to update it because I only applied it to the placeholder image, so it won't apply the glow to the actual iframe"

My bad didn't see the message. I looked at the code briefly but couldn't figure out how everything worked lol. You can run `pnpm web dev` to run the embedded web app to fix it if you want

MarquesCoding:
> > I did throw a DM but presume it wasn't read 🤷
> > "I just noticed that you might have to update it because I only applied it to the placeholder image, so it won't apply the glow to the actual iframe"
> 
> My bad didn't see the message. I looked at the code briefly but couldn't figure out how everything worked lol. You can run `pnpm web dev` to run the embedded web app to fix it if you want

When doing this, I get a weird error: `note: to see what the problems were, use the option `--future-incompat-report`, or run `cargo report future-incompatibilities --id 1`
     Running `target\debug\server.exe`
error: process didn't exit successfully: `target\debug\server.exe` (exit code: 0xc0000135, STATUS_DLL_NOT_FOUND)
`

utkubakir (approved):
LGTM, there is this insane lag when trying to resize the window but since this is temporary I think we can go with it.

MarquesCoding:
Hmm, are you able to record this lag? I can’t reproduce this :/

MarquesCoding:
Uh, I also don’t see the image showing??

utkubakir:
> Uh, I also don’t see the image showing??

Yeah, I don't see the image as well. The web app should hide after 5 minutes but it's not being replaced by the image weird

utkubakir (approved):

PR #984, opened 2023-06-21
[ENG-777] Revert disabled categories

niikeec:


utkubakir:
I still think categories that don't work shouldn't be clickable by the user. If we want them to show up in UI, let's implement the backend for them instead. At worst, seeing this ugly thing will force us to reconsider it every time we boot it up.

PR #991, opened 2023-06-21
Updater channels

Brendonovich:
Adds server support for updating on a specified release channel (`alpha` or `stable`), or picking any random version. Uses github's API and some caching to fetch data about releases via tags. Desktop atm is hardcoded to `alpha`, will need some UI and extra Tauri logic to support switching.


Brendonovich:
this gon be harder than i thought 😳

oscartbeaumont (approved):

PR #992, opened 2023-06-21
[ENG-770 / ENG-773] Copy/paste / Explorer context menu duplicate

fogodev:
This PR fixes issues 770 and 773


utkubakir:
Just tested this and it seems to work, we need to hide FS actions on Overview though.

HeavenVolkoff:
> Just tested this and everything seems to work. We do need to hide FS actions on Overview though.

I am working on it, also fixing a small bug I found in this PR. Will commit all the changes here as they are related

interface/app/$libraryId/Explorer/ContextMenu.tsx
```diff
@@ -83,36 +76,32 @@ export default (props: PropsWithChildren) => {
...
+						store.locationId &&
+							copyFiles.mutate({
+								source_location_id: store.cutCopyState.sourceLocationId,
+								sources_file_path_ids: [store.cutCopyState.sourcePathId],
+								target_location_id: store.locationId,
+								target_location_relative_directory_path: params.path ?? '/',
+								target_file_name_suffix: store.cutCopyState.sourceParentPath === (params.path ?? '/') ? ' copy' : null
+							});
+					} else {
+						store.locationId && store.cutCopyState.sourceParentPath !== params.path &&
```
HeavenVolkoff, line 94: This check is missing a default value for path (I am fixing it)

HeavenVolkoff (changes requested):

interface/app/$libraryId/Explorer/ContextMenu.tsx
fogodev: Damn good catch, I noted that after writing both checks, but only fixed on the first one

HeavenVolkoff:
So, I added the checks for hiding the FS actions in the Overview and the MediaView. I also added error handling to all the actions in the context menus, so that the user has some feedback when things fail. The actions seem to be working as intended from what I could observe from my tests. However, there are 3 points we should look into (don't know if in this PR or another):

 1. Errors that occur in the paste action are not raised to the front. This happens due to the action running inside a job, which the API route does not wait for (which I think is on purpose). Maybe we should change this in the front to be some kind of subscription, so the user has some feedback if any error occurs when pasting files.

 2. Cutting files does not automatically update the file list in the destination. The file is removed from the origin, but only shows up in the destination after a re-scan

 3. Duplicate / Copying a file to the same folder, creates a suffixed copy of the file, however that only work once, as the suffix is hard-coded and it doesn't take into account other previous copies of the same file already present.

interface/app/$libraryId/Explorer/File/ContextMenu.tsx
```diff
@@ -38,31 +44,39 @@ interface Props {
...
+	const location = useLocation();
+	const objectData = data ? getItemObject(data) : null;
 	const explorerView = useExplorerViewContext();
 	const explorerStore = useExplorerStore();
-	const [params] = useExplorerSearchParams();
-	const objectData = data ? getItemObject(data) : null;
+	const [{ path: currentPath }] = useZodSearchParams();
+	const { cutCopyState, showInspector, ...store } = useExplorerStore();
+
+	const isOverview = location.pathname.endsWith('/overview');
```
utkubakir, line 55: As mentioned on Slack, this should cover tags too. I think we can just do `isLocation` instead of filtering those 2 separately.

utkubakir (changes requested):

fogodev:
I tried pretty hard to reproduce this faulty cut behavior but failed hard lol. Both on my Linux and Mac machines it works as expected. 

Regarding the other points, they're more improvements than problems so we can address them later and we sholdn't postpone our alpha release anymore.

fogodev:
Actually just found out, just happens on Linux if you cut and paste between different locations. It's a watcher bug

jamiepine (approved):

PR #996, opened 2023-06-22
[ENG-786] Hide List view on categories and hide columns view

utkubakir:
- Hide list view on categories bcs it lags
- hide columns view bcs it doesn't exist

### DIFF
--- a/interface/app/$libraryId/overview/index.tsx
+++ b/interface/app/$libraryId/overview/index.tsx
@@ -24,7 +24,7 @@ export const Component = () => {
 	const page = usePageLayout();
 
 	const { explorerViewOptions, explorerControlOptions, explorerToolOptions } =
-		useExplorerTopBarOptions();
+		useExplorerTopBarOptions({ disableViewOptions: ['List view'] });
 
 	const [selectedCategory, setSelectedCategory] = useState<Category>('Recents');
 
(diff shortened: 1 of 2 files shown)


ameer2468 (approved):

interface/app/$libraryId/overview/index.tsx
```diff
@@ -24,7 +24,7 @@ export const Component = () => {
 	const page = usePageLayout();
 
 	const { explorerViewOptions, explorerControlOptions, explorerToolOptions } =
-		useExplorerTopBarOptions();
+		useExplorerTopBarOptions({ disableViewOptions: ['List view'] });
```
Brendonovich, line 27: can we just do the filtering before we use `explorerViewOptions` rather than doing it internally?

Brendonovich (changes requested):

interface/app/$libraryId/overview/index.tsx
utkubakir: I think this is much cleaner. It can be expanded to filter control options and tool options too.

Brendonovich: I think just doing `.filter` in the component is nicer, why should the hook do any filtering

Brendonovich (approved):

PR #1008, opened 2023-06-23
replace github star button in home cta component

YounesBessa:


YounesBessa:
Oh okay I thought that's what Jamie wanted, he mentioned it on stream earlier today

Brendonovich:
Ah, didn't catch that. Will leave it to him to review then

oscartbeaumont:
Requiring an internet connection for a button to load doesn't sit right with me on an offline first desktop application.

utkubakir:
Great PR, thank you for your contribution. Can we make the GitHub button or our waitlist button the same height?

YounesBessa:
Yes, the width and height are adjustable, should I shrink the Github button or make the waitlist button bigger?

utkubakir:
> Yes, the width and height are adjustable, should I shrink the Github button or make the waitlist button bigger?

Let's shrink the Github button, also instead of the middle. Can we put it to the top right and to the footer? Because we will replace the current buttons with download links soon.

jamiepine:
Op sorry I was thinking about putting this in the footer like cal.com do.

YounesBessa:
Okay I just reverted that first commit and put the button in the footer instead

utkubakir:
image

The location of it is great, but it's a big boi that's for sure haha @jamiepine

PR #1017, opened 2023-06-24
Job System - Simplify data

oscartbeaumont:
~~This will merge into `job-system-better-progress` so can be merged once reviewed, although I would recommend reviewing the other PR first so you can see them in isolation. This means it can be merged prior to the alpha going out but it will only enter the `main` branch after the alpha so we don't introduce bugs.~~

Changes:
Previously we had: given type `A` and `B` -> `impl StatefulJob for A where <T as StatefulJob::Init> = B`
Now we have: given type `A` -> `impl StatefulJob for A` (basically merge init data and the job struct itself).

This change is mostly to setup for some bigger things I have planned.


core/src/job/mod.rs
```diff
@@ -126,83 +121,52 @@ pub struct Job<SJob: StatefulJob> {
...
-}
-
-impl<SJob, Init> Job<SJob>
-where
-	SJob: StatefulJob<Init = Init> + 'static,
-	Init: JobInitData<Job = SJob>,
-{
-	fn new(init: Init) -> Box<Self> {
+impl<SJob: StatefulJob> Job<SJob> {
+	pub(super) fn new(init: SJob) -> Box<Self> {
```
fogodev, line 129: This `pub(super)` here can be a bit misleading, the `super` module of `job` module, is the core's `lib.rs` so it can just be `pub` like the `new_with_action` method. As `job` is already a `pub(crate)` module, its visibility already has limited scope.

fogodev (dismissed):

Brendonovich (approved):

PR #1021, opened 2023-06-24
[ENG-803] Set date upon tag creation

### LINKED ISSUE #1020: dates return as `null` for created tags
### Describe the bug

New and seeded tags return a `null` value upon creation.

image

### Reproduction

1. Open Spacedrive
2. Create new tag

### Expected behavior

Create a brand new tag and save to DB with `date_created` and `date_modified` filled with the correct timestamps

### Platform and versions

```Shell
PNPM 8.6.1
cargo 1.70.0 (ec8a8a0ca 2023-04-25)
rustc 1.70.0 (90c541806 2023-05-31)

Operating System: Ubuntu 22.04.2 LTS x86_64
```

### Stack trace

_No response_

### Additional context

This may affect all platforms

gedeondoescode:
This PR intends to solve an issue with `date_created` and `date_modified` returning `NULL` inside main library database. As of now, this is partially functioning, but has a few issues such as the time being stored is in UNIX time and that when attempting to create a new tag, it fails to execute. At the moment this PR exists as a draft as I continue to make changes and improve it.

Closes #1020


interface/app/$libraryId/settings/library/tags/CreateDialog.tsx
```diff
@@ -5,7 +5,9 @@ import { ColorPicker } from '~/components';
 
 const schema = z.object({
 	name: z.string(),
-	color: z.string()
+	color: z.string(),
+	date_created: z.string()
```
utkubakir, line 9: the core should handle this, no need to pass it from frontend

gedeondoescode: I've already attempted this from the start. Because it's generated as a required argument for `data`, it has to be there. If there's a way to omit that from being sent through that'd be great.

utkubakir: hmm @brxken128

brxken128: On it

brxken128:
This is essentially finished, tested, and working great. Nice start btw @gedeondoescode, I hope the additional code helps to clear things up. Feel free to ping me in the Discord (same usename) if you need any future advice!

utkubakir (approved):

core/src/object/tag/seed.rs
```diff
@@ -24,9 +28,23 @@ pub async fn new_library(library: &Library) -> prisma_client_rust::Result<()> {
...
+		.db
+		.tag()
+		.update_many(
+			tags.iter()
+				.map(|x| tag::color::equals(Some(x.color.clone())))
+				.collect(),
+			vec![tag::date_created::set(Some(Utc::now().into()))],
+		)
+		.exec()
+		.await?;
```
Brendonovich, line 47: `date_created` is already set inside `exec`, no need to do it here

core/src/object/tag/seed.rs
```diff
@@ -24,9 +28,23 @@ pub async fn new_library(library: &Library) -> prisma_client_rust::Result<()> {
 		},
 	];
 
-	for tag in tags {
+	for tag in tags.clone() {
```
Brendonovich, line 31: `clone` isn't necessary since `tags` only has to be used in this loop

Brendonovich (changes requested):

Brendonovich (approved):

PR #1027, opened 2023-06-26
Filter search implementation

LuanderFarias:
I modified search.rs to let the future implementation of filters on "find a file"

### DIFF
--- a/core/src/api/search.rs
+++ b/core/src/api/search.rs
@@ -138,9 +138,9 @@ struct FilePathFilterArgs {
 	path: Option<String>,
 	#[specta(optional)]
 	object: Option<ObjectFilterArgs>,
-	#[spectra(optional)]
+	#[specta(optional)]
 	size_range: Option<(u64, u64)>,
-	#[spectra(optional)]
+	#[specta(optional)]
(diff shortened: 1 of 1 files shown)


core/src/api/search.rs
```diff
@@ -120,6 +138,10 @@ struct FilePathFilterArgs {
 	path: Option<String>,
 	#[specta(optional)]
 	object: Option<ObjectFilterArgs>,
+	#[spectra(optional)]
```
jamiepine, line 141: shouldn't it be "specta"?

LuanderFarias: oh yeah, my bad

utkubakir:
Clippy not happy

jamiepine:
Upon a slightly closer look this isn't how we would implement this, you're directly reading a directory on the filesystem, not querying our library database. The range params are correct, but you'd have to upgrade the query in the search.paths and search.objects procedure to include those params. You wouldn't need to create a new struct for a File as we already have data structures for that too.

LuanderFarias:
oh, ok. as i commented in the code i didn't find the path for the database and the nodes lol

PR #1030, opened 2023-06-27
Update recently created files

fogodev:
We discussed on slack that the location watcher wasn't properly updating a file after its initial creation, reporting as an empty file. So in this PR I'm trying to improve how the watcher handles creation and updates of files.


jamiepine:
So this works for a re-index, but still doesn't work via the location watcher. sent screenshot on Slack

jamiepine (approved):

PR #1035, opened 2023-06-27
Fix job resuming/pausing and cancelling

utkubakir:
We don't know why this works, but it does.


jamiepine:
I think its because one worker handles all jobs in a group, so to pause a worker you would need to send the id of the first job in the group.

jamiepine (approved):

fogodev:
It works because this `data.id` here https://github.com/spacedriveapp/spacedrive/blob/main/interface/app/%24libraryId/Layout/Sidebar/JobManager/JobGroup.tsx#L83 is the job group id, not the individual job's id.

PR #1047, opened 2023-06-28
[ENG-846] Preferences

Brendonovich:


core/src/sync/manager.rs
```diff
@@ -303,6 +303,7 @@ impl SyncManager {
 						.await?;
 				}
 			},
+			ModelSyncData::Preference(_, _) => todo!(),
```
brxken128, line 306: 😠

brxken128 (approved):
Looks good!

I saw some `serde`, but could we maybe get `bincode ^2.0.0-rc.3`  support added at some point? Doesn't have to be in this PR, but I tend to use `bincode` quite a lot and it'd definitely help.

Additionally, what's stopping direct usage of `db.preference()`? Just a curiosity, if needs be we can keep an eye out for direct usage within future PR reviews.

core/src/sync/manager.rs
Brendonovich: dw about this i'll fix it in due time haha

Brendonovich:
> I saw some serde, but could we maybe get bincode ^2.0.0-rc.3 support added at some point? Doesn't have to be in this PR, but I tend to use bincode quite a lot and it'd definitely help.

I'm using msgpack for this and would rather it over bincode, am gonna be using msgpack in the sync system eventually too.

> Additionally, what's stopping direct usage of db.preference()? Just a curiosity, if needs be we can keep an eye out for direct usage within future PR reviews.

Type safety of preference keys + DX when creating preferences

oscartbeaumont (approved):

PR #1048, opened 2023-06-28
Indexer rule `No Git`

### LINKED ISSUE #621: Indexer rule `Git Projects` not working as expected
### Describe the bug

Hi,

While testing the UI that I implemented in #620 for informing indexer rules when adding locations, I noticed that the behavior of the `Git Repositories` rule seemed odd to me. I thought it would only hide the `.git` directory of any Git project. However, as far as I could test, it hid all normal files and directories, including the `.git` directory, and only showed the hidden files. Is this the correct behavior for this indexer rule?

Here are some screenshots of the same Location files with different indexer rules enabled:

- Without indexer rule:
photo_2023-03-18_15-57-10

- With `Reject Hidden Files` rule:
photo_2023-03-18_15-57-07

- With `Git Repositories` rule:
photo_2023-03-18_15-57-02

### Reproduction

1. Add a new Location pointing to a Git project directory and enable the `Git Repositories` indexer rule
2 . Only see hiden files and nothing else when opening the location

### Expected behavior

IMHO I think the `Git Repositories` should only hide git related files and directories and index everything else.

### Platform and versions

```Shell
7.29.3
cargo 1.68.0

$> uname -a
```

### Stack trace

```Shell
None
```

### Additional context

None

iammatthi:
Implemented the `No Git` indexer rule to hide all git-related files and directories. 

List of files and directories that this indexer rule hides: `.git`, `.gitignore`, `.gitattributes`, `.gitkeep`, `.gitconfig`, `.gitmodules`.

- Without indexer rule:
image

- With `No Git` rule:
image

Closes #621


jamiepine:
Makes more sense than "Only Git" for sure, but I wonder if just a toggle to show/hide hidden files would be better than making this a system indexer rule. Part of me wants to remove the latter two system indexer rules anyway. 

Also, we need to get previews for text files done next so we can be done with that text icon being everywhere.

Will talk to the team, might approve this in the meantime.

utkubakir (approved):

PR #1051, opened 2023-06-28
[ENG-823] Fixing bad handling of dots in directory name

fogodev:
Found out that directories with dots in its name were being incorrectly handled.


jamiepine (approved):
hallelujah

PR #1053, opened 2023-06-29
[ENG-858] Source analytics paths from router rather than regex replacement

Brendonovich:
all my homies hate regex

### DIFF
--- a/interface/app/index.tsx
+++ b/interface/app/index.tsx
@@ -65,28 +65,43 @@ export const routes = [
 	}
 ] satisfies RouteObject[];
 
+/**
+ * Combines the `path` segments of the current route into a single string.
+ * This is useful for things like analytics, where we want the route path
+ * but not the values used in the route params.
+ */
 const useRawRoutePath = () => {
-	const lastMatchId = useMatches().slice(-1)[0]!.id;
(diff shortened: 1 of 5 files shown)


oscartbeaumont (dismissed):

interface/app/index.tsx
```diff
@@ -60,3 +64,29 @@ export const routes = [
...
+
+					return [`${rawPath}/${item.path}`, item];
+				},
+				['' as string, { children: routes }] as const
+			),
+		[lastMatchId]
+	);
+
+	return rawPath
+}
```
utkubakir, line 92: Is this hook tested on all our screens, does the output matches what we expect? It looks complicated.

utkubakir (changes requested):

interface/app/index.tsx
Brendonovich: Should work on all screens, it just uses some properties that are intrinsic to react router to join the current route segments into one string.

utkubakir: Some documentation sounds nice. Can we also have some comments on `RootContext` to just get an idea of what it's used for and where? We should have this for all our context in my opinion.

utkubakir (approved):

PR #1059, opened 2023-06-30
[ENG-866, ENG-535] Thumbnailer fixes (double-free, thumbnails for portrait videos)

HeavenVolkoff:
- [X] Fix double-free during `MovieDecoder` Drop
 		Happened whenever trying to parse a video file that contained extra metadata, because our code was casting the raw pointers containing the metadata received from ffmpeg with [`CString::from_raw`](https://doc.rust-lang.org/std/ffi/struct.CString.html#method.from_raw). However, this function results in undefined behavior when called with foreign pointers, which here was resulting in a double-free due to the `CString` freeing the pointer when dropped and the `MovieDecoder` also attempting to free the same pointer when dropped by calling `avformat_close_input`
 - [X] Fix broken thumbnails for portrait videos
 		Portrait thumbnails were broken due to some obtuse logic for calculating the thumbnail size. This was solved by removing some broken logic for handling anamorphic videos and simplifying the thumbnail size filter logic
 - [X] Fix some inconsistencies between our MovieDecoder impl and [ffmpegthumbnailer's MovieDecoder impl](https://github.com/dirkvdb/ffmpegthumbnailer/blob/e0bf01d97e0261219b06d1e560e923354949b93e/libffmpegthumbnailer/moviedecoder.cpp)
 - [X] Make sd-desktop-macos use the correct macOS deployment target

Closes: ENG-866, ENG-535


brxken128 (approved):

utkubakir (approved):
Tested and everything lgtm.

PR #1060, opened 2023-06-30
[ENG-869] Fix identifier missing comma

jamiepine:
Convert "Processed 34543 of 89344 paths" into "Processed 34,543 of 89,344 paths".


interface/app/$libraryId/Layout/Sidebar/JobManager/useJobInfo.tsx
```diff
@@ -127,6 +134,21 @@ function plural(count: number, name?: string) {
 }
 
 function comma(x: number) {
-	if (!x) return 0;
+	if (!x) return '0';
 	return x.toString().replace(/\B(?=(\d{3})+(?!\d))/g, ',');
 }
```
HeavenVolkoff, line 139: I think it would be better to use: `new Intl.NumberFormat().format(number)` instead of this comma function. This way the number will be formatted according to each user's locale (ex: commas for English, dots for Portuguese) and it avoids the regexp:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/format#see_also

HeavenVolkoff (dismissed):

interface/app/$libraryId/Layout/Sidebar/JobManager/useJobInfo.tsx
Brendonovich: Is `Intl` available everywhere?

utkubakir: > Is `Intl` available everywhere?

https://caniuse.com/?search=Intl.NumberFormat

interface/app/$libraryId/Layout/Sidebar/JobManager/useJobInfo.tsx
```diff
@@ -127,6 +134,21 @@ function plural(count: number, name?: string) {
...
 
 function comma(x: number) {
-	if (!x) return 0;
+	if (!x) return '0';
 	return x.toString().replace(/\B(?=(\d{3})+(?!\d))/g, ',');
```
Brendonovich, line 138: never compile regexes inline, move it to a const outside the function instead

interface/app/$libraryId/Layout/Sidebar/JobManager/useJobInfo.tsx
HeavenVolkoff: Yeah, it should be available for all our supported platforms

packages/client/src/utils/index.ts
```diff
@@ -34,3 +34,8 @@ export type UnionToTuple<T> = UnionToIntersection<T extends never ? never : (t:
 ) => infer W
 	? [...UnionToTuple<Exclude<T, W>>, W]
 	: [];
+
+export function formatNumber(n: number) {
+	if (!n) return '0';
+	return Intl.NumberFormat().format(n);
+}
```
Brendonovich, line 41: Is this available on React Native? It should be in `interface` if not

utkubakir: Mobile CI passed so I am gonna say yes.

utkubakir (approved):

PR #1065, opened 2023-07-01
[ENG-807] List view improvements

niikeec:


interface/hooks/useIsTextTruncated.ts
```diff
@@ -1,29 +1,26 @@
...
+
+	useCallbackToWatchResize(
+		() => setIsTruncated(determineIsTruncated()),
+		[determineIsTruncated],
+		element
+	);
+
+	useEffect(() => {
+		setIsTruncated(determineIsTruncated());
+	}, [element, determineIsTruncated, text]);
```
Brendonovich, line 23: Ik this was here before you changed it but could the state + callback + effect be combined into a memo?

Brendonovich (changes requested):
all da contexts

interface/hooks/useIsTextTruncated.ts
niikeec: If you prefer this
```ts
export const useIsTextTruncated = (element: RefObject<HTMLElement>, text: string | null) => {
    const { width } = useResizeObserver({ ref: element });

    return useMemo(() => {
        if (!element.current) return false;
        return element.current.scrollWidth > element.current.clientWidth;
    }, [element, width, text]);
};
```

Brendonovich: Ye that should do it

Brendonovich (approved):

PR #1066, opened 2023-07-01
[ENG-765] Reported Total capacity and Free space are wrong

HeavenVolkoff:
WIP (Still needs more testings)


core/src/volume.rs
```diff
@@ -59,102 +62,307 @@ impl From<VolumeError> for rspc::Error {
...
+// 	// cleanup: remove all unmodified volumes associate with this client
+
+// 	Ok(())
+// }
+
 // #[test]
 // fn test_get_volumes() {
 //   let volumes = get_volumes()?;
 //   dbg!(&volumes);
 //   assert!(volumes.len() > 0);
```
brxken128, line 367: dead code?

brxken128 (commented):
Very happy with the changes - I'm just going to test on my Linux machines as they have quite odd storage configurations.

brxken128 (approved):
Tested on both MacOS and Linux (Fedora 38), works perfectly. Great work (as usual)!

core/src/volume.rs
HeavenVolkoff: maybe. IMHO, I don't see any advantage of saving the volume info to the DB, considering we are already saving the Library stats object and querying the system is reasonably inexpensive now that we are exclusively refreshing only the disk info. I think we should drop this and the respective DB tables, but wanted to hear @Brendonovich or @jamiepine take before removing the code...

brxken128: That's a good call. Wanna get this merged and bring this up afterwards, or wait for their call?

utkubakir: Let's open this discussion on Slack and merge this

utkubakir (approved):

core/src/volume.rs
HeavenVolkoff: Will merge this and open a new Draft PR removing the Volume model from DB, so we can discuss there

PR #1069, opened 2023-07-03
Refactor category scroll handlers

Brendonovich:
mega useEffect muahahaha


utkubakir (commented):
> can't get rid of useEffect

hides it into a hook. nice!

utkubakir (approved):

PR #1071, opened 2023-07-04
[ENG-882] Overview page scroll position on category change

niikeec:


interface/app/$libraryId/index.tsx
```diff
@@ -3,7 +3,7 @@ import settingsRoutes from './settings';
 
 // Routes that should be contained within the standard Page layout
 const pageRoutes: RouteObject = {
-	lazy: () => import('./PageLayout'),
+	lazy: () => import('./Layout/Page'),
```
Brendonovich, line 6: I'd prefer if this left as `./PageLayout`.
As stated in the [`@sd/interface` guidelines](https://www.spacedrive.com/docs/developers/prerequisites/guidelines), `Layout` should be reserved for the layout that applies to _all_ routes in a given folder, whereas `PageLayout` only applies to some. Making `PageLayout` a folder and putting everything in there would be good.

Brendonovich (changes requested):

Brendonovich (approved):

PR #1073, opened 2023-07-05
[ENG-883] Fix macOS close button exiting app instead of hiding main window

HeavenVolkoff:


apps/desktop/src-tauri/src/main.rs
```diff
@@ -85,16 +85,16 @@ async fn main() -> tauri::Result<()> {
 	#[cfg(debug_assertions)]
 	let data_dir = data_dir.join("dev");
 
-	let _guard = Node::init_logger(&data_dir);
```
utkubakir, line 88: is this not important?

oscartbeaumont: Good catch! That is our entire logging setup including file-based logging so we probs don't wanna remove it.

HeavenVolkoff: Oh shit, sorry about that, it was an automatic thing, I saw the underscored and that it wasn't being used and removed. Will revert and add a comment. Thanks for catching this @utkubakir

fogodev (approved):

PR #1077, opened 2023-07-06
[ENG-884] Update file_path metadata

fogodev:
Now we properly keep track of possible file changes when Spacedrive is closed, each indexer job or shallow indexing segregates `file_path`s to have its metadata updated on DB.


HeavenVolkoff (approved):
I am still green on how the indexer/watcher works, but the rust code looks good to me. Also did some basic tests (create some locations, copy/cut/delete files, moves some files with the app closed) and they all worked fine

PR #1082, opened 2023-07-07
[ENG-767, ENG-916] Improvements and fixes for `Open`/`Open With`

### LINKED ISSUE #1013: [Linux] “Open with” only works with browsers
### Describe the bug

Open with seams to only work with browsers (Firefox, Chrome, Edge) and not with e.g. Gimp, Inkscape or Okular (the file I tested with was a pdf).

Note that “Open in file manager” also opens Firefox (default web browser) for me instead of Dolphin (default file browser).

Clicking just “Open” also opens Firefox instead of Okular for pdf.

I also checked with different file managers (Okular and Deepin) and it worked as expected.

### Reproduction

1. right-click a PDF
2. “Open with”
3. Select any program which is not a web browser (I tested with Gimp, Inkscape, and Ocular)
4. Doesn't open
5. Repeat, but select a web browser (I tested with Chrome, Firefox, and Edge) and it opens.

### Expected behavior

"Open with" should work

### Platform and versions

```Shell
Arch Linux, using the latest release from the action (697dfff).
```

### Stack trace

_No response_

### Additional context

No error messages logged.

HeavenVolkoff:
- [X] Add logic to normalize PATH, XDG and `GStreamer` plugin environment variables on Linux
- [X] Add some missing `GStreamer` dependencies for Linux systems in `setup-system.sh`
- [X] Fix `AppImage` environment influencing external apps launched through Spacedrive
- [X] Replace custom Desktop file parser and open with code with `Gio.AppInfo` API:
	https://github.com/gtk-rs/gtk-rs-core/tree/master/gio
	https://docs.gtk.org/gio/iface.AppInfo.html
	https://gitlab.gnome.org/GNOME/gtk/-/blob/main/tools/gtk-launch.c
- [X] Replace `opener` with `GLib` API for opening files and directories in Linux (directories only for AppImage, as the opener implementation is better, but crashes when used from within an AppImage)
- [X] Implement `Open With` fallback logic for macOS < 12
- [X] Exports apps icon data on macOS swift native implementation of `get_open_with_applications`
- [X] Fix logic for retrieving system apps that can open multiple files `paths`. Instead of returning the union of all apps that can open any of the given files, return the intersection of apps that can open all the given files.

Closes #1013
Resolves ENG-767, ENG-916


crates/macos/Cargo.toml
```diff
@@ -5,10 +5,8 @@ license = { workspace = true }
...
 edition = { workspace = true }
 
-# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
-
-[dependencies]
+[target.'cfg(target_os = "macos")'.dependencies]
 swift-rs = { workspace = true }
 
-[build-dependencies]
+[target.'cfg(target_os = "macos")'.build-dependencies]
```
Brendonovich, line 11: Seeing as `sd-macos` should be conditionally depended on, is this really necessary?

Brendonovich: Hmm that's a good point

Brendonovich (approved):

PR #1083, opened 2023-07-07
Ou

Isratja45:
Closes #(issue)

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,77 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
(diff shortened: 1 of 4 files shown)


oscartbeaumont:
Can you please stop!

Our pull request templates say "Closes #(issue)" as an issue must be opened prior to a PR. Your changes have all been rejected as that process was not followed correctly.

The related issue should include information about why you are going to make this change which you have also not provided on these PRs so we can discuss if the change aligns with our goals before opening the PR.

Continually opening PRs with non-descript names may result in a ban because it wastes our time.

Also bumping OpenSSL's version is a problematic change because we use a patch to support IOS simulator which I don't think you have accounted for.

PR #1087, opened 2023-07-10
Check winget version and require update if necessary

etahamad:
Added a check for the winget version to ensure it meets the minimum required version. If the winget version is outdated, the script now exits with an appropriate error message.

This is just a commit that addresses this https://github.com/spacedriveapp/spacedrive/blob/5df1d9a091fec792c72fff45ef9acc5afd4a6a2b/.github/scripts/setup-system.ps1#L165

### DIFF
--- a/.github/scripts/setup-system.ps1
+++ b/.github/scripts/setup-system.ps1
@@ -162,7 +162,13 @@ https://learn.microsoft.com/windows/package-manager/winget/
 '@
     }
 
-    # TODO: Check system winget version is greater or equal to v1.4.10052
+    # Check system winget version is greater or equal to v1.4.10052
+    $wingetVersion = [Version]((winget --version)  -replace '.*?(\d+)\.(\d+)\.(\d+).*', '$1.$2.$3')
+    $requiredVersion = [Version]'1.4.10052'
+    if ($wingetVersion.CompareTo($requiredVersion) -lt 0) {
+        $errorMessage = "You need to update your winget to version $requiredVersion or higher."
+        Exit-WithError $errorMessage
+    }
 
     # Check connectivity to GitHub
     $ProgressPreference = 'SilentlyContinue'


.github/scripts/setup-system.ps1
```diff
@@ -162,7 +162,13 @@ https://learn.microsoft.com/windows/package-manager/winget/
 '@
     }
 
-    # TODO: Check system winget version is greater or equal to v1.4.10052
+    # Check system winget version is greater or equal to v1.4.10052
+    $wingetVersion = (winget --version) -replace '(\d+)\.(\d+)\.(\d+).*', '$1.$2.$3'
```
HeavenVolkoff, line 166: This regex doesn't seem to clean the version string completely:
image

Something like this works:
```powershell
$wingetVersion = [Version]((winget --version)  -replace '.*?(\d+)\.(\d+)\.(\d+).*', '$1.$2.$3')
```
image

.github/scripts/setup-system.ps1
```diff
@@ -162,7 +162,13 @@ https://learn.microsoft.com/windows/package-manager/winget/
...
+    $requiredVersion = '1.4.10052'
+    if ($wingetVersion -lt $requiredVersion) {
```
HeavenVolkoff, line 168: I would prefer that this used the [`Version` ](https://learn.microsoft.com/en-us/dotnet/api/system.version?view=net-5.0) class for the comparison:

```powershell
$requiredVersion = [Version]'1.4.10052'
if ($wingetVersion.CompareTo($requiredVersion) -lt 0) {
```

HeavenVolkoff (changes requested):
Hey, thank you for your contribution. Just need some changes to approve it

etahamad:
Soft reset and forced push the commit, kidnly check.

HeavenVolkoff (approved):

PR #1092, opened 2023-07-12
[ENG-862, ENG-921] Ephemeral locations

fogodev:
Currently, a first draft on how non-indexed locations should work. ~Need some frontend guys support to approve the provided Core API for listing files, directories and locations on a given path that isn't an indexed location.~

Edit (Vítor): Front-end is done, still need to implement `open`/`open with`/`reveal` for non indexed files/directories, but that requires refactoring our backend logic for those functionalities as they are intrinsically dependent on library/location/file ID, which we don't have for ephemeral locations (maybe leave for a future PR?).


HeavenVolkoff (changes requested):
- [ ] Fix indexed Locations sometimes selecting other folder/files in Ephemeral location besides it, due to conflicting IDs

HeavenVolkoff (changes requested):
- [x] Fix Empty context menu sometimes (Ephemeral and indexed locations)
- [x] Fix loading an ephemeral path containing a file without read permissions breaks UI
- [x] ListView flashes EmptyNotice when changing file sorting

apps/desktop/crates/macos/src-swift/env.swift
```diff
@@ -0,0 +1,7 @@
+import Foundation
+import SwiftRs
+
+@_cdecl("get_user_home_directory")
+func getUserHomeDirectory() -> SRString {
+    return SRString(FileManager.default.homeDirectoryForCurrentUser.path)
+}
```
Brendonovich, line 7: We can use Tauri's JS API to do this: https://tauri.app/v1/api/js/path/#homedir

HeavenVolkoff: Yeah, but I am not a fan of the lib tauri is using for this:
https://github.com/tauri-apps/tauri/blob/dev/core/tauri/src/path/desktop.rs#L123C1-L132
https://github.com/xdg-rs/dirs/blob/master/dirs-sys/src/lib.rs#L28-L61

And while `$HOME` is defined in macOS, I think getting the path from the native swift api to be a bit better, and in *nix there is no fallback for missing HOME envvar (not common, but can happen), which I implemented in our code. (For windows the api we are using is the same that tauri is using, but it was so little code that I just copied to our windows crate)

interface/app/$libraryId/Explorer/ViewContext.ts
```diff
@@ -1,6 +1,6 @@
-import { ReactNode, RefObject, createContext, useContext } from 'react';
+import { type ReactNode, type RefObject, createContext, useContext } from 'react';
 
-export type ExplorerViewSelection = number | Set<number>;
+export type ExplorerViewSelection = string | string[];
```
niikeec, line 3: Can be removed

interface/app/$libraryId/Explorer/Inspector/index.tsx
```diff
@@ -104,7 +111,7 @@ const Thumbnails = ({ items }: { items: ExplorerItem[] }) => {
 					pauseVideo={!!explorerStore.quickViewObject || thumbs.length > 1}
 					frame={thumbs.length > 1}
 					childClassName={(type) =>
-						type !== 'icon' && thumbs.length > 1
+						type !== ThumbType.Icon && thumbs.length > 1
```
niikeec, line 114: imo it should be kept as a string

interface/app/$libraryId/Explorer/Inspector/index.tsx
```diff
@@ -170,48 +187,41 @@ const SingleItemMetadata = ({ item }: { item: ExplorerItem }) => {
...
+					<MetaData
+						icon={Path}
+						label="Path"
+						value={fileFullPath}
+						onClick={() => {
+							// TODO: Add toast notification
+							fileFullPath && navigator.clipboard.writeText(fileFullPath);
+						}}
+					/>
+				)}
```
niikeec, line 216: this should be always rendered imo

interface/app/$libraryId/Explorer/Inspector/index.tsx
```diff
@@ -383,14 +364,26 @@ const MultiItemMetadata = ({ items }: { items: ExplorerItem[] }) => {
...
+				{metadata.indexed && (
+					<MetaData icon={Barcode} label="Indexed" value={formatDate(metadata.indexed)} />
+				)}
+				{metadata.accessed && (
+					<MetaData
+						icon={FolderOpen}
+						label="Accessed"
+						value={formatDate(metadata.accessed)}
+					/>
+				)}
```
niikeec, line 386: this should be always rendered imo

interface/app/$libraryId/Explorer/View/ListView.tsx
```diff
@@ -117,21 +117,22 @@ export default () => {
...
+			? explorerView.padding
+			: explorerView.padding?.x) || 16;
 
 	const paddingY =
-		(typeof explorerView.padding === 'object'
-			? explorerView.padding.y
-			: explorerView.padding) || 12;
+		(typeof explorerView.padding === 'number'
+			? explorerView.padding
+			: explorerView.padding?.y) || 12;
```
niikeec, line 127: Don't think this is necessary

niikeec (dismissed):

apps/desktop/crates/macos/src-swift/env.swift
brxken128: In all fairness I've never came across a *nix system configuration that had no `$HOME` var, even if it was set to `/var/empty` or `/nonexistent` (TrueNAS uses the latter).

Brendonovich: I reckon if it's good enough for Tauri it's good enough for us, unless you reeeealy think it's worth it

interface/app/$libraryId/Explorer/Inspector/index.tsx
HeavenVolkoff: Should we leave these fields showing up without a value? Because items from ephemeral locations will never have this info due to not be indexed

interface/app/$libraryId/Explorer/Inspector/index.tsx
HeavenVolkoff: Why? This seems like such a by the book use case for enum. Using strings here just loose on type safety and code clarity without any meaninful gain from what I can see, unless I am missing something here

interface/app/$libraryId/Explorer/Inspector/index.tsx
HeavenVolkoff: same as above

interface/app/$libraryId/Explorer/Inspector/index.tsx
niikeec: Can we make it so they only don't show in ephemeral locations?

interface/app/$libraryId/Explorer/Inspector/index.tsx
niikeec: Not a big fan of enums, but I guess we can keep it.

apps/desktop/crates/macos/src-swift/env.swift
HeavenVolkoff: Ok, I changed to use the tauri API, the Linux $HOME is being solved anyway due to an environment normalization step I introduced some time ago:
https://github.com/spacedriveapp/spacedrive/blob/7d2136627cb1c8ac3e5fd38a8fd121b1f02aa773/apps/desktop/crates/linux/src/env.rs#L120-L178

interface/app/$libraryId/Explorer/FilePath/Thumb.tsx
```diff
@@ -22,13 +22,11 @@ import { getExplorerStore } from '../store';
...
-	ORIGINAL: 'original',
-	THUMBNAIL: 'thumbnail'
-} as const;
-
-type ThumbType = (typeof THUMB_TYPE)[keyof typeof THUMB_TYPE];
+export enum ThumbType {
+	Icon,
+	Original,
+	Thumbnail
+}
```
niikeec, line 29: Can we change back to const?

HeavenVolkoff: enums are read-only by default, or you mean re-writing this into a const object?

niikeec: back to const object as it was before

utkubakir (approved):

PR #1095, opened 2023-07-12
[ENG-906] Initial library sync

oscartbeaumont:


apps/mobile/package.json
```diff
@@ -32,6 +32,7 @@
 		"@tanstack/react-query": "^4.29.1",
 		"class-variance-authority": "^0.5.3",
 		"dayjs": "^1.11.8",
+		"event-target-polyfill": "^0.0.3",
```
utkubakir, line 35: can we remove this if it's not used?

Brendonovich (approved):

PR #1099, opened 2023-07-14
[ENG-288, ENG-601] Media data (image) extraction, thumbnail orientation fix

brxken128:
This pull request extracts EXIF data from images, and stores them in the `media_data` DB table.

Some of these columns will be reserved for videos only (e.g. `fps`. `duration`). Not all images have EXIF data, and not all images have all fields (in fact, most don't).

This also fixes mismatched thumbnail orientation by reading/parsing the exif data, and modifying the `DynamicImage` accordingly.


core/src/object/file_identifier/mod.rs
```diff
@@ -226,36 +232,83 @@ async fn identifier_job_step(
...
+					.iter()
+					.filter(|x| fp.extension.clone().unwrap_or_default() == x.to_string());
+
+				let create_media_data_items = create_media_data_items
+					.filter_map(|_| {
+						image::get_data_for_image(location_path.join(&path))
+							.ok()
+							.map(|data| {
+								media_data::create_unchecked(vec![
+									media_data::date_created::set(Some(Utc::now().into())),
```
utkubakir, line 293: is this correct? Why is it `now()`?

utkubakir: And don't we have this field on Object? What's the difference between this and that one?

brxken128: Because that's the date the _field_ was created, not the date from the image, in order to keep consistency with the rest of the schema. The line directly beneath defines when the image was taken.

brxken128: And the difference is that media data won't always be extracted at the same time an object is created.

Test it out - index a location, find an object and copy the `date_created` field, and then search for it in the `media_data` table - it won't be there as the fields are created at different times.

brxken128:
Reverting to draft again as it turns out there's a LOT more we can improve upon to ensure coverage for the most amount of images.

core/src/object/file_identifier/mod.rs
utkubakir: I reckon just remove it or rename it to `date_extracted` then. "// filesystem creation time" is also not really a comment I understand to tbh

Also about schema consistency, it's fine to have a modal without some fields, no need to worry about it.

jamiepine: Swaying side to side with my opinion for a minute I think `date_extracted` makes a lot of sense, I'm in favour of communicating the purpose of the field accurately over consistency of having `date_created` exist on every model. Reality is not all models have timestamps so it would never be consistent anyway.

brxken128:
It worked, until I merged main and @Brendonovich's sync stuff completely broke it. I'm unsure where to go from here.

core/src/object/file_identifier/mod.rs
brxken128: I settled for an enum, that is either: the local time from the image, UTC (via `chrono::DateTime<FixedOffset>`) or undefined if it was undefined within the image.

brxken128:
Converting to a draft once again as we need to integrate the `MediaDataImage` struct (and the nested structs) within the frontend, but I'm not too sure how.

core/prisma/schema.prisma
```diff
@@ -293,18 +293,21 @@ model Object {
...
+    copyright    Bytes?
+    artist       Bytes?
+
+    // video-specific
+    duration Int?
+    fps      Int?
+    streams  Int?
+    codecs   String? // eg: "h264,acc"
 
     object Object? @relation(fields: [id], references: [id], onDelete: Cascade)
```
Brendonovich, line 312: @jamiepine Just wanted to confirm that we do want to store media data on objects and not file paths? Seeing as multiple video/image files with different media data could be stored under the same object, I'm not sure this will work like we want it to.

crates/media-data/src/flash/values.rs
```diff
@@ -0,0 +1,139 @@
...
+			0x5d => Self::AutoFiredRedEyeReductionNoReturn,
+			0x5f => Self::AutoFiredRedEyeReductionReturn,
+			_ => return Err(Error::Conversion),
+		};
+
+		Ok(res)
+	}
+}
+
+impl Display for FlashValue {
```
Brendonovich, line 85: What's the purpose of this impl?

core/prisma/schema.prisma
jamiepine: Thinking about this, it would be correct to do this unless we had a mechanism to handle cases where an object has varying `cas_ids`, it shouldn't happen often and when it does we could handle it. For now, lets leave it on the Object for the sake of saving engineering hours and preventing data duplication. We'll come back to this and create a system that solves the problem while also avoiding data duplication.

crates/media-data/src/flash/values.rs
brxken128: So we can explain in-depth how the flash fired?

Brendonovich (changes requested):
Seems like migrations are still a bit messed up. I'll go in and fix them.

core/prisma/schema.prisma
```diff
@@ -293,20 +293,25 @@ model Object {
...
+    id Int @id @default(autoincrement())
+
+    dimensions     Bytes
+    media_date     Bytes
+    media_location Bytes?
+    camera_data    Bytes
+    artist         Bytes?
+    description    Bytes?
+    copyright      Bytes?
+    exif_version   Bytes?
```
Brendonovich, line 305: These all need to be optional

Brendonovich (dismissed):

utkubakir:
let's fix types and merge this pls

core/src/api/files.rs
```diff
@@ -43,11 +47,49 @@ pub(crate) fn mount() -> AlphaRouter<Ctx> {
...
+			#[derive(Type, Deserialize)]
+			pub enum MediaDataType {
+				Image,
+				Video,
+				Audio,
+			}
+
+			#[derive(Type, Deserialize)]
+			pub struct GetMediaDataArgs {
+				pub id: i32,
```
Brendonovich, line 65: use `object::id::Type`

core/src/api/files.rs
```diff
@@ -43,11 +47,49 @@ pub(crate) fn mount() -> AlphaRouter<Ctx> {
...
 						.await?)
 				})
 		})
+		.procedure("getMediaData", {
+			#[derive(Type, Deserialize)]
+			pub enum MediaDataType {
+				Image,
+				Video,
+				Audio,
+			}
```
Brendonovich, line 61: This is redundant - you already fetch `object`, so the type can be inferred from it.

core/src/api/files.rs
```diff
@@ -43,11 +47,49 @@ pub(crate) fn mount() -> AlphaRouter<Ctx> {
...
+				pub md_type: MediaDataType,
+			}
+
+			R.with2(library())
+				.query(|(_, library), args: GetMediaDataArgs| async move {
+					let i = library
+						.db
+						.object()
+						.find_unique(object::id::equals(args.id))
+						.include(object::include!({ media_data }))
```
Brendonovich, line 75: `select` would be nicer here
```suggestion
						.select(object::select!({ id kind media_data }))
```

Brendonovich (changes requested):

core/src/api/files.rs
brxken128: oh good catch, bit redundant selecting media data when we're querying media data

core/src/api/files.rs
brxken128: i was on the fence about this, will remove it as yeah it's pretty pointless

brxken128:
This should be ready to go, there's no/minimal UI work but that can come in another PR as this one is quite large already.

Brendonovich (approved):

PR #1121, opened 2023-07-18
Upgrade to rspc-rc.1

oscartbeaumont:
Well, it was v1.0.0-rc.1 when I started. Now it's now v1.0.0-rc.4.


utkubakir:
Mostly working on mobile/desktop, I couldn't find any issues except the already mentioned tag deletion issue

apps/mobile/modules/sd-core/ios/SDCore.podspec
```diff
@@ -0,0 +1,40 @@
+#
```
captain-hoity-toity, line 1: Overall, the code looks good. However, there are a few areas that could be improved:

1. License Identifier:
```diff
- s.license = 'APGL-3.0'
+ s.license = 'AGPL-3.0'
```
The correct identifier for the GNU Affero General Public License v3.0 is `AGPL-3.0`, not `APGL-3.0`.

2. Source Repository:
```diff
- s.source = { git: 'https://github.com/spacedriveapp/spacedrive' }
+ s.source = { git: 'https://github.com/spacedriveapp/spacedrive', tag: s.version.to_s }
```
It's recommended to specify the tag corresponding to the pod version in the source attribute. This ensures that the correct version of the source code is used when someone installs the pod.

3. Module Map:
```diff
- s.module_map = "#{s.name}.modulemap"
+ s.module_map = 'SDCore.modulemap'
```
The module map should be a constant string. Using string interpolation with `s.name` could lead to issues if the pod name is changed in the future.

brxken128: > Overall, the code looks good. However, there are a few areas that could be improved:
> 
> 1. License Identifier:
> ```diff
> - s.license = 'APGL-3.0'
> + s.license = 'AGPL-3.0'
> ```
> The correct identifier for the GNU Affero General Public License v3.0 is `AGPL-3.0`, not `APGL-3.0`.

We use workspace inheritance for this everywhere, although I'm not sure if it extends to a crate this nested.

PR #1123, opened 2023-07-19
[ENG-778] Location clean up job

fogodev:
This PR introduces a new background actor that can be invoked by hand and will also run from time to time, cleaning out stale thumbnails from file system. It also tweaks the Orphan Remover actor to run automatically from time to time.

### DIFF
--- a/core/src/library/library.rs
+++ b/core/src/library/library.rs
@@ -8,7 +8,10 @@ use crate::{
 		LocationManager,
 	},
 	node::NodeConfigManager,
-	object::{orphan_remover::OrphanRemoverActor, preview::get_thumbnail_path, thumbnail_remover::ThumbnailRemoverActor},
+	object::{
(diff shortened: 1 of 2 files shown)


core/src/location/mod.rs
```diff
@@ -706,11 +706,10 @@ pub async fn delete_directory(
 		})],
 	);
 
-	for params in children_params.chunks(512) {
-		db.file_path().delete_many(params.to_vec()).exec().await?;
-	}
```
Brendonovich, line 711: don't we have to chunk here? we could run into the 'Too may params' issue if not.

Brendonovich (changes requested):

core/src/location/mod.rs
fogodev: This params vec always have at most 2 elements, that chunking was doing nothing at all lol. The first one being the location id and the second being an optional parent directory to delete all of its descendents

PR #1127, opened 2023-07-21
[ENG-926] Prevent thumbnail destruction and fix the remover

brxken128:
The thumbnail remover has a bug that results in all thumbnails being deleted, even freshly generated ones.

It now indexes and collects the directory contents before even thinking of removing a file. It then does a simple comparison, and the file is deleted (or not).

The `file_path` full query may be too much in the future, but I do believe it's fine for now.


fogodev:
The root problem is that we put all our thumbnails in the same `thumbnails` directory, just doing some sharding using the `cas_id`. My previous implementation and your new implementation suffers from this same problem, that an `Actor` running for a `Library` will nuke thumbnails generated in another `Library`. To solve this we have to separated thumbnails by `Library`. I will implement `Library` thumbnails separation here.

fogodev (changes requested):

Brendonovich (approved):

PR #1133, opened 2023-07-27
Networked sync

oscartbeaumont:
Among other minor changes, this PR introduces the process by which sync events will be requested & sent over the network. It will probably be modified and cleaned up in the future, but this works for now.


Brendonovich (approved):
it workey 👍

PR #1139, opened 2023-07-28
[ENG-848] Implemented "Copy as Path" functionality

### LINKED ISSUE #1119: [ENG-848] Add copy as path to context menu
Let's temporarily put this to 

```
More actions >
```

 for now in the future we should probably show this when alt is pressed for example, idk.

<sub>From [SyncLinear.com](https://synclinear.com) | [ENG-848](https://linear.app/spacedriveapp/issue/ENG-848/add-copy-as-path-to-context-menu)</sub>

Raghav-45:
I addressed issue #1119 by successfully implementing the "Copy as Path" functionality within the "More Actions" section of the Context Menu. This feature enhances user experience and offers a convenient way to copy file paths. While considering the valuable suggestion of potentially associating this option with the "Alt" key in the future, I opted to make it accessible through the "More Actions" section for now. All tests have passed smoothly

Closes #1119


utkubakir:
This looks good @Raghav-45 , thank you for your contribution! I think it would be great to have this functionality on different screens like Overview (under categories), Search and tag's explorer too.

Raghav-45 (commented):
Hello @utkubakir ,

I'm excited to share that I've successfully extended the "Copy as Path" functionality to the Overview (under categories), Search, and tags explorer sections. Now users can conveniently copy file paths from these different screens, enhancing their overall experience with the application.

I've thoroughly tested the implementation, and everything appears to be working seamlessly. The changes are ready for review and integration into the project.

Thank you for your valuable suggestion, which inspired me to make these enhancements. I'm proud to be part of this project and look forward to contributing further.

Best regards,
Aditya

niikeec:
Hey @Raghav-45, thank you for your contribution. Could we change it so the location path is returned by a [rspc](https://www.rspc.dev/) library mutation using mutateAsync? Could be placed in [locations.rs](https://github.com/spacedriveapp/spacedrive/blob/main/core/src/api/locations.rs).

Raghav-45 (commented):
Hello @niikeec ,

Thank you for your valuable suggestion! I completely agree with you. The approach you suggested to return the location path using the RSPC library mutation with mutateAsync is better and efficient. but while considering your input, I explored an alternative approach for the implementation. Instead of using mutateAsync, I utilized the RSPC library by creating a Query called `locationIdToPath` in RSPC specifically to retrieve only the location path from its location ID. This way, we can avoid unnecessary information retrieval and streamline the process. Moreover, we already have access to other details like file extensions, File Name, Relative File Path with respect to location through `const filePath = data.item` in our contextMenu.

I've made the necessary changes to implement this approach.

niikeec (dismissed):

Raghav-45 (commented):
hey @niikeec ,

I have made the requested changes to the file, and it now utilizes the library's mutateAsync function to obtain the filePath. i also formatted it with prettier & now it copies Absolute File Path when user clicks on `Copy as path` from context menu

But I also think that we can use useLibraryQuery(['files.locationIdToPath', { location_id: filePath?.location_id || -1 }]) only when the user clicks on `Copy as path` from the context menu. By adding this to the onClickHandler, data will only be fetched when the user intends to copy the file path.

Currently i used the `mutateAsync` approach which you requested. but feel free to let me know, I can implement this `useLibraryQuery` approach as well.

HeavenVolkoff:
Hey @Raghav-45, thanks for the contribution. I improved your original implementation by resolving the file's full path in the backend instead of the front-end, as I believe this is a more future-proof solution. Also, I agree that using a query is the correct approach here, but @niikeec's point about not running it in every component render is valid. Therefore, I exposed the rspc internal client to allow us to call queries without having to rely on hooks for this use case. I think this is good to be merged. I'll approve it, but I won't put it in the merge queue yet, so @niikeec can take another look.

HeavenVolkoff (approved):

Raghav-45 (commented):
Hey @HeavenVolkoff , thanks for the contribution.

I completely agree with your points about using query instead of mutation, I also got point of @niikeec , what he was saying not to fetch file path on every render. I acknowledge my error in comprehending the point.

with that being said I did some following improvements
- I did some formatting with prettier and used spacedrive prettier config file it formats a little differently ( i think )
- i also removed unused imports.

Now I think this is good to be merged. you can check and approve it, and @niikeec can take another look.

niikeec:
@HeavenVolkoff Good eye. LGTM

niikeec (approved):

PR #1158, opened 2023-08-03
[ENG-937] Work-around running out of space in GitHub CI for cache factory

HeavenVolkoff:
Attempt to workaround Linux and Windows CI runners out-of-space problems


utkubakir:
Maybe for mobile too?

utkubakir (approved):

utkubakir:
> > Maybe for mobile too?
> 
> Mobile CI is having problem with disk space too? I did not find much discussion around clearing disk space in the macOS version of Github CI runner though, but maybe we could try deleting some unused Xcode SDKs if there are any...

Mobile is alright, I was just curious if it was possible to do it there too. Oh well.

LGTM!

PR #1160, opened 2023-08-03
Library "state"

oscartbeaumont:
This doesn't actually solve library state but is getting very big so gonna get merged and come back to finish off and cleanup.


core/src/p2p/pairing/mod.rs
```diff
@@ -77,195 +79,7 @@ impl PairingManager {
...
-				}
-				PairingResponse::Rejected => {
-					info!("Pairing '{pairing_id}' rejected by remote");
-					self.emit_progress(pairing_id, PairingStatus::PairingRejected);
-				}
-			}
-		});
-
-		pairing_id
+		todo!();
```
Brendonovich, line 82: ?

Brendonovich (approved):

PR #1179, opened 2023-08-05
Fix remove from recents

Brendonovich:
Introduces an `extra` render prop for customising the context menu, and injects the RemoveFromRecents option via it.


niikeec (changes requested):
LGTM besides the type error

niikeec (approved):

niikeec (approved):

PR #1183, opened 2023-08-07
lock libheif-sys version

Brendonovich:


brxken128 (approved):
this is because ubuntu uses 1.14.2 right?

PR #1203, opened 2023-08-11
[ENG-968] Cache job progress events

Brendonovich:
Introduces a `JobManagerContextProvider` that caches job progress events above the job manager, and moves all progress event related logic into `useProgress`. It's kinda unwieldy but it does its job pretty well.
Also, jobs no longer track their own progress. That's up to `JobGroup` only now. Doing otherwise is annoying.


jamiepine (approved):
Works, great job

PR #1206, opened 2023-08-12
Move onboarding form handling logic to Layout

Brendonovich:
Currently, each onboarding page contains its own `useZodForm`, inserts data into `onboardingStore`, and then the `/onboarding/creating-library` route executes the `library.create` mutation in a `useEffect` (eew yucky).
Now, the onboarding `Layout` holds a singular form state that is provided via context to all the sub-routes. The sub-routes then consume this form in their own `Form` components, but none except the final `/onboarding/privacy` route actually call the main `onSubmit` handler, which when called is in charge of navigating to `/onboarding/creating-libraries` (rather than the other way around!).

### DIFF
--- a/interface/app/onboarding/Layout.tsx
+++ b/interface/app/onboarding/Layout.tsx
@@ -14,6 +14,7 @@ import { useOperatingSystem } from '~/hooks/useOperatingSystem';
 import DebugPopover from '../$libraryId/Layout/Sidebar/DebugPopover';
 import { macOnly } from '../$libraryId/Layout/Sidebar/helpers';
 import Progress from './Progress';
+import { OnboardingContext, useContextValue } from './context';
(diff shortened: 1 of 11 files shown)


interface/app/onboarding/Layout.tsx
```diff
@@ -36,41 +37,47 @@ export const Component = () => {
 			// This is neat because restores the last active screen, but only if it is not the starting screen
 			// Ignoring if people navigate back to the start if progress has been made
 			if (obStore.unlockedScreens.length > 1 && !library) {
+				console.log(obStore.unlockedScreens);
```
ameer2468, line 40: You forgot a console log here oops - but looks good!

utkubakir (approved):

PR #1210, opened 2023-08-12
[ENG-735] Tauri 1.4.0

brxken128:
Since the prior PR mentioned in the issue on Linear has been merged into `tauri-cli`, we are able to freely update Tauri version without issue. I've tested this on Aarch64 (`aarch64-apple-darwin`) and ran into no issues.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -156,9 +156,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "1.0.2"
+version = "1.0.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "43f6cb1bf222025340178f382c426f13757b2960e89779dfcb319c32542a5a41"
+checksum = "86b8f9420f797f2d9e935edf629310eb938a0d839f984e25327f3c7eed22300c"
 dependencies = [
  "memchr",
 ]
@@ -234,9 +234,9 @@ dependencies = [
 
 [[package]]
 name = "anstyle-wincon"
-version = "1.0.1"
+version = "1.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "180abfa45703aebe0093f79badacc01b8fd4ea2e35118747e5811127f926e188"
+checksum = "c677ab05e09154296dd37acecd46420c17b9713e8366facafa8fc0885167cf4c"
 dependencies = [
  "anstyle",
  "windows-sys 0.48.0",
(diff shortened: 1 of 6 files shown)


utkubakir:
no mergerino without my AND @HeavenVolkoff review. 

@HeavenVolkoff can you clean up all the patched stuff if this upgrade no more needs it?

brxken128:
> no mergerino without mine AND @HeavenVolkoff review pls
> 
> @HeavenVolkoff can you clean up all the patched stuff if this upgrade no more needs it?

As far as I can tell, we use a [custom wrapper of the Tauri CLI](https://github.com/spacedriveapp/spacedrive/blob/main/apps/desktop/package.json#L9-L10) anyway in order to setup our `Frameworks` ([apps/desktop/src-tauri/scripts/tauri.js](https://github.com/spacedriveapp/spacedrive/blob/main/apps/desktop/src-tauri/scripts/tauri.js)) so it'll need to remain this way anyway. [However, it still makes use of the `tauri-cli` binary](https://github.com/spacedriveapp/spacedrive/blob/a253431d0ff933157dcd0f9b6145fbb4adf974e2/apps/desktop/src-tauri/scripts/tauri.js#L66), which was upgraded to `v1.4.1`.

The fact that our wrapper still works even with Tauri `v1.4.1` is very lucky, although now that the aforementioned PR was merged (the one mentioned in Linear), I think @HeavenVolkoff could probably make some alterations and optimisations? I'm not too sure there.

HeavenVolkoff:
I was holding off on doing this, because while the PR with the issue we were having got merged, it still hasn't been released, it was postponed to tauri 1.5. @brxken128 while you updated everything, the macOS `setup-system.sh` downloads and replaces the installed tauri-ci with our patched one from 1.3 and that is probably why you didn't run in the bundle issue while testing this update:
https://github.com/spacedriveapp/spacedrive/blob/a253431d0ff933157dcd0f9b6145fbb4adf974e2/.github/scripts/setup-system.sh#L349-L375

If there is no pressing need for us to update tauri rn, I would prefer that we wait until 1.5 drops, so we can remove all the workaround code alongside the update.

brxken128:
> I was holding off on doing this, because while the PR with the issue we were having got merged, it still hasn't been released, it was postponed to tauri 1.5. @brxken128 while you updated everything, the macOS `setup-system.sh` downloads and replaces the installed tauri-ci with our patched one from 1.3 and that is probably why you didn't run in the bundle issue while testing this update:
> 
> https://github.com/spacedriveapp/spacedrive/blob/a253431d0ff933157dcd0f9b6145fbb4adf974e2/.github/scripts/setup-system.sh#L349-L375
> 
> If there is no pressing need for us to update tauri rn, I would prefer that we wait until 1.5 drops, so we can remove all the workaround code alongside the update.

If you'd prefer to wait, we'll wait. I think @Brendonovich mentioned something about this fixing part of the updater issue, so that may be something to consider, but I'll close this for now until we know more.

PR #1216, opened 2023-08-14
[ENG-974] DB Backup prototype

oscartbeaumont:


core/src/api/backups.rs
```diff
@@ -0,0 +1,370 @@
...
+			library_name: {
+				let len = u32::from_le_bytes(
+					buf[54..58]
+						.try_into()
+						.map_err(|_| BackupError::MalformedHeader)?,
+				);
+
+				let mut name = vec![0; len as usize];
+				file.read_exact(&mut name[..])?;
+				String::from_utf8(name).map_err(|_| BackupError::MalformedHeader)?
```
brxken128, line 345: We should probably version this or set it in stone somewhere, and have a test vector along with it to ensure any changes don't break previous backups.

brxken128: I know we have the test below, but that only tests the current header state, and not any prior ones/changes - if we made a backup, then altered the header format, the test would still pass but the backup would be malformed due to a lack of versioning and such. I don't mind handling that if needs be, I've done quite a lot regarding headers/versions/etc.

oscartbeaumont: Yeah, the test isn't for versioning just to check both sides match cause messy them up is super easy (as I have found on P2P).

It added a `1` to the header (`sdbkp1`) for this exact purpose. We can bump it to`2` and make a breaking change, then match on it when decoding. I think that should be fine. In the unlikely event we need more than 9 versions we can bump one of the chars or just require `0` match on it and then allow a double-digit version number to preceded it.

I have also moved this behind a feature flag so for now it doesn't *need* to be set in stone. Really we probably should be doing an SQLite snapshot instead of just copying the DB file as copying could potentially result in inconsistencies. This was just a one-day hack around for me to see how far I could get, it's not perfect but it's something for us to itterate on.

core/src/api/backups.rs
```diff
@@ -0,0 +1,370 @@
...
+			file.write_all(&(bytes.len() as u32).to_le_bytes())?;
+			file.write_all(bytes)?;
+		}
+
+		Ok(())
+	}
+
+	fn read(file: &mut impl Read) -> Result<Self, BackupError> {
+		let mut buf = vec![0u8; 6 + 16 + 16 + 16 + 4];
+		file.read_exact(&mut buf[..])?;
```
brxken128, line 314: This seems redundant, `file.read_exact(&mut buf)` will fill the buffer, and leave the cursor position at the start of the compressed header bytes ready to be decompressed.

core/src/api/backups.rs
```diff
@@ -0,0 +1,370 @@
...
+	Header {
+		id,
+		timestamp,
+		library_id: library.id.clone(),
+		library_name: library.config.name.to_string(),
+	}
+	.write(&mut bkp_file)?;
+
+	// Regular tar.gz encoded data
+	let mut tar = tar::Builder::new(GzEncoder::new(bkp_file, Compression::default()));
```
brxken128, line 192: I'm on the fence about using compression here. `flate2` is an awesome library, and gzip is great, but the chance of corruption is higher as if any single byte corrupts, the whole block does and you've just lost your backup.

If it's stored plainly in a tar container, if a byte corrupts, you might end up with a malformed GPS location on one of your images, but at least you still have your data.

Honestly, it's fine and compression here is probably the best option, I'm just trying to cover all bases. We can eventually implement backup validation, and a job that verifies that all backups are in good health via a BLAKE3 checksum (created at the same time as the backup and possibly even stored in the secure keystore, so we know it's valid and hasn't been tampered with).

An adversary tampering a backup before the user restores it is also something we need to take into account, but that's definitely a task for later on.

brxken128 (commented):
I'm happy with the Rust and the header format, it's pretty good for a prototype! I can't comment much on the TS so I'll let someone with more experience handle that, but you have my approval 🙏🏻

core/src/api/backups.rs
utkubakir: Can we have some comments on this topic in the code?

core/src/api/backups.rs
oscartbeaumont: Yeah, deref does make that redundant, didn't notice that.

core/src/api/backups.rs
oscartbeaumont: Yeah, you do raise a fair point. We would probably wanna support both compressed and non-compressed backups and let the user make that tradeoff but for now, I think it's probably fine.

I wonder if file formats exists that are partially autohealing to this? Basically like RAID but within blocks of a file where the checksum and data are stored in different parts of the file. Idk if that's smart but could be investigated for the really concerned user.

Another thing we should think about that I just thought of is that SQLite never frees storage, it keeps it as empty pages which right now are included in backups. We would wanna vacuum the DB to reduce the size of the backup although in practice, moving to SQLite snapshot might help with this (idk much about then, tbh) because always vacuuming the prod DB will decrease performance. These pages can also include deleted data so that is also a concern with them not being properly removed.

I have documented all of these points in ENG-975 (which should probs be split into sub-issues to actually be worked on) which is intended to be the follow-up to this PR at some point to improve a lot of it's shortcomings. Given this feature is behind a feature flag none of these should need immediate fixing just documentation, imo.

Then it's up to @jamiepine and @utkubakir to set the priorities on whether work should continue on this or it be put on the backburner.

core/src/api/backups.rs
oscartbeaumont: Yeah done

utkubakir (approved):
I love this! We should put some effort into the UI too and make this a full feature whenever we can.

I don't think we need to worry much about the zipping failures yet, we can deal with them if we ever get them.

PR #1218, opened 2023-08-14
[ENG-620] Update QuickPreview when selected item change

HeavenVolkoff:


jamiepine (approved):

interface/app/$libraryId/Explorer/View/index.tsx
```diff
@@ -14,12 +14,11 @@ import {
...
 import { createSearchParams, useNavigate } from 'react-router-dom';
 import {
-	ExplorerItem,
-	FilePath,
-	Location,
-	Object,
+	type ExplorerItem,
+	type FilePath,
+	type Location,
+	type Object,
```
Brendonovich, line 20: is this an eslint thing?

HeavenVolkoff: Not really, is more of a small optimization for the typescript typechecker performance, and should also help with tree-shaking:
https://stackoverflow.com/questions/61412000/do-i-need-to-use-the-import-type-feature-of-typescript-3-8-if-all-of-my-import

utkubakir (approved):

PR #1220, opened 2023-08-14
[ENG-904] Thumbnail for `svg`

HeavenVolkoff:
Introduces a new `sd-svg` crate inspired by `sd-heif` to allow integrating our `Thumbnailer` to parse SVGs:
[resvg](https://github.com/RazrFalcon/resvg) and [tiny-skia](https://github.com/RazrFalcon/tiny-skia) are used to parse and render the SVG and the RGBA data is converted to a `DynamicImage` that will be handled by the `Thumbnailer`

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -27,6 +27,7 @@ sd-crypto = { path = "../crates/crypto", features = [
 	"serde",
 	"keymanager",
 ] }
+sd-svg = { path = "../crates/svg" }
 sd-heif = { path = "../crates/heif", optional = true }
 sd-file-ext = { path = "../crates/file-ext" }
 sd-sync = { path = "../crates/sync" }
--- a/core/src/object/media/thumbnail/mod.rs
+++ b/core/src/object/media/thumbnail/mod.rs
@@ -136,6 +136,7 @@ pub async fn generate_image_thumbnail<P: AsRef<Path>>(
 	if metadata.len()
 		> (match ext {
 			"svg" => sd_svg::MAXIMUM_FILE_SIZE,
+			#[cfg(all(feature = "heif", not(target_os = "linux")))]
 			_ if HEIF_EXTENSIONS.contains(ext) => sd_heif::MAXIMUM_FILE_SIZE,
 			_ => MAXIMUM_FILE_SIZE,
 		}) {
--- a/core/src/object/media/thumbnail/mod.rs
+++ b/core/src/object/media/thumbnail/mod.rs
@@ -9,7 +9,8 @@ use crate::{
 };
 
 use sd_file_ext::extensions::{Extension, ImageExtension, ALL_IMAGE_EXTENSIONS};
-use sd_media_metadata::image::Orientation;
+use sd_media_metadata::image::{ExifReader, Orientation};
+use std::collections::HashSet;
 
 #[cfg(feature = "ffmpeg")]
 use sd_file_ext::extensions::{VideoExtension, ALL_VIDEO_EXTENSIONS};
@@ -19,14 +20,15 @@ use std::{
 	error::Error,
 	ops::Deref,
 	path::{Path, PathBuf},
+	sync::Arc,
 };
 
 use futures_concurrency::future::{Join, TryJoin};
-use image::{self, imageops, DynamicImage, GenericImageView};
+use image::{self, imageops, DynamicImage, GenericImageView, ImageFormat};
 use once_cell::sync::Lazy;
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
-use tokio::{fs, io, task::block_in_place};
+use tokio::{fs, io, task::spawn_blocking};
 use tracing::{error, trace, warn};
 use webp::Encoder;
 
@@ -86,6 +88,10 @@ pub enum ThumbnailerError {
 	FileIO(#[from] FileIOError),
 	#[error(transparent)]
 	VersionManager(#[from] VersionManagerError),
+	#[error("failed to encode webp")]
+	Encoding,
+	#[error("the image provided is too large")]
+	TooLarge,
 }
 
 #[derive(Debug, Serialize, Deserialize, Clone, Copy)]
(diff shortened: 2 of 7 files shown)


crates/svg/src/lib.rs
```diff
@@ -0,0 +1,77 @@
...
+const SVG_MAXIMUM_FILE_SIZE: u64 = 1048576 * 20;
+
+#[derive(Error, Debug)]
+pub enum SvgError {
+	#[error("error with usvg: {0}")]
+	USvg(#[from] resvg::usvg::Error),
+	#[error("error while loading the image (via the `image` crate): {0}")]
+	Image(#[from] image::ImageError),
+	#[error("io error: {0}")]
+	Io(#[from] std::io::Error),
```
fogodev, line 25: Please don't LMAO

I've been trying to abolish this kind of direct conversion from `io::Error` which doesn't tell what file the error comes from. On Core I've been using `FileIOError`, but for subcrates I'm doing.

```rust
#[error("io error on {}: {0}", .1.display())]
Io(std::io::Error, Box<Path>),
```

Then using `.map_err(|e| SvgError::Io(e, path.to_path_buf().into_boxed_path())?` for error handling. This way, whenever a `io::Error` pops out, we know on which file it happened.

crates/svg/src/lib.rs
```diff
@@ -0,0 +1,77 @@
...
+	Pixbuf,
+	#[error("there was an error while converting the image to an `RgbImage`")]
+	RgbImageConversion,
+	#[error("failed to calculate thumbnail size")]
+	InvalidSize,
+	#[error("the image provided is too large (over 20MiB)")]
+	TooLarge,
+}
+
+pub async fn svg_to_dynamic_image(path: &Path) -> SvgResult<DynamicImage> {
```
fogodev, line 36: As this function is exposed in the public API, change it to receive `(path: impl AsRef<Path>)` instead of a `&Path` please.

crates/svg/src/lib.rs
```diff
@@ -0,0 +1,77 @@
...
+	if fs::metadata(path).await?.len() > SVG_MAXIMUM_FILE_SIZE {
+		return Err(SvgError::TooLarge);
+	}
+
+	let data = fs::read(path).await?;
+
+	let opt = usvg::Options::default();
+
+	let mut fontdb = fontdb::Database::new();
+	fontdb.load_system_fonts();
```
fogodev, line 46: Do this function call perform any kind of I/O internally consulting some OS files? If so it need to runs in a `block_in_place` to avoid blocking the runtime.

core/src/object/preview/thumbnail/mod.rs
```diff
@@ -113,27 +113,40 @@ pub async fn generate_image_thumbnail<P: AsRef<Path>>(
...
+					image::open(file_path.as_ref())?
+				}
+			};
+
+			#[cfg(not(all(feature = "heif", not(target_os = "linux"))))]
+			let img = image::open(file_path)?;
 
-		#[cfg(not(all(feature = "heif", not(target_os = "linux"))))]
-		let img = image::open(file_path)?;
+			img
```
fogodev, line 148: If the file is an SVG, we don't even need this `block_in_place` call here, you can merge the above if let to already get webp data from svg or trying to process with heif or `image` lib otherwise in the `block_in_place`

fogodev (changes requested):

core/Cargo.toml
```diff
@@ -27,6 +27,7 @@ sd-crypto = { path = "../crates/crypto", features = [
 	"serde",
 	"keymanager",
 ] }
+sd-svg = { path = "../crates/svg" }
```
brxken128, line 30: I'll leave this in place, but I'll move the code to my `sd-images` WIP crate to unify everything

core/src/object/media/thumbnail/mod.rs
```diff
@@ -101,39 +106,66 @@ pub struct ThumbnailerMetadata {
...
+		_ if HEIF_EXTENSIONS.contains(ext) => {
+			#[cfg(not(all(feature = "heif", not(target_os = "linux"))))]
+			return Err("HEIF not supported".into());
+			#[cfg(all(feature = "heif", not(target_os = "linux")))]
+			sd_heif::heif_to_dynamic_image(data.clone()).await?
+		}
+		_ => image::load_from_memory_with_format(
+			&fs::read(file_path).await?,
+			ImageFormat::from_path(file_path)?,
+		)?,
```
brxken128, line 163: All of this will be cleaned up by my snazzy image handler too, as eventually we're going to have too many of these "exotic" formats to deal with and matching like this will be very cumbersome and unsightly.

core/src/object/media/thumbnail/mod.rs
```diff
@@ -143,21 +175,29 @@ pub async fn generate_image_thumbnail<P: AsRef<Path>>(
...
-		if let Some(x) = orientation {
-			img = x.correct_thumbnail(img);
+		match ExifReader::from_slice(data.as_ref()) {
+			Ok(exif_reader) => {
+				// this corrects the rotation/flip of the image based on the available exif data
+				if let Some(orientation) = Orientation::from_reader(&exif_reader) {
+					img = orientation.correct_thumbnail(img);
+				}
+			}
+			Err(e) => warn!("Unable to extract EXIF: {:?}", e),
```
brxken128, line 185: I disagree here. Not all images have EXIF data, and if they don't that's perfectly okay - we shouldn't be erroring if it's not present, there's multiple other types of image metadata and we currently only check for the one. Some people even strip this metadata completely, so this would always bring up a warning.

crates/heif/src/lib.rs
brxken128, line 1: I'm not particularly happy with these changes - they were going to be revamped in my next PR anyway but doing it this way makes it harder for us to support non-interleaved HEIF images (which do exist, they are rare but ENG-984 was made to address this issue.

brxken128 (approved):

brxken128 (approved):
Works great for me, the code is solid, we can get this merged into a single `sd-image` handling crate now! Nice job man, keep up the great work - you're killing it 🙏🏻

crates/svg/src/lib.rs
brxken128: ^I would like to note that with clippy 1.72.0 this causes a lot of missing `Send` errors, whereas `&Path` doesn't from my experience

PR #1226, opened 2023-08-17
Refactor explorer settings to use useExplorer and useExplorerSettings

Brendonovich:
This PR isn't targeted at `main`.

A lot of `explorerStore` is now gone and replaced with a settings store that can be created using `useExplorerSettings`, which is then passed into `useExplorer` (the hooks are separate bc `items` of `useExplorer` can be dependent on settings).

This will probably introduce some bugs that we'll need to tackle later (one being properly handling default ordering settings), but pleases both rust + ts and is a design i'm comfortable with for now.


ameer2468:
Lgtm!

ameer2468 (approved):

PR #1228, opened 2023-08-17
[ENG-976] Fixed Missing scrollbar in the Explorer

### LINKED ISSUE #1219: [ENG-976] Missing scrollbar in the Explorer
This could potentially be Linux (or even Gnome) specific, but the Explorer has no scrollbar which makes scrolling incredibly difficult. I'm not sure if this happens on Windows or MacOS too

<sub>From [SyncLinear.com](https://synclinear.com) | [ENG-976](https://linear.app/spacedriveapp/issue/ENG-976/missing-scrollbar-in-the-explorer)</sub>

Raghav-45:
Hey @brxken128,

I appreciate your observation regarding the missing scrollbar on the Explorer. I've looked into this, and it turns out that the scrollbar is actually present but not easily visible due to its color being the same as the background of the Explorer. However, you can still interact with it by hovering over it and dragging from the right edge to initiate scrolling.

Here are some screenshots - Before Fix:
Screenshot (140)

Screenshot After Fix:
Screenshot (141)

To address this issue, I removed the `custom-scroll` Class that was causing the scrollbar to be constantly visible. Instead, I've added the `overflow-y-auto` Class, which ensures that the scrollbar appears only when needed.
Additionally, I've made some modifications to the CSS file (`.scss`) to ensure that the scrollbar is visible. Previously, it was transparent.

Best regards,
Aditya

Closes #1219

### DIFF
--- a/interface/app/$libraryId/Explorer/index.tsx
+++ b/interface/app/$libraryId/Explorer/index.tsx
@@ -44,7 +44,7 @@ export default function Explorer(props: PropsWithChildren<Props>) {
 				<div className="flex-1 overflow-hidden">
 					<div
 						ref={explorer.scrollRef}
-						className="custom-scroll explorer-scroll relative h-screen overflow-x-hidden"
+						className="explorer-scroll relative h-screen overflow-x-hidden overflow-y-auto"
 						style={{
 							paddingTop: TOP_BAR_HEIGHT,
 							paddingRight: explorerStore.showInspector ? INSPECTOR_WIDTH : 0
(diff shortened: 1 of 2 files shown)


utkubakir:
No need to bump the width, I think we wanna keep the thinness of it

Raghav-45:
> No need to bump the width, I think we wanna keep the thinness of it

Now it should be good, i thought increasing it by 2px will make it more visible.

utkubakir (approved):

PR #1233, opened 2023-08-18
[ENG-972] quick view improvements

niikeec:


niikeec:
> this is amazing.
> 
> can we get a loading indicator before the original media is displayed? if the component is trying to load an original, the thumbnail isn't shown, instead just the blurred background. before it was just the bg color.
> 
> not sure if there's room in this PR but it could be a great opportunity to show a "failed to load original" icon or indicator in a corner if for example the location is offline.

There's an errorFallback in place as an optional prop but it's probably worth also adding a default one. For the loading state, I do like the transition from a blurred thumbnail to the original one (current version), but adding a loading indicator might be a better idea.

utkubakir (dismissed):

utkubakir (approved):

PR #1237, opened 2023-08-21
[ENG-980] Barebones, essential telemetry

brxken128:
New telemetry default entirely anonymous and backed by [Plausible](https://plausible.io/).

This changes onboarding to have to privacy options: minimal telemetry, or full telemetry. Minimal telemetry just shares that you're an active user, along with which version of the core/app you're running, width of your display and type of device (e.g. phone, desktop, web). This contains zero PII and absolutely nothing that could be linked back to you, as Plausible would never allow that to happen (and they handle all of the IP-stripping anyway).

This adds a loop so for every 4.75minutes that passes, we send a `ping` event to Plausible to let them know that you're still "active" and on Spacedrive (as you very well could just be waiting for a large task to run in the back). This also sends a `ping` event at the same time as every `pageview` event, as they're placed directly next to eachother in the code.

The latter may be redundant and the timer may be more than enough, but only time will tell.

The full telemetry option includes all events, such as library creation/deletion, tag creation/deletion, location addition/removal and a few more. There's always room for extra.


interface/app/$libraryId/Layout/index.tsx
```diff
@@ -30,6 +33,21 @@ const Layout = () => {
...
+	usePlausiblePingMonitor({ currentPath: rawPath });
+
+	useEffect(() => {
+		const interval = setInterval(() => {
+			plausibleEvent({
+				event: {
+					type: 'ping'
+				}
+			});
+		}, 270 * 1000);
```
utkubakir, line 45: half minute or my math is bad?

Brendonovich: 270,000ms = 4.5min

interface/app/$libraryId/Layout/index.tsx
```diff
@@ -30,6 +33,21 @@ const Layout = () => {
...
+		console.log('ran');
+
+		return () => clearInterval(interval);
+		// eslint-disable-next-line react-hooks/exhaustive-deps
+	}, []);
```
Brendonovich, line 50: no ignore eslint plz

utkubakir:
Can we add it to mobile too? Probably copying your code where `usePlausiblePageViewMonitor` is should work.

utkubakir (approved):

PR #1190, opened 2023-08-09
P2P Assorted Changes

oscartbeaumont:
Brendan has been having some reliability issues with a certain component. So investigating that and fixing a bunch of other random stuff.

Changes:
 - [x] Upgrade libp2p
 - [x] Decouple discovery
 - [ ] Discovery event bus for component communication
 - [ ] Connection state that is updated on stream open/receive so we don't need the header?
 - [x] Delayed construction of `MetadataManager`
 - [ ] Regig instance advertisements
 - [ ] Respond to advertisements with debounce for more reliable discovery.
 - [ ] Connection Header + decouple receiving?
 - [ ] Remove SD's connection status stuff and push more into the P2P layer?
 - [ ] Stream multiplexing
 - [ ] ~~`Discovery` abstraction on mdns~~
 - [ ] Renaming `ManagerStream` into `p2p::System`?
 - [ ] Logging into the P2P system as it's more being removed from `sd-core`?
 - [ ] Proper instance shutdown

PR #1238, opened 2023-08-22
[ENG-994] Minor P2P fixes

oscartbeaumont:
https://linear.app/spacedriveapp/issue/ENG-994/minor-p2p-fixes

#1190 is taking longer than I wanted.


core/src/api/p2p.rs
```diff
@@ -23,11 +23,13 @@ pub(crate) fn mount() -> AlphaRouter<Ctx> {
 						};
 					}
 
-					// // TODO: Don't block subscription start
-					// for peer in ctx.p2p_manager.get_connected_peers().await.unwrap() {
-					// 	// TODO: Send to frontend
-					// }
 
+					// TODO: Don't block subscription start
+					for peer_id in node.p2p.manager.get_connected_peers().await.unwrap() {
```
fogodev, line 28: Fix warning please, with proper error handling or with `expect` if this can't fail

core/src/lib.rs
```diff
@@ -142,49 +159,13 @@ impl Node {
...
-							)
-					} else {
-						EnvFilter::from("info")
-					}),
+					.with_filter(
+						EnvFilter::builder()
+							.from_env()
+							.unwrap()
+							// We don't wanna blow up the logs
+							.add_directive("sd_core::location::manager=info".parse().unwrap()),
```
fogodev, line 167: These can just be `expect` calls, as they can't fail anyway, due to parsing being from a hardcoded string

core/src/api/p2p.rs
oscartbeaumont: I am of the opinion this one is fine for now because `core/p2p/mod.rs` already has `#![allow(clippy::unwrap_used, clippy::panic)]` which is something that will be removed once it is stabilised.

This is also only not possible to call without the beta P2P feature flag being enabled on the frontend.

core/src/lib.rs
oscartbeaumont: Have fixed!

Brendonovich (approved):

PR #1241, opened 2023-08-23
show "see more" if locations > 5, not if locations >= 5

PineappleRind:
"See more" is shown when locations.length == 5

Maximum locations that can be shown == 5

so "See more" is needed when locations.length > 5

https://github.com/spacedriveapp/spacedrive/assets/65303441/6996696c-acb2-4580-bdf9-736490d81371


HeavenVolkoff (approved):

HeavenVolkoff:
Nice catch

PR #1242, opened 2023-08-23
[ENG-939, ENG-1173] PDF Thumbnails

HeavenVolkoff:
- [X] Implement PDF handler in `sd-images` (Load, render and convert PDF to `DynamicImage`)
- [X] Integrate with Thumbnailer
- [X] ~Add build steps for PDFium to each OS target (https://github.com/bblanchon/pdfium-binaries):~ (Done in: #1368 and #1405)

~While this PR can be merged as is, it will only work correctly in all supported OSes after #1405 is merged~


fogodev (approved):
LGTM

PR #1157, opened 2023-08-02
feat: shortcuts page

Raduc4:
This PR is just a base representation of how the shortcuts page can look like.
I'm looking for feedback to improve this.

[Issue](https://github.com/spacedriveapp/spacedrive/issues/1112)

Issue #1112, opened 2023-07-15
[ENG-815] Shortcuts page


utkubakir:
Let's show our shortcuts on a page (in settings).  They will be customizable/changeable in the future but for now just displaying them will do. We can section them by their place e.g.

* General
  * Mouse 4 & Alt + Left → Navigate back

* Explorer
  * Command + V → Paste

<sub>From [SyncLinear.com](https://synclinear.com) | [ENG-815](https://linear.app/spacedriveapp/issue/ENG-815/shortcuts-page)</sub>

Raduc4:
assigned

PineappleRind:
I took a go at it 
<img width="1018" alt="image" src="https://github.com/spacedriveapp/spacedrive/assets/65303441/95551fb1-7054-4ba1-be58-b854691a5e38">

PR #1245, opened 2023-08-24
Attempt shortcuts page

PineappleRind:
Relevant issue: #1112 

This is the successor of #1157

Screenshot 2023-08-24 at 10 21 41 AM


utkubakir:
Need some cleanup with the paddings etc but overall, I like it!

utkubakir:
Can we increase the padding a bit? Keys look crowded

ameer2468:
@PineappleRind - please re-open if this is still something you're interested on working on. So far so good!

Screenshot 2023-09-14 at 1 04 39 PM

You need add some spacing in-between.

PineappleRind:
alright, will do, thanks! 👍

PineappleRind:
added more vertical spacing between keybinds & used `keybindForOs` for OS-specific keybinds
Screenshot 2023-09-14 at 11 51 39 AM

also @ameer2468 - I'm not sure if GitHub allows me to reopen the PR?

ameer2468:
> added more vertical spacing between keybinds & used `keybindForOs` for OS-specific keybinds Screenshot 2023-09-14 at 11 51 39 AM
> 
> also @ameer2468 - I'm not sure if GitHub allows me to reopen the PR?

Thanks for the response. Re-opened.

I'd remove the navigation shortcuts which are: ```⌘ + ]``` and ```⌘ + [``` - these are not being used for navigation.

We also don't have shortscuts added yet for showing enclosing folder and opening a selected item.

Though space does a quick preview.

I'm gonna add support for these shortcuts soon.

PineappleRind:
I've updated the data:

image

The `keybind` function returns a single string for macOS (like `⌘→`), so I've just been splitting it by character (`.split('')`), but realized that doesn't work for keybinds which have multiple letters in one key (like the "action with fn" action, for demonstration purposes)

Another option is to split by the plus sign:

Screenshot 2023-09-14 at 8 23 31 PM

On macOS it groups the keys together, but from what it looks like from the function, when not on macOS it will separate the keys (because the keybind function returns `key+key` instead of `keykey`)

So I could change the keybind function to return an array, or change the styling on macOS so it looks like the native context menus... I'm not really sure what to do here 😅

ameer2468:
> I've updated the data:
> 
> image
> _(The fn action was for testing purposes, it's not there now)_
> 
> The `keybind` function returns a single string for macOS (like `⌘→`), so I've just been splitting it by character (`.split('')`), but realized that doesn't work for keybinds which have multiple letters in one key (like the "action with fn" action)
> 
> Another option is to split by the plus sign:
> 
> Screenshot 2023-09-14 at 8 35 48 PM
> On macOS it groups the keys together, but from what it looks like from the function, when not on macOS it will separate the keys (because the keybind function returns `key+key` instead of `keykey`)
> 
> So I could change the keybind function to return an array, or change the styling on macOS so it looks like the native context menus... I'm not really sure what to do here 😅

Hey thanks for the continuous effort on this!

It's completely fine to have it without the '+' - because I'm working on tooltips offering shortcuts aswell, and they will not have a '+'.

**Take this design as an example** 

image

Let's skip adding any form of '+' and just display the keys.

PR #1249, opened 2023-08-25
[ENG-1002] update executable icon

niikeec:


utkubakir:
Let's remove the old one, no need to store it as it should be in figma anyway.

jamiepine (approved):

PR #1258, opened 2023-08-28
[ENG-1009] Fetch item count to increase scrollbar size in explorer

Brendonovich:
This PR doesn't alter pagination behaviours, but just fetches the number of items in a particular search and passes it along to the explorer accordingly. How we want to handle pagination can be figured out later.

### DIFF
--- a/interface/app/$libraryId/Explorer/View/ListView.tsx
+++ b/interface/app/$libraryId/Explorer/View/ListView.tsx
@@ -274,7 +274,7 @@ export default () => {
 	const tableLength = table.getTotalSize();
 
 	const rowVirtualizer = useVirtualizer({
-		count: explorer.items ? rows.length : 100,
+		count: explorer.count ?? explorer.items ? rows.length : 100,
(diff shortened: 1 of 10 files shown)


interface/app/$libraryId/Explorer/View/ListView.tsx
```diff
@@ -274,7 +274,7 @@ export default () => {
 	const tableLength = table.getTotalSize();
 
 	const rowVirtualizer = useVirtualizer({
-		count: explorer.items ? rows.length : 100,
+		count: explorer.count ?? explorer.items ? rows.length : 100,
```
niikeec, line 277: The default count can be set to 0 instead of 100 as we don't use skeleton loading anymore

niikeec: Actually nvm, it seems like we still do on the list view.  Could we then also remove the skeleton loader?

Brendonovich: Done both

niikeec (approved):

PR #1266, opened 2023-08-29
Remove react devtools dependency

Brendonovich:
`pnpm dlx` can do the job, no need for installing it as a dev dependency


brxken128 (approved):

oscartbeaumont:
No more Electron in the Spacedrive repo!!!

PR #1270, opened 2023-08-29
[ENG-944] Sidebar UX Improvements

jamiepine:
This makes the Sidebar make sense.

- Added "See more" as a default property of a group or sidebar buttons.
- Added custom icons to ephemeral section, which is renamed "Local" from "Explore".
- Added network screen, showing local peers.
- Removed unused swift package `macOS`.
- Created a watcher for system volumes to invalidate ui when drives are added/removed.
- Clicking a volume that is also a location will open the location instead of the ephemeral explorer.
- Removed fake devices "Spacephone" and "Titan" from the sidebar when in dev mode.
- Added Locations to Inspector, revealing which Locations an Object is present in at least once.

PR is a little large, apologies.

Screenshot 2023-10-08 at 6 31 56 AM

Screenshot 2023-10-08 at 8 06 31 AM


utkubakir:
Can we clean and merge this if it's done?

utkubakir (approved):
lgtm

PR #1279, opened 2023-08-30
[ENG-809] Enter key submits dialog

ameer2468:
- Enter key submits the form within the dialog.
- Dialog only closes if form is valid and successful.


ameer2468:
> I feel like this should be resolved by being more careful in how we manage focus state. Event listeners are an ugly workaround, if a button is focused it will trigger when enter is pressed.

I'll take a closer look

PR #1280, opened 2023-08-30
[ENG-953] Zero-byte files occasionally receive a random thumbnail

### LINKED ISSUE #1172: [ENG-953] Zero-byte files occasionally receive a random thumbnail
Zero-byte files occasionally receive a random thumbnail, which can be misleading when an unrelated image appears where it shouldn't.

<sub>From [SyncLinear.com](https://synclinear.com) | [ENG-953](https://linear.app/spacedriveapp/issue/ENG-953/zero-byte-files-occasionally-receive-a-random-thumbnail)</sub>

fogodev:
We were generating a bogus `cas_id` for empty files, sometimes it messed things up for empty files, like assigning the same thumbnail or messing the file object kind.

Closes #1172

### DIFF
--- a/core/src/location/manager/watcher/utils.rs
+++ b/core/src/location/manager/watcher/utils.rs
@@ -269,7 +269,7 @@ async fn inner_create_file(
 	let created_file = create_file_path(
 		library,
 		iso_file_path,
-		Some(cas_id.clone()),
+		cas_id.clone(),
 		FilePathMetadata {
 			inode,
 			device,
@@ -285,7 +285,7 @@ async fn inner_create_file(
 	let existing_object = db
 		.object()
 		.find_first(vec![object::file_paths::some(vec![
-			file_path::cas_id::equals(Some(cas_id.clone())),
+			file_path::cas_id::equals(cas_id.clone()),
 			file_path::pub_id::not(created_file.pub_id.clone()),
 		])])
 		.select(object_just_id::select())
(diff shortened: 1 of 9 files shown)


brxken128:
`touch {a..z}` should help out, creates a bunch of empty files. I'll test this PR quickly 🙏🏻

core/src/location/manager/watcher/utils.rs
```diff
@@ -464,113 +466,115 @@ async fn inner_update_file(
...
-						))),
-					),
-					{
-						let date = DateTime::<Utc>::from(fs_metadata.modified_or_now()).into();
+						(date_modified::NAME, json!(date)),
+						Some(date_modified::set(Some(date))),
+					)
+				},
+				{
+					// TODO: Should this be a skip rather than a null-set?
```
brxken128, line 496: I think skip here as the integrity checksum would be the same as all other null files, and that may cause issues in the future. But on the flip side, if users want to delete _all_ null files at once this would be the way to do it. I'm on the fence now haha

brxken128: I'd say just `None` it for now and we can come back to redo the implementation once we get to the more complex filesystem action functionality.

brxken128 (approved):
Works a charm from my testing, they still get unique `pub_id`s (as they should) but they aren't treated as a file type that they're not so that's a plus. Nice work man!

core/src/location/manager/watcher/utils.rs
fogodev: Will keep this in mind

PR #1281, opened 2023-08-31
[ENG-1039] Apply hard `clippy` lints to the entire `ffmpeg` crate

brxken128:
This was done in hopes of making it both safer, and slighty more performant (and easier to use in external crates, such as the metadata one!).

We don't have `miri` in CI yet, so a lot of `unsafe` issues may go unnoticed, and this aims to hopefully solve any (if there are/were any to begin with).

This also removes an old comment, and changes `Webp` to `Thumbnail` in the context of "Webp creation has blocking code" within a comment.


crates/ffmpeg/src/movie_decoder.rs
```diff
@@ -217,9 +217,13 @@ impl MovieDecoder {
 			));
 		}
 
-		video_frame.width = unsafe { (*new_frame.as_mut_ptr()).width as u32 };
-		video_frame.height = unsafe { (*new_frame.as_mut_ptr()).height as u32 };
-		video_frame.line_size = unsafe { (*new_frame.as_mut_ptr()).linesize[0] as u32 };
+		// SAFETY: these should always be positive, so clippy doesn't need to alert om them
```
oscartbeaumont, line 220:
```suggestion
		// SAFETY: these should always be positive, so clippy doesn't need to alert on them
```

crates/ffmpeg/src/movie_decoder.rs
```diff
@@ -256,7 +260,9 @@ impl MovieDecoder {
 		Ok(())
 	}
 
-	pub(crate) fn get_video_duration(&self) -> Duration {
+	// SAFETY: this should always be positive, so clippy doesn't need to alert om them
```
oscartbeaumont, line 263:
```suggestion
	// SAFETY: this should always be positive, so clippy doesn't need to alert on them
```

crates/ffmpeg/src/thumbnailer.rs
```diff
@@ -125,24 +125,28 @@ impl ThumbnailerBuilder {
 	/// - `quality`: 80
 	/// - `prefer_embedded_metadata`: true
 	/// - `with_film_strip`: true
+	#[must_use]
```
oscartbeaumont, line 128: I feel like a `#[must_use]` on the new method is a bit pointless.

oscartbeaumont: That being said I don't like the type implementating `Default` and all of the settings being `#[must_use]` cause all of the methods being `#[must_use]` implies it doesn't have a default.

oscartbeaumont (dismissed):
Looks great!

I left a couple of comments but they are all nitpicks so they don't really *need* to block merge.

crates/ffmpeg/src/thumbnailer.rs
brxken128: I agree, I was just following clippy in this instance but we can probably alter it to require not so many `#[must_use]`s. I myself don't dislike them and use them where possible/practical (or whenever clippy tells me to, really), but I can see where you're coming from and I'll try to clean some of that up.

brxken128: I think that should do it - I just marked the whole type as `#[must_use]` instead, so all of the sub-functions don't need to worry about it.

oscartbeaumont: I think I am wrong here about the combination of `Default` and `#[must_use]` being a problem. I thought `#[must_use]` applied to the function (e.g. you must call all the functions) but actually, it applies to the return values (Eg. must use return type) which makes sense.

```rust
ThumbnailerBuilder::default().some_method(); // Would warn about must_use.

ThumbnailerBuilder::default().some_method().build(); // Would be valid.

**Review by Brendonovich — CHANGES_REQUESTED** [review]

File: core/src/object/media/thumbnail/mod.rs [hunk 4]
```diff
@@ -106,6 +106,7 @@ pub struct ThumbnailerMetadata {
 	pub skipped: u32,
 }
 
+#[cfg(all(feature = "heif", not(target_os = "linux")))]
```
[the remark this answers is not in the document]
brxken128: Yeah we should probably just make them always available instead of splitting them up like this.

My next PR entirely cleans this up, so we can fix it now or just wait for that to be merged (which won't be too long).

Brendonovich (approved):

PR #1282, opened 2023-08-31
[ENG-566] Stream files through P2P

oscartbeaumont:
Pretty self explanatory.


oscartbeaumont:
TODO:
 - [ ] Fix feature flagging -> Doesn't restore after restart
 - [ ] Make damn sure everything is feature flag cause we don't wanna leak peoples files to the network
 - [ ] Check I am doing file path joining correctly. I couldn't find it but I swear I did it the *really* wrong way.

Brendonovich (approved):
very fucking cool

PR #1287, opened 2023-09-01
[ENG-1028] Media data for ephemeral locations

fogodev:
New rspc route to extract media data from a non-indexed file on demand

Fixing some minor stuff

Running pnpm format

### DIFF
--- a/crates/sync/example/web/index.html
+++ b/crates/sync/example/web/index.html
@@ -1,4 +1,4 @@
-<!DOCTYPE html>
+<!doctype html>
 <html lang="en">
 	<head>
 		<meta charset="utf-8" />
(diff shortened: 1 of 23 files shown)


crates/sync/example/web/index.html
```diff
@@ -1,4 +1,4 @@
-<!DOCTYPE html>
+<!doctype html>
```
brxken128, line 1: According to [this](https://html.spec.whatwg.org/multipage/syntax.html#the-doctype), it should be `<!DOCTYPE html>`. It does say it's case-insensitive, but all caps seems to be the "unofficial" standard.

Brendonovich: if pnpm format says lowercase then i saw lowercase it is

brxken128: `pnpm format` is the speaker of truth i guess LMAO

brxken128 (approved):

Brendonovich (approved):

PR #1288, opened 2023-09-02
[ENG-1063] Actually regenerate thumbnails on `Regen Thumbnails` button

fogodev:
Introducing a new `renegerate` param to `MediaProcessorJob` asking for regenerating thumbnails already existing instead of skipping.


brxken128 (approved):
Seems to work great from my testing!

PR #1290, opened 2023-09-02
[ENG-1064] replace alert dialogs with toasts

niikeec:


brxken128 (approved):
Works great and looks super sleek, I love this change. Nice job as usual!

Screenshot depicting a pop-up toast alerting the user of an error

PR #1292, opened 2023-09-04
[ENG-1070] Dialog form button submit disabled if form is not valid

ameer2468:
- [x] Forms within dialogs, submit button is disabled if the form is not valid.
- [x] Feedback dialog cleaned up and removed un-necessary code
- [x] TextAreaField component, this is already similar to existing ones like InputField component. This makes the error side of textarea visible and the styling changes.

- [x]  Has been tested

### DIFF
--- a/interface/app/$libraryId/Layout/Sidebar/FeedbackDialog.tsx
+++ b/interface/app/$libraryId/Layout/Sidebar/FeedbackDialog.tsx
@@ -1,10 +1,10 @@
 import clsx from 'clsx';
 import { useState } from 'react';
 import { useZodForm } from '@sd/client';
-import { Dialog, TextArea, UseDialogProps, toast, useDialog, z } from '@sd/ui';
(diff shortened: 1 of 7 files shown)


interface/app/$libraryId/Layout/Sidebar/FeedbackDialog.tsx
```diff
@@ -65,10 +62,10 @@ export default function FeedbackDialog(props: UseDialogProps) {
 				</div>
 			}
 		>
-			<TextArea
+			<TextAreaField
+				{...form.register('feedback', { required: true })}
```
Brendonovich, line 66: Does this `required` do anything? From what I can tell, using zod disables RHF's validation.

ameer2468: > Does this `required` do anything? From what I can tell, using zod disables RHF's validation.

Removed - not required

Brendonovich (approved):

PR #1298, opened 2023-09-05
[ENG-285] Document officially supported targets

brxken128:
This PR may be incomplete or contain incorrect information - I had to go digging around to find our target versions for mobile. We have our Android SDK minimum set at 21 I believe, and iOS 13.

We may also be able to build sole `x86` builds for some platforms, and maybe even `Aarch64` builds for Windows - I'm unsure and will leave this as a draft until we get further clarification.


utkubakir:
I don't think we need to have this on readme, maybe a page on `docs/` ?

brxken128:
> I don't think we need to have this on readme, maybe a page on `docs/` ?

I was unsure about this one, if we have it in the readme then users are able to easily see if they can even use our app, instead of digging through the docs to potentially be disappointed. @Brendonovich what do you think we should do here?

brxken128:
I spoke with @Brendonovich and it was agreed that we should probably define these targets in both the README and the documentation. I also added a link to the contributing guide in the setup docs, which will help people build the app from source when they view the docs.

brxken128:
These are all technically correct, except we only currently build releases for 64-bit desktop systems so I'm on the fence about what/what not to include.

oscartbeaumont:
I personally feel like this is better off in the `CONTRIBUTING.md` than `README.md` but tbh does it really need to be defined in two places? They will just get out of sync and we can easily link between them so one defining the targets and the other being a link at most would probs be fine right?

brxken128:
@oscartbeaumont True, and we don't even have release builds configured for the majority of these just yet. I'll close the PR for now, but re-open and make the changes once we do!

PR #1306, opened 2023-09-07
[ENG-941] Jobs running in another library do not appear in the job manager

### LINKED ISSUE #1162: [ENG-941] Jobs running in another library do not appear in the job manager
Jobs running in another library do not appear in the job manager, even though the indicator is active. This could be perplexing. There should be an indication for this circumstance.

<sub>From [SyncLinear.com](https://synclinear.com) | [ENG-941](https://linear.app/spacedriveapp/issue/ENG-941/jobs-running-in-another-library-do-not-appear-in-the-job-manager)</sub>

fogodev:
The rspc route returns job reports clustered by library ids, but the frontend isn't able to handle that yet, maybe it will need some UI changes for that, which aren't my strong suit.

For now the frontend just gather and shows up job reports from all libraries together without any indication of from what library that report belongs to.

Some minor clippy warnings

pnpm format

Closes #1162

### DIFF
--- a/core/src/custom_uri.rs
+++ b/core/src/custom_uri.rs
@@ -206,8 +206,8 @@ pub fn router(node: Arc<Node>) -> Router<()> {
 							}
 
 							// TODO: Support `Range` requests and `ETag` headers
-
-							match *state.node.nlm.state().await.get(&library_id).expect("library exists").instances.get(&identity).expect("instance exists") {
+							#[allow(clippy::unwrap_used)]
+							match *state.node.nlm.state().await.get(&library_id).unwrap().instances.get(&identity).unwrap() {
 								InstanceState::Discovered(_) | InstanceState::Unavailable => Ok(not_found(())),
 								InstanceState::Connected(peer_id) => {
 									let (tx, mut rx) = tokio::sync::mpsc::channel::<io::Result<Bytes>>(150);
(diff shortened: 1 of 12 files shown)


core/src/custom_uri.rs
```diff
@@ -203,7 +207,7 @@ pub fn router(node: Arc<Node>) -> Router<()> {
 
 							// TODO: Support `Range` requests and `ETag` headers
 
-							match state.node.nlm.state().await.get(&library_id).unwrap().instances.get(&identity).unwrap().clone() {
+							match *state.node.nlm.state().await.get(&library_id).expect("library exists").instances.get(&identity).expect("instance exists") {
```
oscartbeaumont, line 210: These should stay `unwrap` and Clippy should override on them. `expect` imples no need for error handling but here there is, it's just in a feature that can only be triggered by enabling experimental feature flags.

Sorry, missed Clippy on my PR to bring in these warnings.

core/src/custom_uri.rs
```diff
@@ -622,7 +626,9 @@ impl AsyncWrite for MpscToAsyncWrite {
 	) -> Poll<Result<usize, io::Error>> {
 		match self.0.poll_reserve(cx) {
 			Poll::Ready(Ok(())) => {
-				self.0.send_item(Ok(Bytes::from(buf.to_vec()))).unwrap();
+				self.0
```
oscartbeaumont, line 629: Same here. That should be probally be properly handled not just asserted.

core/src/custom_uri.rs
fogodev: Sure thing, will revert them back

oscartbeaumont (approved):

PR #1312, opened 2023-09-07
Feature/target gate HEIF on Linux and Windows

brxken128:
Due to Debian's extremely delayed update regarding the `libheif-dev` package, I've had to disable HEIF completely again on Linux, even though pretty much *every* other distro has an up-to-date and supported package. Ubuntu 23.04 has a supported package, but Pop_OS! doesn't offer 23.04 as I believe they're busy working on their Cosmic DE.

This has been tested by @gedeondoescode in the DIscord, and it was them who originally brought the issue up (they're using Pop_OS!, which lacks the up to date packages required). I'd like to thank you for testing out this patch, as I don't have anything Debian-based to test it on myself.


oscartbeaumont:
I wonder if we could just require `pnpm build --features linux-heif` to build with heif on Linux so it's still *possible* just disabled by default.

brxken128:
> I wonder if we could just require `pnpm build --features linux-heif` to build with heif on Linux so it's still _possible_ just disabled by default.

I actually attempted to add a `cfg(linux_libheif)` so that `RUSTFLAGS="--cfg linux_libheif" pnpm desktop dev` would enable it, but I struggled with getting the `cfg`s correct lol

brxken128:
Converting to draft for now until I get chance to add a `--cfg` or a feature.

brxken128:
I've upgraded the `libheif` deps, which is fine as we build 1.16.2 (used by 99% of distros, just not Debian-based ones).

The `libheif` related deps added `bindgen` to their default features, so in order to upgrade them I had to disable default features for both crates, but the app builds and generates HEIF/HEIC thumbnails perfectly still.

Linux systems with `libheif-dev(el)` 1.16.2 should be able to enable HEIF/HEIC thumbnails with `RUSTFLAGS="--cfg linux_libheif" pnpm desktop dev`. We should probably add this `cfg` to the server too, so that @jamiepine can generate HEIF thumbnails on his server.

jamiepine:
well HEIF generation works great on freebsd (my nas), so don't break it

brxken128:
@jamiepine your NAS runs TrueNAS SCALE which is Debian based, we'd just build the server/container with the cfg flag

utkubakir (approved):
lgtm

PR #1313, opened 2023-09-08
[ENG-1084] Correctly format `Orientation` EXIF data in the UI

brxken128:
This was done with a TS function and an object, it's quite possible that we don't need the function but other components had them so I followed.


Brendonovich (changes requested):
Formatting like this should be done in TS - backend should only send identifiers, not full strings

jamiepine (approved):

brxken128:
> Formatting like this should be done in TS - backend should only send identifiers, not full strings

I'll get this changed later today

interface/app/$libraryId/Explorer/Inspector/MediaData.tsx
```diff
@@ -15,8 +15,22 @@ function formatMediaTime(loc: MediaTime): string | null {
...
+		MirroredHorizontalAnd90CW: 'Mirrored horizontally and rotated 90° clockwise',
+		MirroredHorizontalAnd270CW: 'Mirrored horizontally and rotated 270° clockwise',
+		MirroredVertical: 'Vertically mirrored',
+		CW90: 'Rotated 90° clockwise',
+		CW180: 'Rotated 180° clockwise',
+		CW270: 'Rotated 270° clockwise'
+	};
+
+	return orientations[orientation] ?? null;
 }
```
Brendonovich, line 34: A dedicated function here isn't necessary, you can just define an object at the top level and do the lookup in the JSX

Brendonovich (changes requested):

Brendonovich (approved):

PR #1314, opened 2023-09-08
Update Rust dependencies

oscartbeaumont:
General:
 - Left `tauri` and `notify` version
 - Didn't touch `libheif` or `ffmpeg-sys-next`
 - Left the entire Crypto crate outta this

P2P:
 - Did a heap to P2P to deal with libp2p's and ed25519-dalek's breaking changes
 - sd-p2p doesn't leak a small amount of memory anymore. Thanks libp2p update!

Patches:
 - Refactored to avoid need for mdns-sd patch
 - Upgraded swift-rs removing need for patch as the fixes are now published
 - Drop the OpenSSL patch - I don't think we need it, if we do some upgrades to the Android build system. (done in #1336 )


core/Cargo.toml
```diff
@@ -53,56 +53,56 @@ tokio = { workspace = true, features = [
...
 enumflags2 = "0.7.7"
 http-range = "0.1.5"
-mini-moka = "0.10.0"
-serde_with = "2.3.3"
-dashmap = { version = "5.4.0", features = ["serde"] }
-notify = { version = "5.2.0", default-features = false, features = [
+mini-moka = "0.10.2"
+serde_with = "3.3.0"
+dashmap = { version = "5.5.3", features = ["serde"] }
+notify = { version = "6.1.1", default-features = false, features = [
```
fogodev, line 85: Notify had a breaking change on Linux if I'm not mistaken, how events are reported

oscartbeaumont: Will downgrade it then! Good spot!

brxken128 (dismissed):
Works great on MacOS from my anecdotal testing, and at least builds on Linux!

oscartbeaumont:
Android is broken on main and this PR makes it more broken (but broken is broken so I don't view this as a regression). Will fix on a follow up PR so it's easier to track the changes as we basically need to reinit the whole Android project.

utkubakir (approved):

PR #1315, opened 2023-09-08
Improve Location Formatting in UI

oscartbeaumont:
https://github.com/spacedriveapp/spacedrive/assets/21004798/916d0f04-3479-4fbc-95fb-49c61bd3db0b


brxken128 (approved):
Awesome feature to have, keeps the UI clean while still offering all information and even an easy way to actually see where the coordinates lead to. Nice job!

PR #1321, opened 2023-09-09
Twitter -> X

tomheaton:
**changes**

- `twitter.com` -> `x.com`
- Twitter SVG -> X SVG ([source](https://about.twitter.com/en/who-we-are/brand-toolkit))

**images**

image

image

**questions**
Do we want to rename Twitter to X (or to XTwitter (like `react-icons` does) for icons/components and object keys?


utkubakir:
Icon change is nice but doesn't Twitter still use `twitter.com`? And I like the XTwitter idea. Just X can be mistaken for well, a X lol

tomheaton:
`x.com` is just a redirect to `twitter.com` right now.

utkubakir (approved):

PR #1323, opened 2023-09-10
Update recommended VS Code extensions

tomheaton:
Closes #(issue)

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -3,7 +3,8 @@
 		"tauri-apps.tauri-vscode",
 		"rust-lang.rust-analyzer",
 		"oscartbeaumont.rspc-vscode",
-		"EditorConfig.EditorConfig",
-		"bradlc.vscode-tailwindcss"
+		"editorconfig.editorconfig",
+		"bradlc.vscode-tailwindcss",
+		"prisma.prisma"
 	]
 }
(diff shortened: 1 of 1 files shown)


.vscode/extensions.json
```diff
@@ -3,7 +3,8 @@
 		"tauri-apps.tauri-vscode",
 		"rust-lang.rust-analyzer",
 		"oscartbeaumont.rspc-vscode",
-		"EditorConfig.EditorConfig",
-		"bradlc.vscode-tailwindcss"
+		"editorconfig.editorconfig",
+		"bradlc.vscode-tailwindcss",
+		"prisma.prisma"
```
brxken128, line 8: This should be `Prisma.prisma`.

We could also add `esbenp.prettier-vscode` and `dbaeumer.vscode-eslint` I reckon, as they are both rather helpful and I'm pretty sure most of the team use them.

tomheaton: > This should be `Prisma.prisma`.
fun fact is it is not case sensitive

tomheaton: have added the others

brxken128: > > This should be `Prisma.prisma`.
> 
> fun fact is it is not case sensitive
> 
> 

well you changed the case of the editorconfig extension so i was just sticking with what changes you previously made

brxken128 (approved):

PR #1327, opened 2023-09-11
Make Windows setup for customizable

gaetschwartz:
This PR refactors the windows setup script so that it allows to easily choose not to use winget as well as clearer informations on what's about to be ran. Let me know if there's something you would like to be changed.


utkubakir:
We updated our setup script recently so I am not sure if this still applies @gaetschwartz. Can you take a look if there is anything we can use still?

gaetschwartz:
I'll close this for now and see if needed :)

HeavenVolkoff:
@gaetschwartz, I apologize for completely overlooking this. However, as @utkubakir mentioned, I recently refactored our setup scripts to separate the download and setup logic for our native dependencies into a JavaScript script. This change was made to enhance maintainability across all of our supported environments. 

Currently, the sole purpose of the setup script is to download essential system packages for development using the respective package manager for each OS (winget on Windows, apt/pacman/dnf/... on Linux, brew on macOS). If you prefer not to use winget, you now have the option to manually install the system dependencies and skip running the script altogether.

Additionally, I like your approach at organizing the installation of each dependency as tasks. If you'd like, you can open another pull request specifically for that, and I'd be more than happy to review it.

gaetschwartz:
No worries, I just pretty annoyed with the fact it would install llvm and rust via winget etc. Now all seems good ;)

PR #1329, opened 2023-09-11
[ENG-970] Job dropdown with actions

ameer2468:
Nice upgrade to the button in jobs. The button was not plugged in, now it is!

Screenshot 2023-09-11 at 4 22 42 PM


brxken128 (dismissed):
Can only get it to be transparent and not the same, even with a clean build. 

image

utkubakir (approved):

PR #1333, opened 2023-09-11
Update description

tomheaton:
Update all descriptions of Spacedrive being a "file manager" to "file explorer" to match hero title on landing.

I wasn't sure about the blog post so can revert that if you want.


utkubakir:
I personally prefer "file manager", any reason why you are going with explorer?

tomheaton:
Sorry, I meant to make this more clear.

The landing page says "A file explorer from the future."

jamiepine:
Ah, I guess I use both cause I can't decide what I like more. Manager is most accurate, explorer sounds cooler. We do both. For now I will let it simmer and see what people prefer.

PR #1335, opened 2023-09-12
[ENG-1097] DMS coordinate display support

brxken128:
This PR adds support for displaying coordinates in the DMS format. It converts them from lat/long in the frontend, and is configurable in the appearance settings page.

I have also added some disabled examples of other formats which could be configurable in the future, such as temperature and distance.

I tried to ensure that the DMS coordinates match up perfectly with Google, and every image I tested was a perfect match. I have also extended the DD display to 4 decimal places (with it being the full 8 on hover), as I feel as though it's both more expressive and in-line with everything else ~~(and odd numbers make me uncomfortable)~~.

I used the theme store in order to store this configuration information - I wasn't too sure where else it should go, but if anyone has any suggestions please let me know!

image

Screenshot 2023-09-12 at 03 04 32


utkubakir:
Instead of just DD or DMS, can we do with examples so the user can understand what they are.? like option a: XX.XX (DMS) option b: XX YY (DD)

brxken128:
> Same for Celsius options too, we can derive it from the user's environment when needed.

I somewhat disagree with this - e.g. an American living in the UK will likely be using a UK locale, and will be then shown celsius and KM, whereas they probably would prefer to see formats they're familiar with. I'll remove the configuration anyway though.

jamiepine:
I'd say keep the option, just update the defaults based on the user's system. I just foresee someone requesting this eventually anyway.

brxken128:
There's now a dedicated store for units, and we do offer changing them.

During onboarding, if `en-US` is detected as the locale then miles and fahrenheit will be automatically selected. I'm sure there's a whole list of rules for this (e.g. the UK use miles/mph for speed, KM for distance (but not on roads), grams for weighing things such as flour, stone and lbs for weighing people - it's a metrication failure, to say the least).

If anyone can enlighten me on their locale and what units they use, that'd be great. There are a few territories (primarily American) that use Fahrenheit still, but we'd have a hard time detecting them based on locale alone I believe so this will be good enough for now.

We can always add dynamic options too, like miles per hour/kilometres per hour depending on th distance unit selected.

image

ameer2468 (approved):

PR #1336, opened 2023-09-12
Fix Android bridge + drop IOS and Android projects

oscartbeaumont:
- Fix Android build config so OpenSSL is happy without being patched
 - Move Android to Rust bridge into an Expo Native Module - I had to reset the Android project so thought why not just do this.
 - Remove the entire Android project from Git to force us to make it reproducible
 - This will greatly help with maintenance and RN upgrades as we can wipe the project and start fresh.
 - Backport IOS `Podfile` for using Expo Native Module to build and link the Rust core from #1121  which will never be merged.
 - Rewrite most of the IOS bridge into Swift. Some small stuff is still Objective C but the core of it is now Swift as Expo Native Modules require that.
 - Drop the IOS project from the repo to force us to keep it reproducible.


oscartbeaumont:
@utkubakir Mobile's onboarding is completely broken. It hangs on creating a library. I am 99% sure the bridge is working correctly because the mutations `onSuccess` callback is being hit with the correct data.

oscartbeaumont:
It would seem we rely on the invalidation system for the navigation out of onboarding and that invalidation system was not working due to a mistake on my end. I am not a fan of that design but regardless problem should be fixed.

utkubakir (approved):

PR #1345, opened 2023-09-13
[ENG-1107] `file_path` unlinked from `object`

fogodev:
It was caused by a heisenbug on how we were joining files that should be removed in the walker. Now it takes 1 extra query but the results are being always correct in my tests


HeavenVolkoff (approved):
LGTM

PR #1349, opened 2023-09-13
[ENG-1076] Opening a sub dir closes the app

fogodev:
The issue was due to a bug found in the `futures_concurrency` crate, reported at https://github.com/yoshuawuyts/futures-concurrency/issues/155

For the time being, I changed all occurrences of `TryJoin` and `Join` from our codebase to use equivalents from `futures` crate, even though I dislike their ergonomics.

Also fixed some small bugs.


jamiepine (approved):
works a treat

PR #1350, opened 2023-09-14
Improve QuickPreview

oscartbeaumont:


apps/desktop/vite.config.ts
```diff
@@ -27,11 +27,12 @@ export default defineConfig(({ mode }) => {
...
 			devtoolsPlugin,
-			sentryVitePlugin({
-				authToken: process.env.SENTRY_AUTH_TOKEN,
-				org: 'spacedriveapp',
-				project: 'desktop'
-			})
+			mode === 'development' &&
+				sentryVitePlugin({
+					authToken: process.env.SENTRY_AUTH_TOKEN,
+					org: 'spacedriveapp',
```
brxken128, line 33: Should this not be limited to signed release builds only? That way we catch both true errors within the app, and not a bug someone introduced in order to (attempt to) waste our time.

core/src/custom_uri/serve_file.rs
```diff
@@ -0,0 +1,153 @@
...
+};
+use http_range::HttpRange;
+use tokio::{fs::File, io::AsyncSeekExt};
+use tokio_util::io::ReaderStream;
+use tracing::error;
+
+use super::{async_read_body::AsyncReadBody, utils::*};
+
+// default capacity 64KiB
+const DEFAULT_CAPACITY: usize = 65536;
```
brxken128, line 21: We can easily raise this if needs be, I know it can handle a lot more if needs be.

brxken128 (commented):
I can;t really comment on the TS but the Rust looks good!

apps/desktop/vite.config.ts
oscartbeaumont: Whoops yeah it would probs be only when `process.env.SENTRY_AUTH_TOKEN` is present. Without the check you need an internet connection for `pnpm run build` which is cringe.

interface/components/TextViewer/index.tsx
```diff
@@ -1,103 +1,154 @@
...
+		if (contentRef.current) {
+			const cb: IntersectionObserverCallback = (events) => {
+				for (const event of events) {
+					if (
+						!event.isIntersecting ||
+						contentRef.current?.getAttribute('data-highlighted') === 'true'
+					)
+						continue;
+					contentRef.current?.setAttribute('data-highlighted', 'true');
+					Prism.highlightElement(event.target, false); // Prism's async seems to be broken
```
HeavenVolkoff, line 114: Prism async works by just offloading the synchronous task to a background scripts, almost the same way we were doing before with `comlink`, however the way it expects to load its worker script is not compatible with our build system, that is why it doesn't work.

interface/components/TextViewer/worker.ts
```diff
@@ -1,45 +0,0 @@
...
-	const language = languageMapping.get(ext) ?? ext;
-	const grammar = Prism.languages[language];
-
-	return grammar
-		? {
-				code: Prism.highlight(code, grammar, language),
-				language
-		  }
-		: null;
-};
```
HeavenVolkoff, line 45: Does the text lazy loading works fine with synchronous Prism? I am asking because when I originally tested this processing just 100 lines of text synchronously through Prism already resulted in UI stutters in my MacBook, 1000+ lines would freeze the UI for a good 2 to 3 seconds. So, I am not sure if we should drop this, unless the lazy loading + Prism really fixed the hanging

core/src/custom_uri/serve_file.rs
oscartbeaumont: Possibly but it was taken from [tower-http](https://github.com/tower-rs/tower-http/blob/e8eb54966604ea7fa574a2a25e55232f5cfe675b/tower-http/src/services/fs/serve_dir/mod.rs#L27) so without benchmarking (also of different file types like text, pdf and video) I think we just leave it as is for now.

interface/components/TextViewer/worker.ts
oscartbeaumont: Our original system would highlight the entire document, now it only highlights rows that are visible on the screen so it's doing wayyyy less work.
I would be curious how it's performing for you locally but for me, it is working really well with large text files.

Brendonovich (approved):

PR #1353, opened 2023-09-14
[ENG-1060] Fix the `PlatformProvider` bug

oscartbeaumont:
This PR **needs** to be run by **many other poeple** to test it before merging. I have completed removed the `viteAliasResolver` with no replacement for it but it seems to be working for me on Windows and macOS.

## The issue
If you haven’t heard of the bug yet, in production builds, and especially on Windows, we keep running into the “PlatformProvider was not mounted” error when using the `usePlatform` hook.

## What I know
 - I managed to work out that the context 100% exists prior to React Router but after the router (so any of the child pages or layouts) the context is gone. This is fairly logical from the error message being "missing context" but it was less clear the router was the barrier where it's lost. This seems to be an issue for all contexts not just the `PlatformProvider`. It just happens to be the first one to be used.
 - Then I managed to work out if the `routes` array (defined in `@sd/interface`) can be moved from the file `app/index.tsx` to `index.tsx` then it started working.
 - Then I managed to work out that disabling the custom Vite alias resolver causes the bug to go away (~~and seemingly no new bugs to popup from it not being enabled~~ spoke too soon this broke web but we can bring back the resolver on web only).

Wait so to summarise you just said that if you enable our custom Vite `resolve.alias` for resolving files with Typescript aliases, the React context is unable to pass through the React Router, specifically on Windows and specifically in production builds. Yes, I sure did, does that make sense, no, is that how it's working yes.


oscartbeaumont:
It does look like I have broken web. Would seem it uses the Vite alias resolver but gonna try and fix that now.

oscartbeaumont:
I have tested this PR on Windows and macOS but the more testing the merrier.

I have brought back the Vite alias resolver only for sd-web. Idk how desktop works without it but i'm not gonna question it too hard as long as other people testing the app come back with it working, we are set to merge!

Brendonovich:
Seems to be working on my mac

HeavenVolkoff:
Just tested prod and dev on Windows and Linux, and it is working fine now. Great work @oscartbeaumont

oscartbeaumont:
Thanks, @Brendonovich @HeavenVolkoff for testing. Can we get this PR's code reviewed/approved then and we are ready to merge into main!

utkubakir (approved):

PR #1354, opened 2023-09-18
Increase ram for server release

Brendonovich:


utkubakir (approved):
prettier will convert them back to a single quote someday

PR #1359, opened 2023-09-19
[ENG-1125] Mouse navigation support

ameer2468:
This PR adds navigation support using forward and backward buttons on a mouse.


brxken128 (approved):
Nice work!

PR #1361, opened 2023-09-20
[ENG-927, ENG-735, ENG-766] Fix Updater & Tauri 1.5

Brendonovich:
- Uses ~~[this Tauri PR](https://github.com/tauri-apps/tauri/pull/7773)~~ Tauri 1.5
- Disables Tauri's updater dialog in favour of our own toasts and buttons
- Most version numbers removed in favour of `desktop/src-tauri/Cargo.toml`

To test updating, change the updater origin from `spacedrive.com` to `spacedrive-landing-git-eng-927-fix-updater-spacedrive.vercel.app`, and if you want to test in dev mode then remove the `&& import.meta.env.PROD` in `desktop/src/updater.ts`.

Tested on:
- [x] macOS aarch64
- [ ] macOS x86_64
- [x] linux x86_64
- [x] windows x86_64


apps/desktop/src-tauri/scripts/tauri.js
```diff
@@ -131,7 +131,7 @@ switch (args[0]) {
 }
 
 let code = 0;
-spawn('pnpm', ['tauri', ...args])
+spawn('pnpm', ['exec', 'tauri', ...args])
 	.catch((exitCode) => {
 		code = exitCode;
 		console.error(`tauri ${args[0]} failed with exit code ${exitCode}`);
```
utkubakir, line 137: Do we use this script anywhere?

apps/desktop/src/updater.ts
```diff
@@ -0,0 +1,82 @@
...
+				body: `Version ${update.version}`
+			};
+		},
+		{
+			onClose() {
+				onInstallCallbacks.delete(cb);
+			},
+			duration: 10 * 1000,
+			action: {
+				label: 'Install',
```
utkubakir, line 68: "Update" sounds better and more correct imo

interface/app/$libraryId/Layout/Sidebar/Footer.tsx
```diff
@@ -20,8 +22,24 @@ export default () => {
...
 		<div className="space-y-2">
+			{updater && updaterState && (
+				<>
+					{updaterState.status === 'updateAvailable' && (
+						<Button
+							variant="outline"
+							className="w-full"
+							onClick={updater.installUpdate}
+						>
+							Install Update
```
utkubakir, line 38: just "Update"

apps/desktop/src/updater.ts
Brendonovich: Idk if I agree, saying 'New Update Available' and then also using the word Update as the action seems weird

utkubakir: I don't think it is. Install to me means, installing something new. Update literally means;

> an act of updating something or someone or an updated version of something.

Brendonovich: Meaning-wise it's fine, but using it as both copy and as an action verb feels weird

utkubakir: I am personally very used to seeing it so it doesn't come as weird to me. "Download & Install Update" is also common but if it's just 1 word we are looking for, it should be "update" here.

Brendonovich: Idk if you're understanding what I mean but I'll change it to Update

utkubakir (approved):
Not tested but it lgtm

apps/desktop/src/updater.ts
oscartbeaumont: For what it's worth, I agree with Brendan on this. Could be worth asking a few others to see what they think.

I feel like the toast represents the update (via the title) and the action your doing to the update to "install" the update. You're not updating the update.

oscartbeaumont: but then again either way doesn't really matter. It's pretty obvious what it does.

oscartbeaumont: image
image

Prior art seems to go either way.

.cargo/config.toml.mustache
```diff
@@ -8,10 +8,10 @@ FFMPEG_DIR = "{{{ffmpeg}}}"
 
 {{#isMacOS}}
 [target.x86_64-apple-darwin]
-rustflags = ["-L", "{{{projectRoot}}}/target/Frameworks/lib"]
+rustflags = ["-L", "{{{projectRoot}}}/target/CustomFrameworks/lib"]
```
HeavenVolkoff, line 11: Why the change from Framework to CustomFramework?

Brendonovich: Tauri uses the `Frameworks` library itself, so i figured we'd just use `CustomFrameworks`. Can't remember whether Tauri does anything destructive with the `Frameworks` folder but better to avoid collisions anyway I reckon.

HeavenVolkoff: The thing is that I think changing this will impact dev builds. That is because the `../Frameworks/*.framework` path, from the binary, is a know structure for `.app` bundles, and when I did it this way in the target folder it was to replicate the same structure for dev, as the binary will be put in `target/debug/spacedrive` and when it lunch it will be able to locate all the shared libraries the same way that it would do in prod. If you run `otool -L target/debug/Spacedrive` it will show that our dev build use `@executable_path/../Frameworks/FFMpeg.framework/...` for all of our custom `dylibs`, also the `FFMpeg.framework` was build expecting this specific structure:

https://github.com/spacedriveapp/spacedrive/blob/b62ca655a0110fb73fbc829ada1a636d2d6b8828/.github/scripts/ffmpeg-macos/build.sh#L373

Brendonovich: Ah I see, I"ll look into this

HeavenVolkoff (dismissed):

brxken128 (approved):
Tested on Win11, Void Linux, Fedora 38, and MacOS 13.5 - works great.

PR #1362, opened 2023-09-20
[ENG-812] Folder size calculation

### LINKED ISSUE #1149: [ENG-812] Folder size calculation
Conversation about the issue below

<sub>From [SyncLinear.com](https://synclinear.com) | [ENG-812](https://linear.app/spacedriveapp/issue/ENG-812/folder-size-calculation)</sub>

fogodev:
This PR presents a way to calculate directory sizes for all directories in an indexed location in a really lightweight manner, using data about each file size already extracted in the walker for indexer job and shallow indexer. It also takes into account size updates in a directory for files being added through the watcher. 

In case of new or updated files found through indexer jobs or shallow indexers that were using a sub path in a given location, we back propagate the updated size calculations back to ancestor directories.

Now we also dispatch sub path scans for new directories found through the shallow indexer.

Closes #1149

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -77,5 +77,6 @@ dev.db-journal
 /core/migration_test
 sd_init.json
 
+spacedrive
 .cargo/config
 .github/scripts/deps
--- a/core/src/location/file_path_helper/mod.rs
+++ b/core/src/location/file_path_helper/mod.rs
@@ -80,6 +80,7 @@ file_path::select!(file_path_walker {
 	date_modified
 	inode
 	device
+	size_in_bytes_bytes
 });
 file_path::select!(file_path_to_handle_custom_uri {
 	pub_id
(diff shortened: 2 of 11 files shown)


Brendonovich (approved):
This works from my testing, but doesn't use the approach discussed in the issue. This is probably fine and we can move to a different approach if needed.

fogodev:
> This works from my testing, but doesn't use the approach discussed in the issue. This is probably fine and we can move to a different approach if needed.

Yeah, I preferred to use a different approach as I managed to find a very lightweight manner to implement the feature, just reusing data already present in the walker. The computational costs from my approach are just some extra data structures to hold directories size data and some extra DB queries to update each directory size. The mitigations discussed in the issue weren't needed as I haven't introduced any extra cost traversing the FS again just to calculate directories sizes.

Brendonovich:
From my perspective the issue isn't that calculating the folder sizes could be expensive, it's that we'd be generating sync messages for folder sizes when we could be calculating them on each node instead

PR #1363, opened 2023-09-21
Done

fogodev:
Reverting [ENG-941]

Fixing the issue with the job is active animation being shown for jobs from another library.

Running pnpm format


Brendonovich (approved):
Works for me :)

PR #1364, opened 2023-09-21
[ENG-1096] Image Conversions

brxken128:
This implements the API to allow for users to convert between different supported image formats.

The UI is not in place, but that should be a quick alteration of the context menu and a dialog with all of the options (including a quality percentage between 1-125 (showing yellow when the progress bar > 100)). I can open another issue and do this in a later PR.

This also does not check if the user is converting to the same type, but they may be rescaling so it's a valid use-case.


core/src/api/files.rs
```diff
@@ -317,6 +321,82 @@ pub(crate) fn mount() -> AlphaRouter<Ctx> {
 					}
 				})
 		})
+		.procedure("convertImage", {
+			#[derive(Type, Deserialize)]
+			struct ConvertImageArgs {
+				path: PathBuf,
+				delete_src: bool, // if set, we delete the src image after
+				desired_extension: String,
```
Brendonovich, line 329: Can this be typesafe?

brxken128: Not unless we make and maintain a gigantic enum, as the current `Extensions::ImageExtension` enum doesn't match up with what we support. It's matched in the `sd-images` crate, although I suppose I could try to macro together an enum.

Brendonovich: i'm fine with big enums 🤷

brxken128: I will make one then 🥱

fogodev (approved):

PR #1366, opened 2023-09-21
[ENG-1135] Locations: Total size query and return to Frontend

fogodev:
Now that we have accurate directories calculation for an entire location, it's pretty easy to have a full location size.

### DIFF
--- a/core/prisma/migrations/20230921203938_add_size_bytes_in_location_table/migration.sql
+++ b/core/prisma/migrations/20230921203938_add_size_bytes_in_location_table/migration.sql
@@ -1,0 +1,2 @@
+-- AlterTable
+ALTER TABLE "location" ADD COLUMN "size_in_bytes" BLOB;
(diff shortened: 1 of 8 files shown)


ameer2468 (approved):

Brendonovich (changes requested):
For directories I don't mind storing location size in the db too much, but doing the same for locations doesn't sit right with me seeing as they previously didn't have a size column. I'd much prefer if location size was calculated on the fly and potentially cached in-memory.

ameer2468:
Frontend LGTM

utkubakir:
> For directories I don't mind storing location size in the db too much, but doing the same for locations doesn't sit right with me seeing as they previously didn't have a size column. I'd much prefer if location size was calculated on the fly and potentially cached in-memory.

I think caching it in DB also isn't too bad, it's not a cheap query and to get let's say 100 location's total storage on the fly (lets say first launch), seems a bit too much.

Brendonovich (approved):

PR #1368, opened 2023-09-22
[ENG-265] Improve setup scripts

HeavenVolkoff:
Complete revamp of the setup script for all platforms.

Tested on Windows 11, Arch Linux and macOS 13. More tests are welcome

Changes:

 - Remove all logic for download external dependencies from the `setup-system` scripts, now they only have a single purpose, which is to bootstrap a new user environment. A.K.A no more re-running `setup-system.sh`
 - Introduce a postinstall step for `pnpm i` that handles downloading and configuring all of our shared lib dependencies (FFMpeg, protoc, pdfium, libheif, etc...), completly written in JS and multiplatform. This will greatly reduce the complexibility of mantaining and adding new shared libraries. No more cursed bash and powershell XD
 - Add a caching system for request to the github API, this should solve the Rate Limit problem some people were facing
 - Fix the duplicated cargo config warnings, now we have a `.cargo/config.toml.mustache` template
 - Rewrote the setup-system.sh script from scratch
 - Adds new shared lib dependency: pdfium (For PDF thumbnails)


Brendonovich (approved):
Works great on mac, needs more testing tho

Brendonovich (approved):

utkubakir (approved):
Can confirm script works now

PR #1378, opened 2023-09-26
Desktop + Web auth flow

Brendonovich:
~~_shouldn't be merged until https://github.com/spacedriveapp/api/pull/3 is_~~

image


apps/web/src/App.tsx
```diff
@@ -42,7 +42,15 @@ const platform: Platform = {
 			locationLocalId
 		)}/${encodeURIComponent(filePathId)}`,
 	openLink: (url) => window.open(url, '_blank')?.focus(),
-	confirm: (message, cb) => cb(window.confirm(message))
+	confirm: (message, cb) => cb(window.confirm(message)),
+	auth: {
+		start(key) {
+			return window.open(`http://localhost:3000/device-session?key=${key}`);
```
oscartbeaumont, line 48: We should use `spacedriveURL` like the other things do?

They assume the server and the UI are on the same port and I think we should keep it consistent but we probally do also need to allow for someone to override the port using an env var or something and we can set it in our `pnpm web dev` script or whatever given it does run the server and UI on different ports which I assume is why this is hardcoded.

core/src/api/auth.rs
```diff
@@ -0,0 +1,84 @@
...
+				Token(String),
+			}
+
+			R.subscription(|_, _: ()| async move {
+				let device_session_url = format!(
+					"{}/api/auth/device-session",
+					std::env::var("SD_API_URL").unwrap()
+				);
+
+				let client = reqwest::Client::new();
```
oscartbeaumont, line 26: This is such a nitpick but we should probs just capture this into the rspc handler so it's reused for all subscriptions.

oscartbeaumont (changes requested):
Love the usage of feature flags.

Clippy warning should probably also be addressed before merging.

apps/web/src/App.tsx
Brendonovich: url is provided by `verification_url_complete`

oscartbeaumont (approved):

oscartbeaumont (approved):

PR #1386, opened 2023-09-26
[ENG-1140] Bind an entire location to a single device number

fogodev:
Removing device data from file_path as it can change overtime for a location mounted over SMB for example


jamiepine:
IT WORKS, I AM HAPPY

jamiepine (approved):

PR #1388, opened 2023-09-26
Fix mobile core not building on non-darwin machines

### LINKED ISSUE #1387: Can't run Android app on Linux/Windows
### Describe the bug

Gradle build fails when attempting to run the app on Windows and Linux due to a missing `llvm-ranlib` file.

### Reproduction

1. Complete required steps mentioned in contributing.md (e.g. run system-setup.sh mobile, pnpm i, pnpm prep, etc.)
2. run the android command for Expo (e.g. `pnpm mobile android` from root)
3. Wait for gradle build to fail due to missing `llvm-ranlib`

### Expected behavior

Successfully finish building and load the app.

### Platform and versions

```Shell
% pnpm --version && cargo --version && rustc --version

8.7.6
cargo 1.71.0 (cfd3bbd8f 2023-06-08)
rustc 1.71.0 (8ede3aae2 2023-07-12)
```

### Stack trace

_No response_

### Additional context

_No response_

gedeondoescode:
Changed `build.gradle` in the android folder to properly find `llvm-ranlib` for Linux and Windows machines. Windows has not been tested yet. 

Closes #1387

### DIFF
--- a/apps/mobile/modules/sd-core/android/build.gradle
+++ b/apps/mobile/modules/sd-core/android/build.gradle
@@ -112,7 +112,7 @@ cargo {
     targetDirectory = "../../../../../target" // Monorepo moment
 		exec { spec, toolchain ->
 
-			def dir = "${android.ndkDirectory}/toolchains/llvm/prebuilt/${osName}-x86_64/bin/llvm-ranlib"
+			def dir = "${android.ndkDirectory}/toolchains/llvm/prebuilt/${osName.contains("mac") ? "darwin" : osName}-x86_64/bin/llvm-ranlib"
 
 			spec.environment("RANLIB_armv7-linux-androideabi", "${dir}")
 			spec.environment("RANLIB_aarch64-linux-android", "${dir}")
--- a/apps/mobile/modules/sd-core/android/build.gradle
+++ b/apps/mobile/modules/sd-core/android/build.gradle
@@ -110,36 +110,14 @@ cargo {
     profile = 'release' // 'debug'
     targets = cargoTargets
     targetDirectory = "../../../../../target" // Monorepo moment
-		if (osName.contains("linux")) {
-			exec { spec, toolchain ->
+		exec { spec, toolchain ->
 
-				def dir = "${android.ndkDirectory}/toolchains/llvm/prebuilt/linux-x86_64/bin/llvm-ranlib"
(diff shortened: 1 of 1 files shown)


apps/mobile/modules/sd-core/android/build.gradle
```diff
@@ -101,21 +102,45 @@ if (System.getenv('SPACEDRIVE_CI') != "1") {
...
+	  final String osName = System.getProperty("os.name").toLowerCase();
+
     module  = "./crate"
     libname = "sd_mobile_android"
     pythonCommand = 'python3'
     profile = 'release' // 'debug'
     targets = cargoTargets
     targetDirectory = "../../../../../target" // Monorepo moment
+		if (osName.contains("linux")) {
+			exec { spec, toolchain ->
```
oscartbeaumont, line 114: Can we not have a single `exec` block and then use `.../prebuilt/${osName}-x86_64/bin/...` for the `dir`?

I am not super familiar with Gradle so could definitely not work but that seems like basically what we already do with `android.ndkDirectory` and it would be a heap neater.

For personal note: Checking my local system (M1 mac) there is only an `x86_64` dir so it's expected for that to be hardcoded.

oscartbeaumont (changes requested):
My bad for letting this bug get merged 🤦

apps/mobile/modules/sd-core/android/build.gradle
gedeondoescode: I'll try and see if that's possible. I mostly did it this way since I wasn't sure how to include the architecture in the most efficient way but since they're all the same I can get that done quick

oscartbeaumont: All good! If you don't have time I will give it a shot when I have a few minutes.

gedeondoescode: I'll do it when I wake up in the morning. I wanna try and see this through on my own.

utkubakir:
@oscartbeaumont does this fix the issue I am having on Mac or is it unrelated?

oscartbeaumont:
Honestly, no idea.

gedeondoescode:
@utkubakir if you get the chance, can you try it on your machine and see if anything's different?

PR #1390, opened 2023-09-27
[ENG-1143] Media View sort by date taken

brxken128:
This uses the unix epoch time, it's stored as a `BigInt?`. We can successfully ascendingly and descendingly order by the true date an image was taken.

The `BigInt` is retrieved on-demand, and never go anywhere near RSPC so we are more than okay in that regard.

I cleaned up some other code and made a start on fixing the date displayed in the UI. Keeping this as a draft until specta, rspc and dayjs learn to play nicely.


brxken128:
Couldn't get `dayjs` to work so we just show the raw UTC/Naive time in the UI instead.

jamiepine (dismissed):
good to go

Brendonovich (approved):

PR #1392, opened 2023-09-27
[ENG-1145] Keep more info open if it is

ameer2468:
This keeps 'More Info' open if it is toggled.

### DIFF
--- a/interface/components/Accordion.tsx
+++ b/interface/components/Accordion.tsx
@@ -7,6 +7,12 @@ interface Props {
 	title: string;
 	variant?: keyof typeof styles;
 	className?: string;
+	//valtio store to keep the toggle state
+	valtio?: {
+		getStore: () => { [key: string]: any };
+		store: { [key: string]: any };
+		stateKey: string;
+	};
 }
 
 const styles = {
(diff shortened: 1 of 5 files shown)


brxken128 (approved):
I think this should remain open if you change directory or location with the inspector open, but on app-launch it should be disabled by default. Just my thoughts, rest looks good!

IIrc @jamiepine did say it should close on a certain condition, but I can't remember it off the top of my head (turned out to be the current path).

interface/components/Accordion.tsx
```diff
@@ -7,6 +7,12 @@ interface Props {
 	title: string;
 	variant?: keyof typeof styles;
 	className?: string;
+	//valtio store to keep the toggle state
+	valtio?: {
+		getStore: () => { [key: string]: any };
+		store: { [key: string]: any };
+		stateKey: string;
+	};
```
utkubakir, line 15: I don't like this, why even have valtio here? It's not the accordion's job to persist it. You could pass `isOpen` and `onToggle` to programmatically control it on the MediaData component.

utkubakir (changes requested):

brxken128 (dismissed):

utkubakir (dismissed):

brxken128 (approved):

PR #1394, opened 2023-09-28
[ENG-1081] Ignore the case of extensions when matching

fogodev:
We can't force extensions to always be lowercased in the database as this results in an information loss. As the original extension can have uppercased characters and we will fail to match the file name for many purposes.

This PR solves that introducing a lowercase comparison in the sd-file-ext sub crate and also a query that handles lowercasing inside the database. Previously I tried to set the extension column to be case insentitive. But this is only applied when the migrations run, which is exclusively in prod builds. So dev builds are totally broken in this regard. This PR handles it for good.

The only caveat was the PCR doesn't support raw queries with vec parameters for SQLite for some reason, so I had to interpolate the query string myself. It was ok in this case as extensions strings are hardcoded in our sd-file-ext subcrate, not coming from the user. The other params are using `PrismaValue` to be interpolated as they should.


brxken128 (commented):
LGTM but I'll leave this one for @Brendonovich to fully approve to ensure there are no adverse effects.

Brendonovich (approved):

PR #1399, opened 2023-09-28
settings pages: consistent padding on every side

PineappleRind:
- Add consistent bottom padding to each settings page (make bottom padding block div unshrinkable)
- Remove extra padding on Library -> General page
- Remove extra bottom padding on Resources -> About page

### DIFF
--- a/interface/app/$libraryId/settings/resources/about.tsx
+++ b/interface/app/$libraryId/settings/resources/about.tsx
@@ -81,7 +81,7 @@ export const Component = () => {
 				</p>
 			</div>
 			<Divider />
-			<div className="mb-20">
+			<div>
 				<h1 className="my-5 text-lg font-bold text-ink">
 					We also would like to thank all our contributors
 				</h1>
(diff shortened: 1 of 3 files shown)


ameer2468:
Hey Pineapple thanks for the contribution!

These changes look great, however, we don't want to remove mb-20 completely, rather change it to mb-12 instead. This makes the list of contributers directly stuck to the bottom without any spacing, and this isn't something we visually want.

- Ameer

interface/app/$libraryId/settings/resources/about.tsx
```diff
@@ -81,7 +81,7 @@ export const Component = () => {
 				</p>
 			</div>
 			<Divider />
-			<div className="mb-20">
+			<div>
```
ameer2468, line 84: Hey Pineapple thanks for the contribution!

These changes look great, however, we don't want to remove mb-20 completely, rather change it to pb-12 instead. Without any bottom spacing, this makes the list of contributers directly stuck to the bottom, and this isn't something we visually want.

- Ameer

PineappleRind: now that there's an element that provides bottom padding on every settings page we don't need the margin set on the About page
image

ameer2468: > now that there's an element that provides bottom padding on every settings page we don't need the margin set on the About page image

Great - please make sure its 12 rather than 20. I think 20 is too much.

PineappleRind: image
Done 👍

ameer2468 (approved):

PR #1402, opened 2023-09-29
[ENG-1117] Remove Tauri drag regions on non-macOS platforms

oscartbeaumont:
Windows and Linux have a top bar Mac does not hence the need for the drag regions.


utkubakir (approved):
lgtm

PR #1403, opened 2023-09-29
[ENG-1075] Light mode for QuickPreview

oscartbeaumont:


utkubakir:
Line numbers need to be more subtle imo

utkubakir (approved):

PR #1406, opened 2023-09-29
[ENG-1149]  Tags unassignment UI update and fixes

ameer2468:
Thanks to @niikeec for being involved here **[ENG-1118]** and providing a working solution for context menu bug. Radix does not offer a way to enforce a close, hence the solution.

**This PR fixes multiple minor issues**:

- Shortcut of views routed to meta + b, instead of meta + v, because that is for paste
- Expand in jobs should only show for grouped jobs, not single
- Un-assigning a tag within tag page updates the UI immediately. **[relates to ENG-1118 bug fix]**
- Made path component only render in locations, no other page. (i.e tags)
- Tooltip for long names adjusted in Job manager, it will break-all like file names to see whole name.
- Loading prop added for Empty notice component, so we don't see it for a split second before data loads. Did this for tags page.


niikeec (approved):

utkubakir:
Did you also add these shortcuts to the shortcuts page?

ameer2468:
> Did you also add these shortcuts to the shortcuts page?

Yes

PR #1409, opened 2023-09-29
[ENG-1165] Add a security policy and notices to the appropriate READMEs

brxken128:
Self explanatory. We could probably sign these, but the `security.txt` file is signed which should validate all of the information displayed anyway.


utkubakir:
I don't think we have access to pgp private key though

Brendonovich (approved):
Private key is in BitWarden

PR #1417, opened 2023-10-02
[ENG-1170] - Prevent sidebar scrolling with keyboard

ameer2468:
When navigating with keyboard, this prevents sidebar scrolling from happening. **This has been tested to ensure that it doesn't affect other functionalities.**

- By default all overflowing elements have scrolling support with keyboard by the browsers, we need to prevent that by targeting the up and down arrow keys.


utkubakir (approved):
LGTM

PR #1419, opened 2023-10-02
[ENG-1172] - Update navigation keys

ameer2468:
This makes you able to navigate backwards and forwards using Meta + [ and Meta + ] - at the sametime we also have support with arrow keys, as most apps **(seems to be native)**.


utkubakir:
Can you also add this and all the new shortcuts to the Shortcuts page?

utkubakir (approved):

PR #1422, opened 2023-10-03
[ENG-871] - Rename shortcut

ameer2468:
- For windows, you can now press F2 to rename, this also includes on Mac (FN + F2) and also Enter for Mac.
- Keybindings slightly refactored to accomodate our needs, you can now specificy what OS to target for a specific shortcut or keys for 'all' regardless of the OS.

**Notes**: 

- Enter will be re-routed to open a folder while on Mac CMD + O on a separate PR.

### DIFF
--- a/interface/app/$libraryId/settings/client/keybindings.tsx
+++ b/interface/app/$libraryId/settings/client/keybindings.tsx
@@ -17,7 +17,12 @@ import Setting from '../Setting';
 type Shortcut = {
 	action: string;
 	description?: string;
-	key: [ModifierKeys[], string[]][];
+	key: {
+		[K in OperatingSystem | 'all']?: {
+			value: string | ModifierKeys | (string | ModifierKeys)[];
+			split?: boolean;
+		};
+	};
 };
 
 const shortcutCategories: Record<string, Shortcut[]> = {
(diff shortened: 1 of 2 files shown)


interface/app/$libraryId/settings/client/keybindings.tsx
```diff
@@ -17,80 +17,128 @@ import Setting from '../Setting';
...
+			}
 		},
 		{
 			action: 'Open selected item',
-			key: [[[ModifierKeys.Control], ['ArrowUp']]]
+			key: {
+				all: {
+					value: [ModifierKeys.Control, 'ArrowUp']
+				}
+			}
```
utkubakir, line 91: Command + O

interface/app/$libraryId/settings/client/keybindings.tsx
```diff
@@ -17,80 +17,128 @@ import Setting from '../Setting';
 type Shortcut = {
 	action: string;
 	description?: string;
-	key: [ModifierKeys[], string[]][];
+	key: {
```
utkubakir, line 20: "keys" also can you include some documentation here?

interface/app/$libraryId/settings/client/keybindings.tsx
ameer2468: Theres a ticket open for that - so im doing the change in there.

interface/app/$libraryId/settings/client/keybindings.tsx
```diff
@@ -17,80 +17,128 @@ import Setting from '../Setting';
...
 		},
 		{
 			action: 'Show path bar',
-			key: [[[ModifierKeys.Control], ['p']]]
+			key: {
+				all: {
+					value: [ModifierKeys.Control, 'p']
+				}
+			}
 		},
```
utkubakir, line 108: They should also work with capital letters. Not sure if that's the case already?

PR #1425, opened 2023-10-03
[ENG-1176] Append `(x)` to the end of files instead of overwriting

brxken128:
During: copying, cutting, duplicating and renaming we now will append the lowest available digit to the file name if it would overwrite the source file.

The only "bug" I have ran into is when renaming say `File (2).jpg` to `File.jpg`, it will increment the name to `File (3).jpg` when it should in fact remain as `File (2).jpg` (as that's the largest available number, but it already exists so it gets incremented.

A preview:
an image showing how the new names look including the `(x)`

This may need some pretty heavy battle-testing before merge, as it could end up truncating files if it goes wrong (it shouldn't, as the action should only ever happen if the `fs::metadata` call is an `Err` (the file doesn't exist).

### DIFF
--- a/core/src/api/files.rs
+++ b/core/src/api/files.rs
@@ -11,8 +11,8 @@ use crate::{
 	},
 	object::{
 		fs::{
-			append_digit_to_filename, copy::FileCopierJobInit, cut::FileCutterJobInit,
-			delete::FileDeleterJobInit, erase::FileEraserJobInit,
+			copy::FileCopierJobInit, cut::FileCutterJobInit, delete::FileDeleterJobInit,
+			erase::FileEraserJobInit,
 		},
 		media::{
 			media_data_extractor::{
--- a/core/src/api/files.rs
+++ b/core/src/api/files.rs
@@ -428,12 +428,12 @@ pub(crate) fn mount() -> AlphaRouter<Ctx> {
 
 					match fs::metadata(&new_file_full_path).await {
 						Ok(_) => {
-							return Err(rspc::Error::with_cause(
-									ErrorCode::InternalServerError,
-									"Renaming would overwrite a file".to_string(),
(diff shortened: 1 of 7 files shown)


utkubakir:
Copy/Cut/Paste/Duplicate all sounds good but I don't think it makes sense while renaming. It should error just like Finder. This also fixes the bug you encountered so win-win :)

utkubakir (changes requested):
Copy/Cut/Paste/Duplicate all sounds good but I don't think it makes sense for renaming. It should error out just like Finder. This also fixes the bug you encountered so win-win :)

brxken128:
> Copy/Cut/Paste/Duplicate all sounds good but I don't think it makes sense for renaming. It should error out just like Finder. This also fixes the bug you encountered so win-win :)

Should be resolved as of my latest few commits! :D

fogodev:
Solved some small tech debts like unneeded clone and other nitpicks. Also found a bug with the renaming function which was having trouble with spacing like duplicating a `file (1).txt` generated a `file  (1).txt` here due. Also the solution wasn't taking into account numbers with more than 1 digit, as it would generate bad naming due to hardcoded slice with `len() - 3`

fogodev (approved):

fogodev (approved):

brxken128:
> Solved some small tech debts like unneeded clone and other nitpicks. Also found a bug with the renaming function which was having trouble with spacing like duplicating a `file (1).txt` generated a `file (1).txt` here due to this. Also the solution wasn't taking into account numbers with more than 1 digit, as it would generate bad naming due to hardcoded slice with `len() - 3`

Oh yeah that was a very good catch, thank you!

utkubakir (approved):

PR #1429, opened 2023-10-04
[ENG-1175] File named " " in Finder renders name incorrectly

oscartbeaumont:
Changes:
 - [x] Fix styling of " " file name
 - [x] Don't trim when renaming

Potentially also fixes ENG-1137 (although was still unable to reproduce it).


interface/app/$libraryId/Explorer/FilePath/RenameTextBox.tsx
```diff
@@ -72,9 +72,9 @@ export const RenameTextBoxBase = forwardRef<HTMLDivElement | null, Props>(
 		const reset = () => ref.current && (ref.current.innerText = text ?? '');
 
 		const handleRename = async () => {
-			const newName = ref.current?.innerText.trim();
-
-			if (!newName || newName === text) {
+			let newName = ref.current?.innerText;
+			if (newName?.endsWith('\n')) newName = newName.slice(0, -1);
+			if (newName === undefined || newName === '') {
```
HeavenVolkoff, line 77: `ref.current` can be null, which would propagate to `newName` due to the `.?`, better to change `newName === undefined` to `newName == null` or simply do `if (!newName)` as that will catch all falsy values

HeavenVolkoff (changes requested):

interface/app/$libraryId/Explorer/FilePath/RenameTextBox.tsx
```diff
@@ -72,9 +72,9 @@ export const RenameTextBoxBase = forwardRef<HTMLDivElement | null, Props>(
 		const reset = () => ref.current && (ref.current.innerText = text ?? '');
 
 		const handleRename = async () => {
-			const newName = ref.current?.innerText.trim();
-
-			if (!newName || newName === text) {
+			let newName = ref.current?.innerText;
+			if (newName?.endsWith('\n')) newName = newName.slice(0, -1);
+			if (!newName || newName === '') {
```
niikeec, line 77: `!newName` should be enough here

niikeec (changes requested):

oscartbeaumont:
@niikeec said these change was copied onto another PR so closing this.

PR #1430, opened 2023-10-04
[ENG-982] Can't load original after renaming

oscartbeaumont:
LRU cache was never invalidated and it broke.


brxken128:
I debated over applying or allowing the clippy warning, but considering this seems geared towards strictly the invalidation system I opted for applying the lint, if that was the wrong move feel free to revert!

brxken128 (approved):

oscartbeaumont:
Thanks for updating it! Clippy is always right. I just completely forgot to run it. I only added the `match` so LSP could autofill the variants cause I forgor the original API, then forgot to remove it.

PR #1431, opened 2023-10-04
[ENG-1182] Switch quick preview animation from react-spring to tailwindcss-animate

niikeec:


interface/app/$libraryId/Explorer/QuickPreview/index.tsx
```diff
@@ -181,268 +166,253 @@ export const QuickPreview = () => {
...
-									'absolute inset-0 z-50',
-									isDark ? 'bg-black/80' : 'bg-black/60'
+	return (
+		<Dialog.Root open={open} onOpenChange={(open) => (getQuickPreviewStore().open = open)}>
+			<QuickPreviewContext.Provider value={{ background }}>
+				<Dialog.Portal forceMount>
+					<Dialog.Overlay
+						className={clsx(
+							'absolute inset-0 z-50',
+							'data-[state=open]:animate-in data-[state=open]:fade-in-0',
```
Brendonovich, line 185: I think we can use utilities from `tailwindcss-radix`, same goes further down:

```suggestion
							'radix-state-open:animate-in radix-state-open:fade-in-0',
```

Brendonovich (changes requested):

interface/app/$libraryId/Explorer/QuickPreview/index.tsx
niikeec: True we could, will do that

Brendonovich (approved):

PR #1434, opened 2023-10-05
Centralise auth state into context

Brendonovich:
Lifts all auth state up to the root, so login attempts don't get cancelled if you navigate somewhere else 🤦‍♂️


utkubakir (dismissed):

Brendonovich:
i'm dumb i should have used a store for this

ameer2468 (approved):

PR #1444, opened 2023-10-08
[ENG-1196] Allow `WEBUI_PORT` to be specified for host-networked containers

brxken128:
This also changes the Dockerfile to follow some more of the best practices (it's missing a few but it's a lot better, although I may have broke something along the way).

I also cleaned up the other CI files, as `trunk` was emitting quite a few lint warnings - I'm not sure what effects my changes will have, we will have to see.

**Please don't merge or approve this yet, it's experimental but not a draft so CI still runs**

### DIFF
--- a/apps/server/docker/Dockerfile
+++ b/apps/server/docker/Dockerfile
@@ -8,14 +8,19 @@ FROM debian:bookworm as base
 
 ARG DEBIAN_FRONTEND
 
-ADD https://gist.githubusercontent.com/HeavenVolkoff/ff7b77b9087f956b8df944772e93c071/raw \
+COPY https://gist.githubusercontent.com/HeavenVolkoff/ff7b77b9087f956b8df944772e93c071/raw \
 	/etc/apt/apt.conf.d/99docker-apt-config
 
+USER spacedrive
+HEALTHCHECK NONE
+
+# trunk-ignore(terrascan/AC_DOCKER_0002)
 RUN rm -f /etc/apt/apt.conf.d/docker-clean; \
 	echo 'Binary::apt::APT::Keep-Downloaded-Packages "true";' > /etc/apt/apt.conf.d/keep-cache
 
-RUN --mount=type=cache,target=/var/cache/apt --mount=type=cache,target=/var/lib/apt \
-	apt-get update && apt-get upgrade
+# RUN --mount=type=cache,target=/var/cache/apt --mount=type=cache,target=/var/lib/apt \
+# 	apt-get update && apt-get upgrade
+
 
 #--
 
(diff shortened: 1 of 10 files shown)


apps/server/docker/entrypoint.sh
```diff
@@ -26,11 +26,11 @@ adduser --system --disabled-password \
   spacedrive
 passwd -l spacedrive
 
-if [ -n "${TZ:-}" ]; then
```
HeavenVolkoff, line 29: The colon here is required, as it means substitute when unset or null, without a color it only substitutes when unset.
https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_02
image

apps/server/docker/entrypoint.sh
```diff
@@ -5,9 +5,9 @@ set -eu
 # Shortcircuit for non-default commands.
 # The last part inside the "{}" is a workaround for the following bug in ash/dash:
 # https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=874264
-if [ -n "${1:-}" ] && [ "${1#-}" = "${1}" ] \
```
HeavenVolkoff, line 8: The colon here is required, as it means substitute when unset or null, without a color it only substitutes when unset.

HeavenVolkoff (changes requested):

apps/server/docker/Dockerfile
```diff
@@ -60,9 +70,11 @@ RUN pnpm web build
 
 FROM build-base as server
 
-RUN wget -qO- https://sh.rustup.rs | sh -s -- -yq --profile minimal
+# trunk-ignore(hadolint/DL4006)
+RUN ["/bin/bash", "-c", "set -o pipefail && wget -qO- https://sh.rustup.rs | sh -s -- -yq --profile minimal"]
```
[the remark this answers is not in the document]
brxken128: https://docs.docker.com/develop/develop-images/instructions/#using-pipes

apps/server/docker/Dockerfile
```diff
@@ -8,25 +8,35 @@ FROM debian:bookworm as base
 
 ARG DEBIAN_FRONTEND
 
-ADD https://gist.githubusercontent.com/HeavenVolkoff/ff7b77b9087f956b8df944772e93c071/raw \
+COPY https://gist.githubusercontent.com/HeavenVolkoff/ff7b77b9087f956b8df944772e93c071/raw \
 	/etc/apt/apt.conf.d/99docker-apt-config
 
+USER spacedrive
+HEALTHCHECK NONE
```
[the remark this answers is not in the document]
brxken128: Best practices recommend at least defining a `HEALTHCHECK` per layer.

apps/server/docker/entrypoint.sh
brxken128: I don't recall touching this line, maybe auto-formatting

brxken128:
Just going to close this, all I was trying to do was allow exposing the WEBUI on a different port for people who use host networking so their container can see other nodes wtithout having to worry about ports, and so that 8080 doesn't clash with another container (which will almost 100% happen).

Was very clearly a WIP as I'd mentioned at the start, only reason it wasn't a draft is because I wanted CI to keep running.

HeavenVolkoff:
> Just going to close this, all I was trying to do was allow exposing the WEBUI on a different port for people who use host networking so their container can see other nodes wtithout having to worry about ports, and so that 8080 doesn't clash with another container (which will almost 100% happen).

I think this is a valid addition, as almost all images have a similar option exposed. Could we re-open this?

PR #1445, opened 2023-10-09
[ENG-1202] Location dialog visual improvements

ameer2468:
- Adjusted accordian container color back to what it was previously.
- Width of location dialog adjusted so that the transition isn't weird when you toggle.

**Before**:

https://github.com/spacedriveapp/spacedrive/assets/33054370/a5c91725-9d12-41ed-b6fe-00c15434394c

**After**:

https://github.com/spacedriveapp/spacedrive/assets/33054370/75047a7e-2744-40af-8bcb-eb7b76cedcf4


brxken128 (approved):
Much better, I just had to actually scroll down to see the difference (I didn't at first and thought the first video was the final product LMAO). Nice work man 💪🏻

PR #1447, opened 2023-10-09
[ENG-625] Spacedrop improvements

oscartbeaumont:
- [x] Fix the buggy form
 - [x] Only a single dialog
 - [x] Improve error handling on sender
 - [x] Protocol work for multiple-files
 - [x] Keep UUID the same on both devices to aid debugging
 - [x] Transfer not ending in the UI for some reason
 - [x] UI for saving to a directory instead of a single file - sucks due to lack of Tauri's support for creating a directory in the open dialog. Will probs contribute this upstream in near future cause macOS has the API's for it.


brxken128 (approved):
Got a few clippy errors but CI still passed - either way, they were minor and everything else LGTM!

brxken128 (approved):
All looks good, I won't complain about the chunk of commented code as I do assume you'll use it as a reference/refactor it haha

PR #1448, opened 2023-10-10
[ENG-1086, ENG-1087] Add error and loading indicator to thumbnail

HeavenVolkoff:
Show toast when Original file fails loading in QuickPreview

https://github.com/spacedriveapp/spacedrive/assets/9082460/7ebf4379-d3c5-4922-ac56-4e4ed65718f1

### DIFF
--- a/interface/app/$libraryId/Explorer/QuickPreview/index.tsx
+++ b/interface/app/$libraryId/Explorer/QuickPreview/index.tsx
@@ -20,7 +20,16 @@ import {
 	useRspcLibraryContext,
 	useZodForm
 } from '@sd/client';
-import { dialogManager, DropdownMenu, Form, ModifierKeys, toast, Tooltip, z } from '@sd/ui';
+import {
+	dialogManager,
+	DropdownMenu,
+	Form,
+	ModifierKeys,
+	toast,
+	ToastMessage,
+	Tooltip,
+	z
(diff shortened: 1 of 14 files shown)


interface/app/$libraryId/Explorer/QuickPreview/index.tsx
```diff
@@ -396,6 +439,16 @@ export const QuickPreview = () => {
...
+									onLoad={(type) =>
+										type === 'ORIGINAL' && setThumbErrorToast(undefined)
+									}
+									onError={(type, error) =>
+										type === 'ORIGINAL' &&
+										setThumbErrorToast({
+											title: 'Error loading original file',
+											body: error.message
+										})
+									}
```
utkubakir, line 451: Can we not trigger a toast from here with `toast.error `instead?

HeavenVolkoff: We would need a state anyway to allow closing the toast whenever the quick preview is dismissed and the user didn't close the toast manually, so I found it to be a better approach to isolate all the show/hide logic for the toast into a `useEffect`, and just allow the component to `setThumbErrorToast` whenever it needed to show/hide an error toast.

jamiepine (approved):

PR #1449, opened 2023-10-10
chore: conditionally encourage applications based on open positions

amorriscode:
# What?

This PR conditionally displays "If any open positions suit you, apply now!" so that it is only shown when there are open positions.

# Why?

I was browsing the careers page and noticed that I was encouraged to apply even when there weren't open positions! 🥲

## Before

Screenshot 2023-10-09 at 9 06 56 PM

## After

Screenshot 2023-10-09 at 9 06 21 PM

Screenshot 2023-10-09 at 9 06 36 PM


utkubakir (approved):
Thank you!

PR #1450, opened 2023-10-10
fix: dev tools shortcut

amorriscode:
The dev tools are opened using `CmdOrCrtl+Alt+I` so I updated the `CONTRIBUTING.md` to reflect this.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -50,7 +50,7 @@ To quickly run only the desktop app after `prep`, you can use:
 
 - `pnpm tauri dev`
 
-  If necessary, the webview devtools can be opened by pressing Ctrl + Shift + I in the desktop app.
+  If necessary, the webview devtools can be opened by pressing Cmd + Alt + I (Mac) or Ctrl + Alt + I in the desktop app.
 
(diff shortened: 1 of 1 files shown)


fogodev:
https://tauri.app/v1/guides/debugging/application/#webview-console

amorriscode:
@fogodev my bad, I was looking at the code for the menu bar that is triggered by `CmdOrCtrl+Alt+I`. Is there any reason you didn't want me to fix my mistake vs. just closing this PR?

fogodev:
That code is MacOS specific. Sorry for just closing without much reasoning, sometimes we receive some "contributions" from bots. If you include the MacOS part alongside the already existing keybind for Linux and Windows, the PR will be good to go. I left it out because only with MacOS we have that global menu where you can just click to open dev tools, so just put the keybinds for other systems.

fogodev (changes requested):
This code is MacOS specific, you can include it alongside the other keybind already there, but don't remove it please.

amorriscode:
@fogodev totally hear you on that one! I hope to make some more meaningful commits in the future but just noticed this while getting spun up locally. I fixed the reference and also added a link to the docs you shared. 😄

PR #1454, opened 2023-10-10
[ENG-1210] Navigate to location after adding it

brxken128:
In the image below you can see the new checkbox, it is enabled by default. The mutation now returns the new location ID, which we pass back to the caller and navigate to (if the checkbox was selected).

This does indeed navigate to the location, but it does so before the indexer is invoked and therefore is entirely empty. @spacedriveapp/frontend, if you have any free time, please could you take a look at this?

image


PineappleRind:
I feel like the checkbox should be on the left side of the text

brxken128:
> I feel like the checkbox should be on the left side of the label

I'm not a UI dev so I'm not too sure, there's a bug with this branch anyway - hence why it's drafted.

ameer2468 (approved):

brxken128:
Gotta give @ameer2468 the credit for this one, he did the heavy lifting, I just added a checkbox and returned a value from the backend. Nice job man, it works great 🔥

PR #1455, opened 2023-10-10
[ENG-1214] Disable updater for `.deb` distrubutions

Brendonovich:
Updater now has to be enabled by the backend else it won't be available in `usePlatform`.
`.deb` detection logic taken from [Tauri's implementation](https://github.com/tauri-apps/tauri/blob/tauri-v1.3.0/core/tauri/src/app.rs#L944).


brxken128 (approved):
Looks solid to me!

PR #1498, opened 2023-10-11
Dynamic OS/version string and toggle-able download links for multi-download platforms

brxken128:
Linux now just says "AppImage", and the final image is when no platform/a currently unsupported platform is detected. The icons are all clickable, except for Android and Web which have been disabled.

The tooltip for the Apple logo was altered to just say MacOS, as that could ultimately lead to some confusion (as the icon is active). I considered possibly using an iMac icon, but decided against it. Another issue with having iOS on this tooltip is accessibility - a screen reader would've read out "iOS and MacOS" and could've been misleading.

The OS autodetect system is far from perfect - many privacy-centric extensions often spoof your user agent to a more common, less unique one, which is the one of the few reasons I made the icons clickable (and it's just intuitive). 

If I didn't know any better, I wouldn't have actually been able to download a build for my Mac as my UA was set to Windows, and the rest of the download links are all the way at the bottom of the landing page.

Screenshot 2023-10-11 at 14 18 27

### DIFF
--- a/apps/landing/src/pages/index.tsx
+++ b/apps/landing/src/pages/index.tsx
@@ -10,7 +10,7 @@ import { motion } from 'framer-motion';
 import dynamic from 'next/dynamic';
 import Head from 'next/head';
 import Image from 'next/image';
-import { ComponentProps, useEffect, useState } from 'react';
+import { ComponentProps, useEffect, useMemo, useState } from 'react';
 import { Tooltip, TooltipProvider, tw } from '@sd/ui';
 import NewBanner from '~/components/NewBanner';
 import PageWrapper from '~/components/PageWrapper';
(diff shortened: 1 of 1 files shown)


apps/landing/src/pages/index.tsx
```diff
@@ -329,7 +365,14 @@ const Platform = ({ icon: Icon, label, ...props }: ComponentProps<'a'> & Props)
 	return (
 		<Tooltip label={label}>
 			<Outer {...props}>
-				<Icon size={25} className="h-[25px] opacity-80" weight="fill" />
+				<Icon
+					size={25}
+					className={clsx('h-[25px]', {
+						['opacity-20']: props.iconDisabled,
+						['opacity-80']: props.iconDisabled === undefined
```
Brendonovich, line 372: i'd prefer a ternary here

apps/landing/src/pages/index.tsx
```diff
@@ -193,7 +214,7 @@ export default function HomePage() {
 							/>
 						)}
 
-						{multipleDownloads == null && (
+						{downloadEntry === undefined && (
```
utkubakir, line 217: `!downloadEntry`

Issue #1509, opened 2023-10-11
Job Manager number format inconsistency


PineappleRind:
### Describe the bug

<img width="382" alt="image" src="https://github.com/spacedriveapp/spacedrive/assets/65303441/9aba6275-8bea-4156-a40a-00b89e733f70">

Some numbers are comma-separated and some aren't.

### Reproduction

1. Start indexing a directory
2. See the formatting of the numbers

### Expected behavior

Either every number *is* formatted like this, or *isn't*

### Platform and versions

```Shell
Spacedrive v0.1.0 3c55bfdd
```

### Additional context

Working on a PR for this now

PR #1511, opened 2023-10-11
job manager: utils/formatNumber all numbers

PineappleRind:
This formats all numbers shown in the Job Manager on the Typescript side.

There are non-formatted numbers that are shown via Rust that I'm not sure how to handle, specifically:

- Found {} entries (`core::location::indexer::indexer_job` line 538)
	Screenshot 2023-10-11 at 1 11 43 PM
- Writing chunk {} of {} to database (`core::location::indexer::indexer_job` line 340)
	Screenshot 2023-10-11 at 1 18 03 PM

Relevant issue: #1509


jamiepine:
Yeah the job manager needs to emit purely data values, no text. We should change that asap imo.

PineappleRind:
hmm... so add members to the [ScanProgress enum](https://github.com/spacedriveapp/spacedrive/blob/main/core/src/location/indexer/indexer_job.rs#L109) like CurrentPath and FoundEntries, and send those instead of [text](https://github.com/spacedriveapp/spacedrive/blob/main/core/src/location/indexer/indexer_job.rs#L537)?

brxken128:
> hmm... so add members to the [ScanProgress enum](https://github.com/spacedriveapp/spacedrive/blob/main/core/src/location/indexer/indexer_job.rs#L109) like CurrentPath and FoundEntries, and send those instead of [text](https://github.com/spacedriveapp/spacedrive/blob/main/core/src/location/indexer/indexer_job.rs#L537)?

Kind of - I have a branch somewhere where I partially whipped something up. Instead of JobReportUpdate messages, we were going to send data and have the frontend format it appropriately. We can't just send pure random types of course, we should still probably send at least an enum (e.g `JobReport::Progress(i32)` for a percentage between 0-100, `JobReport::EstimatedCompletionTime(DateTime<FixedOffset>)` for the estimated remaining time, etc).

This will also help majorly once we start the huge task of localisation and translating the app so a wider range of the population can use it with ease. Ideally Rust should be providing pure data, and the frontend should be handling the formatting and displaying of that data in a fast an efficient manner.

Sadly this isn't the case just yet, but we are slowly making progress and I think the job system is due a good overhaul regardless - it's inefficient in some cases, and there's been a weird persisting concurrency issue where jobs aren't truly marked as complete for around 5 seconds after the final step. Pausing/resuming could do with some work too, but most of this is out of scope for this PR of course.

PineappleRind:
tried editing some Rust, added Data to JobReportUpdate and used JobReport.metadata to store it on the job report
Screenshot 2023-10-13 at 8 43 30 AM

I may be in over my head - not sure what the best way to do any of this is, never mind type safety (was using explicit `any` on client)

...so I think we can leave this PR as is, which fixes this inconsistency:
image

jamiepine (approved):

PR #1516, opened 2023-10-11
[ENG-870] Fix wording for delete dialog

### LINKED ISSUE #1384: [ENG-870] Deleting a folder should not say "Deleted 1 file"
It should calculate how many items are in it or even "deleted {folderName}" might work

<sub>From [SyncLinear.com](https://synclinear.com) | [ENG-870](https://linear.app/spacedriveapp/issue/ENG-870/deleting-a-folder-should-not-say-deleted-1-file)</sub>

Tilo-K:
*Prefix: I'm really new to the project but wanted to try to tackle an issue, I'm not familiar with the Codebase, so if this is totally wrong or there is need to change something please let me know. I would really appreciate it if someone would take the time to give some feedback on this.*

Currently, the delete dialog always says that you are deleting a file, even if it is a directory or a mix of items.

To fix this:
The DeleteDialog now accepts two props, which specify the type of files to be deleted.
Based on this information, the wording is adjusted.

Example of directory deletion
Example of mix deletion

Closes #1384

### DIFF
--- a/interface/app/$libraryId/Explorer/FilePath/DeleteDialog.tsx
+++ b/interface/app/$libraryId/Explorer/FilePath/DeleteDialog.tsx
@@ -5,12 +5,25 @@ interface Props extends UseDialogProps {
 	locationId: number;
 	rescan?: () => void;
 	pathIds: number[];
+	includesDirectorys?: boolean;
+	includesFiles?: boolean;
 }
 
 export default (props: Props) => {
 	const deleteFile = useLibraryMutation('files.deleteFiles');
 
 	const form = useZodForm();
+
+	let type = "file";
+
+	if(props.includesDirectorys){
+		type = "directory";
+		if(props.includesFiles){
+			type = "item";
+		}
+	}
+
+	const description = `Warning: This will delete your ${type} forever, we don't have a trash can yet...`;
 
 	return (
 		<Dialog
@@ -24,8 +37,8 @@ export default (props: Props) => {
 				props.rescan?.();
 			})}
 			dialog={useDialog(props)}
-			title="Delete a file"
-			description="Warning: This will delete your file forever, we don't have a trash can yet..."
+			title={"Delete a " + type}
+			description={description}
 			loading={deleteFile.isLoading}
 			ctaLabel="Delete"
 			ctaDanger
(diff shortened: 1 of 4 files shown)


brxken128 (dismissed):
I think the second image should say "Delete x items" if at least one directory and one file is selected, and "Delete x directories" and "Delete x files" if they're just purely files and directories respectively.

For single items, it should say "Delete a file" or "Delete a directory" respectively too.

Would take a little bit more logic, but would definitely provide even better wording and clarity to the user.

interface/app/$libraryId/Explorer/FilePath/DeleteDialog.tsx
```diff
@@ -5,13 +5,26 @@ interface Props extends UseDialogProps {
 	locationId: number;
 	rescan?: () => void;
 	pathIds: number[];
+	includesDirectorys?: boolean;
```
brxken128, line 8: This should be `Directories` :)

interface/app/$libraryId/Explorer/FilePath/DeleteDialog.tsx
```diff
@@ -5,13 +5,48 @@ interface Props extends UseDialogProps {
...
+		prefix = (fileCount + dirCount).toString();
+	}
+
+	return { type, prefix };
 }
 
 export default (props: Props) => {
 	const deleteFile = useLibraryMutation('files.deleteFiles');
+	props.dirCount ??= 0;
+	props.fileCount ??= 0;
```
ameer2468, line 42: This is great, tested the PR and works as expected. However these 2 lines of code needs to be adjusted.

Directly mutating props like this in a React component is considered a bad practice. Props in React components are meant to be read-only and should not be modified. If you're looking to set default values for props, consider using defaultProps or set them before passing them into the component.

Tilo-K: Thank you for the feedback, I implemented the first suggestion.

ameer2468 (approved):

ameer2468:
Thanks for the response @Tilo-K 

I have done testing. Currently this implementation only considers the count when you press Delete within the context menu, however, when using the shortcut on keyboard to delete, it does not count.

Tilo-K:
Sorry for the delay @ameer2468
I now also added the file / directory counts to the shortcut. This should work now.

ameer2468 (approved):
Good work! tested and works well.

PR #1523, opened 2023-10-12
[ENG-1225] Put brakes on thumbnail generation for ephemeral locations

fogodev:
This PR introduces new capabilities to our thumbnailer actor, procing thumbnails in batches and with background processing for ephemeral locations, prioritizing the directory the user has opened in the explorer.


jamiepine (approved):
amazing work. tested alongside release version, performance is incomparably better

PR #1525, opened 2023-10-12
Smarter release workflow

Brendonovich:
This PR introduces a new publish action written in JS that will for now run alongside the existing one. It should take care of renaming assets to those expected by the Download & Update APIs, and can automatically upload assets to a new/existing release if the workflow was started for a tag.


.github/actions/publish-artifacts-new/tsconfig.json
HeavenVolkoff, line 1: Nitpick: remove the commented props, no real gain in leaving them, TS docs is a better place to read about what each one does anyway

Brendonovich: yeah can do, they're just included from `pnpm tsc --init`

.github/workflows/release.yml
```diff
@@ -109,8 +116,32 @@ jobs:
           APPLE_API_KEY: ${{ secrets.APPLE_API_KEY }}
           SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
 
-      - name: Publish Artifacts
+      - name: Publish Artifacts (Old)
         uses: ./.github/actions/publish-artifacts
         with:
           target: ${{ matrix.settings.target }}
           profile: release
```
HeavenVolkoff, line 123: I think we can just drop this, the newer one is better

HeavenVolkoff:
LGTM, just want to do a test release before merging to make sure it works fine

.github/workflows/release.yml
Brendonovich: I'd rather we keep both around just to make sure, means we can test it in prod

.github/actions/publish-artifacts/action.yml
HeavenVolkoff, line 1: The build failed, I think the action was replaced by mistake with the workflow content

HeavenVolkoff (changes requested):

.github/actions/publish-artifacts/action.yml
Brendonovich: lol

.github/workflows/release.yml
```diff
@@ -109,8 +116,26 @@ jobs:
           APPLE_API_KEY: ${{ secrets.APPLE_API_KEY }}
           SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
 
-      - name: Publish Artifacts
+      - name: Publish Artifcats
```
HeavenVolkoff, line 119: 🤣

.github/actions/publish-artifacts/index.ts
```diff
@@ -0,0 +1,66 @@
...
+			const artifactPath = `${ARTIFACTS_DIR}/${artifactName}`;
+
+			const updaterPath = files.find((file) => file.endsWith(updaterExt));
+			if (!updaterPath) throw `Updater path not found. Files: ${files}`;
+
+			// https://tauri.app/v1/guides/distribution/updater#update-artifacts
+			await io.mv(updaterPath, artifactPath);
+			await io.mv(`${updaterPath}.sig`, `${artifactPath}.sig`);
+
+			await client.uploadArtifact(artifactName, [`Spacedrive-Updater-*`], ARTIFACTS_DIR);
```
HeavenVolkoff, line 61: I think this is failing to find the updater file, according to the logs in the last 2 test relases runs. We can get better logs, but we need to se the following envvar in the repo settings: `ACTIONS_STEP_DEBUG=true`

HeavenVolkoff (changes requested):

.github/actions/publish-artifacts/index.ts
HeavenVolkoff: image

HeavenVolkoff: `client.uploadArtifact` doesn't accept globs as paths

.github/workflows/release.yml
Brendonovich: 🐈

.github/actions/publish-artifacts/index.ts
```diff
@@ -45,7 +45,7 @@ async function run() {
 		if (!standalonePath) throw `Standalone path not found. Files: ${files}`;
 
 		await io.mv(standalonePath, artifactPath);
-		await client.uploadArtifact(ARTIFACT_NAME, [artifactPath], ARTIFACTS_DIR);
+		await client.uploadArtifact(ARTIFACT_NAME, [artifactPath], './');
```
HeavenVolkoff, line 48: If I understood the API correctly, `rootDirectory`  defines how the structure of the zip will be, so putting `'.'` here would mean the whole `artifactPath` will be included in the zip, no?

https://github.com/actions/toolkit/tree/@actions/artifact@1.1.1/packages/artifact#upload-an-artifact

HeavenVolkoff (changes requested):

.github/actions/publish-artifacts/index.ts
Brendonovich: Yep seems like that's the case

utkubakir (approved):

jamiepine (approved):

PR #1558, opened 2023-10-13
add comments in metadata.rs

alvarosamudio:
I try to make the code understandable for future revisions and improvements


Brendonovich:
These comments don't add any extra information that can't be gleamed from the code itself. Please open issues before making PRs so we can discuss your changes before any code is written.

PR #1575, opened 2023-10-13
[ENG-1248] Add additional size options to media view

niikeec:


brxken128 (approved):

brxken128:
I like this, should give pretty fine-grained control (and it'll look smoother when making large changes as there's less skipping)

PR #1577, opened 2023-10-13
Auto version pulling

misxki:
Just automatic version pulling from Github releases/latest for the spacedrive landing page (so it doesn't have to be manually).

### DIFF
--- a/apps/landing/src/pages/index.tsx
+++ b/apps/landing/src/pages/index.tsx
@@ -125,7 +125,7 @@ export default function HomePage() {
 	}, []);
 
 	useEffect(() => {
-		getLatestSpacedriveVersion().then((res) => setSpacedriveVersion(res));
+		getLatestSpacedriveVersion().then(setSpacedriveVersion);
 	}, []);
 
 	useEffect(() => {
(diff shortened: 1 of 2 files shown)


apps/landing/src/pages/index.tsx
```diff
@@ -124,6 +124,10 @@ export default function HomePage() {
 		};
 	}, []);
 
+	useEffect(() => {
+		getLatestSpacedriveVersion().then((res) => setSpacedriveVersion(res));
```
PineappleRind, line 128:
```suggestion
		getLatestSpacedriveVersion().then(setSpacedriveVersion);
```
shorter

Brendonovich:
Once #1587 is merged I'm going to implement changelog pulling on the server, it'd use a similar fetch as you're doing but on the server instead. Would you like to re-implement this to fetch server-side in a different PR once I'm done with the changelog stuff?

misxki:
Yea sure, I can do that!

Brendonovich:
#1592 is done now, feel free to implement this! Don't use any client components though.

PR #1579, opened 2023-10-13
Tauri 1.5.2 + normalize and update frontend dev deps

HeavenVolkoff:
- Upgrade Tauri to version 1.5.2
- Update and standardize front-end development dependencies (~500 dependencies removed due to multiple slightly different versions of previous deps being installed)
- Code formatting + some ts typing fixes


brxken128 (approved):
Everything seems to work flawlessly on MacOS 🙏🏻

PR #1582, opened 2023-10-14
[ENG-1283, ENG-1284, partially ENG-1303] Fix MacOS window bar showing when full-screen

brxken128:
This properly autohides the MacOS window bar when the window is full-screen (with the green traffic light) - there was some (evidently buggy) swift code, and after removing it and letting the OS handle it everything works fine.

Many options are disabled until a valid library is present, as it wouldn't make sense to offer "Add location" or "Search" with nothing to actually search.

This also:

- Fixes the "Settings" redirect within the menu bar.
- Adds a toast when the explorer is refreshed, and the shallow indexer is invoked.
- Adds  a whole variety of shortcuts (some not hooked up)
- Adds some transitions when maximizing/unmaximizing
- Refreshing the explorer now displays a toast letting the user know it's being shallow indexed (keybind isn't hooked up)
- Adds a "View > Overview" button (that also works and has a keybind)
- Allows the webview to be refreshed in production builds, either via the menu bar or cmd/ctrl+option/alt+shift+r (which frees up cmd/ctrl+r for actually reloading the explorer). This took inspiration from Windows' ctrl+win+shift+b keybind.
- Overhauls the menu bar, with a "dummy" (not hooked-up) library and layout selector, options for creating new files/directories, options for adding a new location, and even an option to create a new library! (The new library option is disabled until after onboarding though, as onboarding should be completed at least once).
- Conditionally compiles a lot of the MacOS-specific code insteasd of just not running it for (marginally) reduced output binary sizes on other platforms
- Added a dedicated Tauri function so that we have fine-grained control over the state of the menu bar items (for example, being able to disable the "Reload explorer" option when the explorer isn't visible).

Many keybinds are not hooked up yet,  but can/will be in future PRs.

~~I'll keep this as a draft as there's still two more things to fix~~:

- [x] The rounded window corners still show even when the window is fullscreened - these need to be conditionally disabled
- [x] The "Settings" button needs to be disabled until we invoke a `hasLibrary` function, or via a similar mechanism

Edit: Another bug

- [x] Rounded corners aren't conditionally showing (when not in fullscreen mode). We can likely make a Tauri invoke function which checks if the window is maximized or not, and use `clsx` to conditionally apply the rounded borders.


interface/hooks/useKeybindEventHandler.ts
```diff
@@ -0,0 +1,36 @@
...
+			} else if (e.detail.action === 'window_maximized') {
+				setIsWindowMaximized(true);
+				e.preventDefault();
+				return;
+			} else if (e.detail.action === 'window_not_maximized') {
+				setIsWindowMaximized(false);
+				e.preventDefault();
+				return;
+			}
+		};
```
brxken128, line 31: Could probably be a switch-case, with `e.preventDefault();` and `return` happening regardless.

brxken128:
There's a tiny tiny bug because Apple can't decide whether full screen should be called "full screen" or "maximized"

ameer2468 (approved):

PR #1583, opened 2023-10-14
[ENG-1229] Show hidden files not working correctly

HeavenVolkoff:
- Ensure hidden property is always updated when file is modified
 - Fix `with_hidden` not being set according to the explorer settings in ephemeral path query
 - Ignore system file `My Documents` on Windows

### DIFF
--- a/core/src/location/indexer/rules/seed.rs
+++ b/core/src/location/indexer/rules/seed.rs
@@ -104,7 +104,7 @@ pub fn no_os_protected() -> SystemIndexerRule {
                         "C:/Users/*/ntuser.dat*",
                         "C:/Users/*/{ntuser.ini,ntuser.dat,NTUSER.DAT}",
                         // User special folders (most of these the user dont even have permission to access)
-                        "C:/Users/*/{Cookies,AppData,NetHood,Recent,PrintHood,SendTo,Templates,Start Menu,Application Data,Local Settings}",
+                        "C:/Users/*/{Cookies,AppData,NetHood,Recent,PrintHood,SendTo,Templates,Start Menu,Application Data,Local Settings,My Documents}",
                         // System special folders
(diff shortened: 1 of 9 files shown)


core/src/location/indexer/rules/seed.rs
```diff
@@ -104,7 +104,7 @@ pub fn no_os_protected() -> SystemIndexerRule {
                         "C:/Users/*/ntuser.dat*",
                         "C:/Users/*/{ntuser.ini,ntuser.dat,NTUSER.DAT}",
                         // User special folders (most of these the user dont even have permission to access)
-                        "C:/Users/*/{Cookies,AppData,NetHood,Recent,PrintHood,SendTo,Templates,Start Menu,Application Data,Local Settings}",
+                        "C:/Users/*/{Cookies,AppData,NetHood,Recent,PrintHood,SendTo,Templates,Start Menu,Application Data,Local Settings,My Documents}",
```
brxken128, line 107: Why are we hiding documents? Not opposed, just curious - some games save their files there, having easy access to our future "redundancy" feature for those files seems like an awesome use-case. I can't count the amount of times I've had to restart a game from scratch due to this haha

HeavenVolkoff: This `My Documents` isn't the user's actual `Documents` directory, but a system-only symlink maintained for compatibility reasons. Substantial changes occurred to the `My Documents` directory from Windows Vista onward, including being renamed as `Documents` (and localized in various languages). To maintain compatibility with older software, Windows kept a hidden `My Documents` symlink in the user's home directory. Although the explorer doesn't display this entry to the user, Spacedrive detects it because it's a real file. So, it should be ignored.

Brendonovich (approved):

jamiepine (approved):
perfect

jamiepine (approved):

PR #1585, opened 2023-10-14
[ENG-1267] Move thumbnails generation away from media processor

fogodev:
This PR remove the thumbnails generation from media processor and makes use of the brand new thumbnailer actor, with blazingly fast performance.


core/src/location/non_indexed.rs
```diff
@@ -161,20 +166,39 @@ pub async fn walk(
...
+						kind,
+						ObjectKind::Image | ObjectKind::Video | ObjectKind::Document
+					)
+				}
+
+				#[cfg(not(feature = "ffmpeg"))]
+				{
+					matches!(kind, ObjectKind::Image | ObjectKind::Video)
+				}
+			};
```
brxken128, line 182: Should `Video` still be here given that `ffmpeg` is disabled - I think it should be `Document` instead maybe?

fogodev: Small copy/paste issue, fixed lol

brxken128 (approved):
I'm happy with these changes, and beyond impressed with the optimizations you made tbh. Nice work man!

PR #1587, opened 2023-10-14
Migrate landing page to app dir

Brendonovich:
felt like it :)

Migrates `sd-landing` to Next's App Router. Also enables tree shaking for `@sd/ui` by adding `sideEffects: false` to the `package.json`. JS on first load is now and order of magnitude lower and landing page organisation is much simpler (thanks nested layouts!)

- ~~Has a bug with fading out the background on scroll~~


ameer2468 (dismissed):

ameer2468:
Tested on Mac (Google chrome & Safari), on Windows (Google Chrome), and iPhone (Chrome & Safari)

PR #1589, opened 2023-10-15
Remove cargo config templating

Brendonovich:
Uses relative envs and build scripts instead of absolute urls generated by the mustache template.

Needs testing on Windows and Linux!

### DIFF
--- a/.cargo/config.toml.mustache
+++ b/.cargo/config.toml.mustache
@@ -1,4 +1,5 @@
 [env]
+DEPS_PATH = { relative = true, value = "apps/.deps" }
 {{#protoc}}
 PROTOC = "{{{protoc}}}"
 {{/protoc}}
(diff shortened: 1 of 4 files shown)


HeavenVolkoff:
`FFMPEG_DIR` must not be set on Linux, because we use the system's ffmpeg to build (for now), and if this envvar is set the `ffmpeg-sys` crate always use it, regardless if the path is not valid, and the build will fail

Brendonovich:
linux makes this hard

Brendonovich:
closing in favour of upcoming @HeavenVolkoff PR

brxken128:
> linux makes this hard

you're lucky tauri doesn't have freebsd support yet

Brendonovich:
nvm keeping this around but only with build script stuff

Brendonovich:
nvm

PR #1604, opened 2023-10-17
[END-1309] Fix explorer query invalidation

jamiepine:
In most cases we invalidate search.paths and not search.objects, this change makes sure both are always invalided. Our invalidation system is terrible, but that's another PR.


brxken128 (approved):
I knew something was missing haha, LGTM

PR #1605, opened 2023-10-17
Bump `.nvmrc` version

oscartbeaumont:
My system is choosing a node v18 version prior to v18.17 and that is causing `pnpm i` to fail the version check.

Our `CONTRIBUTING.md` says v18.17 is the minimum version we support so copying from it.


brxken128:
On that note, Void Linux official (non-`fnm`) support is blocked until we can figure out how to make v18.16.0 work or Void update their package. There's an open PR but CI won't pass so it's somewhat stale.

brxken128 (approved):

PR #1606, opened 2023-10-17
[ENG-1310] Spacedrop better testing

oscartbeaumont:
Adding better tests found a bug where two fields where being encoded and decoding in opposing orders. Incredible it was working.

Why unit test?

Debugging the protocol takes soooo much of my time (especially for small things where it works but not quite right) so automating it saves me more time than writing test costs. It also means people on the team other can me can just jump and modify it without risking breaking something.

### DIFF
--- a/core/src/p2p/p2p_manager.rs
+++ b/core/src/p2p/p2p_manager.rs
@@ -234,7 +234,7 @@ impl P2PManager {
 										debug!("Received ping from peer '{}'", event.peer_id);
 									}
 									Header::Spacedrop(req) => {
-										let id = Uuid::new_v4(); // TODO: Get ID from the remote
+										let id = req.id.clone();
 										let (tx, rx) = oneshot::channel();
 
 										info!(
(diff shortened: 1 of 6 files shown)


brxken128:
Maybe a `p2p` CI workflow using `nextest` would be ideal, [here's the one I have set up for crypto currently](https://raw.githubusercontent.com/spacedriveapp/spacedrive/eng-440/.github/workflows/crypto.yml)?

oscartbeaumont:
Definitely a good idea but will do that separately!

brxken128:
> Definitely a good idea but will do that separately!

Oh for sure - unit tests are probably critical at this point and we should push for them heavily. The #610 branch has over 100 and it's going to be a lot more soon haha

brxken128 (approved):
Tests run fine, I fixed the clippy warning, all looks good to me!

oscartbeaumont:
I gotta get better at checking Clippy locally. For some reason it just keeps slipping my mind.

brxken128:
I use `"rust-analyzer.check.command": "clippy",` in my VSCode `settings.json` - takes a tad longer but prevents this haha

oscartbeaumont:
Huh, never knew that was a thing. Will give it a shot. Hopefully, it doesn't negatively affect LSP performance but we will see.

brxken128:
> Huh, never knew that was a thing. Will give it a shot. Hopefully, it doesn't negatively affect LSP performance but we will see.

I think it adds maybe a second or two, but usually when I work on independent crates that don't rely on other crates in the workspace I just comment them out from `workspace.members` LMAO

PR #1609, opened 2023-10-17
[ENG-1291] enable/disable P2P & configurable port

oscartbeaumont:


core/src/p2p/p2p_manager.rs
```diff
@@ -466,8 +471,10 @@ impl P2PManager {
 			name: config.name.clone(),
 			operating_system: Some(OperatingSystem::get_os()),
 			version: Some(env!("CARGO_PKG_VERSION").to_string()),
-			email: config.p2p_email.clone(),
-			img_url: config.p2p_img_url.clone(),
+			// TODO: Source these via Spacedrive account
+			// TODO: Maybe anonymise them like Apple do
```
brxken128, line 475: `blake3` should do the job here for the email, or `Hkdf<Sha512>` (or SHA3-512).

Personally I'd just `blake3` it or remove it entirely once we become reliant on public keys (sorry for how long crypto has taken), but there are quite a few options for now.

brxken128 (approved):
I haven't built nor tested, but it's feature-gated and the code looks good to me. Provided CI is golden, this is good to go IMO!

core/src/p2p/p2p_manager.rs
oscartbeaumont: My main concern with hashing is it doesn't prevent the tracking element. The hash will always be the same for the same node. I think a lot of modern machines do MAC address switching for example to try and make them harder to track. I was thinking it would be cool if we could have a rotating salt but then we gotta recompute a hash for the entire local address book frequently which would be more compute and potentially make the discovery process slower to the end-user depending on their machine.

I was chatting with Brendan and Jamie and I am probs gonna remove those fields for now cause it's looking like Spacedrive accounts and P2P won't be integrated for a while.

PR #1629, opened 2023-10-19
[ENG-1324] Fix show object size on ephemeral locations

niikeec:


fogodev:
I was checking out this PR out of curiosity as frontend isn't my strong suit, and I noted that while object sizes in list and grid views are nice now, we're still showing size for directories in the inspector, which can be misleading as this size also doesn't correspond to the directory actual size. Is it worth to fix this one too in this PR?

jamiepine (approved):

Brendonovich:
> while object sizes in list and grid views are nice now, we're still showing size for directories in the inspector
ENG-1329

PR #1633, opened 2023-10-19
Landing page GitHub releases pulling

misxki:
Same again as last week for [#1577](https://github.com/spacedriveapp/spacedrive/pull/1577).

~~Nextjs13 is pretty new for me, so I'm not sure if that's exactly what you want.
It is fetching on client side from SD API. I could pass it as a prop (so it would fetch it server side) but that seems a bit jank imo, I might be wrong tho.
Edit: Would be nice to mention, I do have the server side fetching version.~~

Read @Brendonovich comment again, pushed my server side version now. Still feels bit jank ;P
There is probably a cleaner solution code wise.


apps/landing/src/app/Downloads.tsx
```diff
@@ -50,7 +48,11 @@ const platforms = {
 
 const BASE_DL_LINK = '/api/releases/desktop/stable';
 
-export function Downloads() {
+interface DownloadPageProps {
+	latestVersion: ReactNode
+}
```
Brendonovich, line 53: Can be renamed to `Props` + can take in a `string` rather than `ReactNode`

apps/landing/src/app/api/releases/latest/route.ts
```diff
@@ -0,0 +1,13 @@
+import { getLatestRelease, getReleaseFrontmatter, githubFetch } from '../../github';
+
+export const runtime = 'edge';
+
+export async function GET() {
```
Brendonovich, line 5: Don't need this route rn

apps/landing/src/components/LatestVersion.tsx
```diff
@@ -0,0 +1,7 @@
+import { getLatestRelease, githubFetch } from "~/app/api/github";
+
+export async function LatestVersion() {
+    const release = await githubFetch(getLatestRelease);
```
Brendonovich, line 4: This fetch can be done in `app/page.tsx` rather than in this dedicated component.

apps/landing/src/components/LatestVersion.tsx
```diff
@@ -0,0 +1,7 @@
...
+    
+    return (<>Alpha v{ release.tag_name }</>)
```
Brendonovich, line 6: This is probably fine for now but `Alpha` should be sourced from the release frontmatter's `category`.

Brendonovich (changes requested):

Brendonovich (approved):
fixed some stuff so we can get this merged for release. thanks for the help!

PR #1634, opened 2023-10-19
Enable thumbnail support for webp images

### LINKED ISSUE #1614: Adding media files with errors: Had an error generating thumbnail
### Describe the bug

I'm adding a folder of photos as a new location to the app.
And I found that the job completed with errors. 
Screenshot 2023-10-18 at 01 54 09

Here shows many error with
> Had an error generating thumbnail for "xxxxx/pic.webp"

Screenshot 2023-10-18 at 01 53 19

There are several webp photos inside this folder. And all the above errors are occured with webp.
 

### Reproduction

1. Click "Add Location"
2. Choose a folder full of photos.
3. Enable "Only Images" option in Advanced settings
4. Click Add and wait for the job to be completed

### Expected behavior

Completed with no errors

### Platform and versions

```Shell
I'm using Version 0.1.1 (20231013.100750) on MacBook Pro with Apple M1 Pro chip.
```

### Stack trace

_No response_

### Additional context

_No response_

HeavenVolkoff:
Closes #1614

### DIFF
--- a/crates/images/src/consts.rs
+++ b/crates/images/src/consts.rs
@@ -11,9 +11,9 @@ pub const MAXIMUM_FILE_SIZE: u64 = MIB * 192;
 /// These are roughly all extensions supported by the `image` crate, as of `v0.24.7`.
 ///
 /// We only support images that have both good encoding and decoding support, without external C-based dependencies (e.g. `avif`)
-pub const GENERIC_EXTENSIONS: [&str; 16] = [
+pub const GENERIC_EXTENSIONS: [&str; 17] = [
 	"bmp", "dib", "ff", "gif", "ico", "jpg", "jpeg", "png", "pnm", "qoi", "tga", "icb", "vda",
-	"vst", "tiff", "tif",
+	"vst", "tiff", "tif", "webp",
 ];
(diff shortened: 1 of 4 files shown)


crates/images/src/consts.rs
```diff
@@ -67,6 +67,7 @@ pub enum ConvertableExtension {
 	Svg,
 	Svgz,
 	Pdf,
+	Webp,
```
brxken128, line 70: This is fine for now, but when the image conversion stuff is fixed then it'll need to be blacklisted, unless we also use the external `webp` crate (but then the extension should be removed from `GENERIC_EXTENSIONS` and into its own thing). The `webp` crate can be removed from the `Cargo.toml` here as we're only decoding - I had planned to add full encode/decode support with it so we can convert to and from `webp` but this gets thumbnails working.

The image conversion UI stuff also builds on top of this, and implements rules for certain images (e.g. they can be converted to another format, but not the other way around).

brxken128 (changes requested):
Easy fix for sure, just needs the `webp` crate removing from the `Cargo.toml` file as it's entirely unused iirc.

fogodev (dismissed):

brxken128 (approved):

PR #1641, opened 2023-10-20
[ENG-1332] Fix regression causing P2P to be disabled by default

oscartbeaumont:
The `#[serde(default)]` on `is_enabled` will be `false` cause it's `bool::default` not `Self::default`  🤦


Brendonovich (approved):
skill issue

PR #1659, opened 2023-10-22
Improve error message with rustup not found

devqore:
While I've configured `spacedrive` from sources on Arch Linux, I got message like bellow, even I had both `rustc` and `cargo`:
```
$ ./scripts/setup.sh 
Spacedrive Development Environment Setup
To set up your machine for Spacedrive development, this script will install some required dependencies with your system package manager

Press Enter to continue

Rust was not found. Ensure the 'rustc' and 'cargo' binaries are in your $PATH. https://rustup.rs
Rust was not found. Ensure the 'rustc' and 'cargo' binaries are in your $PATH. https://rustup.rs
Rust was not found. Ensure the 'rustc' and 'cargo' binaries are in your $PATH. https://rustup.rs
```
But when I checked in the code, command checks additionally `rustup` package:
```
  if ! has rustup rustc cargo; then
```
So, because on Arch Linux [rustup](https://archlinux.org/packages/extra/x86_64/rustup/) is a different package it would simplify for other users to add also information that `rustup` is needed in the path.

### DIFF
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -52,7 +52,7 @@ if [ "${CI:-}" != "true" ]; then
       'https://pnpm.io/installation'
   fi
 
-  if ! has rustup rustc cargo; then
+  if ! has rustc cargo; then
     err 'Rust was not found.' \
       "Ensure the 'rustc' and 'cargo' binaries are in your \$PATH." \
       'https://rustup.rs'
@@ -72,6 +72,12 @@ if [ "${1:-}" = "mobile" ]; then
     err 'python3 was not found.' \
       'This is required for Android mobile development.' \
       "Ensure 'python3' is available in your \$PATH and try again."
+  fi
+
+  if ! has rustup; then
+    err 'Rust was not found.' \
+      "Ensure the 'rustup' binary is in your \$PATH." \
+      'https://rustup.rs'
   fi
 
   # Android targets
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -75,7 +75,7 @@ if [ "${1:-}" = "mobile" ]; then
   fi
 
   if ! has rustup; then
-    err 'Rust was not found.' \
+    err 'Rustup was not found. It is required for cross-compiling rust to mobile targets.' \
       "Ensure the 'rustup' binary is in your \$PATH." \
       'https://rustup.rs'
   fi
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -8,7 +8,7 @@ fi
 
 err() {
   for _line in "$@"; do
-    echo "$@" >&2
+    echo "$_line" >&2
   done
   exit 1
 }


HeavenVolkoff:
Hey, thank you for your contribution. Actually, `rustup` is not really required for building the desktop app and server, if you already have rust installed from another source. It is only required for mobile builds. So, you could remove the rustup check from that `if`, and create 2 new checks for `rustup` only for the mobile build in the following places:

https://github.com/spacedriveapp/spacedrive/blob/4c785eaca07752377ef17ac1aa62a7897f556964/scripts/setup.sh#L78-L80

https://github.com/spacedriveapp/spacedrive/blob/4c785eaca07752377ef17ac1aa62a7897f556964/scripts/setup.sh#L106-L108

brxken128:
> Hey, thank you for your contribution. Actually, `rustup` is not really required for building the desktop app and server, if you already have rust installed from another source. It is only required for mobile builds. So, you could remove the rustup check from that `if`, and create 2 new checks for `rustup` only for the mobile build in the following places:
> 
> https://github.com/spacedriveapp/spacedrive/blob/4c785eaca07752377ef17ac1aa62a7897f556964/scripts/setup.sh#L78-L80
> 
> https://github.com/spacedriveapp/spacedrive/blob/4c785eaca07752377ef17ac1aa62a7897f556964/scripts/setup.sh#L106-L108

Another great use-case for this would be on Gentoo, where Rust is actually part of the base system I believe. It's both a blessing and a curse though, and you are still able to use `rustup` alongside it luckily (or swap for the rust-bin package).

Either way, I'm all for these changes - provided `cargo` and the appropriate target(s) are installed, all should be good.

devqore:
PR updated but because `MOBILE=1` is set here [scripts/setup.sh#L69](https://github.com/devqore/spacedrive/blob/0b9d46af47e3afbcf06ad1e326b9179aac17ff77/scripts/setup.sh#L69) and on iOS is called only when `MOBILE` is already equal to 1 [scripts/setup.sh#L105](https://github.com/devqore/spacedrive/blob/0b9d46af47e3afbcf06ad1e326b9179aac17ff77/scripts/setup.sh#L105) one check should be enough. But it also means that all android rustup targets are currently installed on the iOS either.

scripts/setup.sh
```diff
@@ -74,6 +74,12 @@ if [ "${1:-}" = "mobile" ]; then
       "Ensure 'python3' is available in your \$PATH and try again."
   fi
 
+  if ! has rustup; then
+    err 'Rust was not found.' \
```
HeavenVolkoff, line 78: Could you change this first message to:
`'Rustup was not found. It is required for cross-compiling rust to mobile targets'`

HeavenVolkoff (changes requested):

HeavenVolkoff:
> PR updated but because `MOBILE=1` is set here [scripts/setup.sh#L69](https://github.com/devqore/spacedrive/blob/0b9d46af47e3afbcf06ad1e326b9179aac17ff77/scripts/setup.sh#L69) and on iOS is called only when `MOBILE` is already equal to 1 [scripts/setup.sh#L105](https://github.com/devqore/spacedrive/blob/0b9d46af47e3afbcf06ad1e326b9179aac17ff77/scripts/setup.sh#L105) one check should be enough. But it also means that all android rustup targets are currently installed on the iOS either.

You are right, thanks for the change. The behavior of installing android targets alongside iOS is intentional, because on macOS we don't know if the user wants to develop for android or ios or both

scripts/setup.sh
devqore: The comment changed, sounds much better this way :)

HeavenVolkoff (approved):

PR #1663, opened 2023-10-23
[ENG-1327] Loading indicator if a location is indexing

ameer2468:
Has been modified to when `completed_task_count` is 0 rather than being completed to offer a faster experience and files/folders will just come in.

https://github.com/spacedriveapp/spacedrive/assets/33054370/25d8d29d-3ba3-447e-91aa-3d15e51b7bc9


brxken128 (approved):
i love it, better than a blank screen

PR #1664, opened 2023-10-23
[ENG-1342] Prompt for tagline and category when creating release

Brendonovich:
Refactors the slash command to create a modal that requires inputting both the category and tagline. Will help avoid a situation like last release 💀


brxken128 (approved):

brxken128:
> Refactors the slash command to create a modal that requires inputting both the category and tagline. Will help avoid a situation like last release 💀

Last release was a breeze, idk what nonsense you're talking about

PR #1665, opened 2023-10-23
[ENG-1354] Shortcut to reveal in finder/explorer

ameer2468:
Shortcut to reveal in explorer/finder

**Tested on**: Windows & MacOS

### DIFF
--- a/interface/hooks/useKeyRevealFinder.ts
+++ b/interface/hooks/useKeyRevealFinder.ts
@@ -58,13 +58,9 @@ export const useKeyRevealFinder = () => {
 	useKeys([shortcuts.Meta.key, 'KeyY'], (e) => {
 		e.stopPropagation();
 		if (!revealItems) return;
-		if (explorer.selectedItems.size > 0) {
 			revealItems(
 				library.uuid,
 				items
 			);
-		}
 	});
-
-
 }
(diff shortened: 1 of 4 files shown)


interface/hooks/useKeyRevealFinder.ts
```diff
@@ -0,0 +1,66 @@
+import {useLibraryContext} from '@sd/client';
+import { usePlatform, type Platform } from '~/util/Platform';
+import { useMemo } from 'react';
+import { useKeysMatcher } from './useKeyMatcher';
+import { useKeys } from 'rooks';
+import { useExplorerContext } from '~/app/$libraryId/Explorer/Context';
+
+export const useKeyRevealFinder = () => {
```
Brendonovich, line 8: Since this hook depends on stuff in `~/app/$libraryId/Explorer`, it should live in `/Explorer`. `/hooks` should be reserved for hooks that can be called globally.

Brendonovich (changes requested):

interface/hooks/useKeyRevealFinder.ts
ameer2468: > Since this hook depends on stuff in `~/app/$libraryId/Explorer`, it should live in `/Explorer`. `/hooks` should be reserved for hooks that can be called globally.

You're right! done

Brendonovich (approved):

PR #1670, opened 2023-10-24
Fix mobile onboarding

utkubakir:
~~Default directories are not thing on mobile and the crate we use fails to deserialize or smth.~~

image


Rocky43007:
Broken on Android now with the same error that was plaguing iOS. 

Image:

brxken128:
> Broken on Android now with the same error that was plaguing iOS.

i'm not sure making it optional via serde is the right way @utkubakir, but instead making it an `Option<>` within the `struct` directly, and only running the relevant code if set (i haven't looked into what the library creation function actually does with the arguments, but it should be doable)

i can probably give this a shot quickly. desktop's `core.ts` should probably also be regenerated when a value that's exported via `specta` is altered

oscartbeaumont:
Using `#[specta(default)]` is probably a mistake because it will **only** affect the types, not the runtime deserialization. That rspc error should only occur when a type mismatch happens.

I kinda feel like `#[serde(default)]` is the better way here but `Option` would also work.

brxken128:
>  Using `#[specta(default)]` is probably a mistake because it will **only** affect the types, not the runtime deserialization. That rspc error should only occur when a type mismatch happens.
> 
> I kinda feel like `#[serde(default)]` is the better way here but `Option` would also work.

the only downside with `Option` is that it has to be defined explicitly as null everywhere, and it turns out we use the query quite a lot (as seen by the previous ci run)

utkubakir:
Passing null for an optional value just sounds weird for TS. However you guys are the rust expert and I don't really mind as long as its fixed lol

Not sure how this have to do with android but will investigate.

brxken128:
> Passing null for an optional value just sounds weird for TS. However you guys are the rust expert and I don't really mind as long as its fixed lol
> 
> Not sure how this have to do with android but will investigate.

We can have it conditional in both ways, the type signature will probably be funky but should still be usable regardless - I'll give it a shot shortly, just trying to get the traffic lights back in order (I know _what_ needs to happen, it's just not happening for whatever cursed reason)

fogodev:
You can use `#[specta(optional)]` or `#[serde(optional)]` not too sure which one is the best approach, to not need to define the attribute as null in TS.

brxken128:
> You can use `#[specta(optional)]` or `#[serde(optional)]` not too sure which one is the best approach, to not need to define the attribute as null in TS. 

I reckon `serde` would be the way to go, it'd result in a slightly odd type signature but it should still work the same with both serde and Rust `Option` types

Rocky43007:
> You can use `#[specta(optional)]` or `#[serde(optional)]` not too sure which one is the best approach, to not need to define the attribute as null in TS.

Could `cfg_attr` be used for os specific usage?

For example:
```rs
#[cfg_attr(target_os='android', specta(optional))]
#[cfg_attr(target_os='ios', serde(optional))]
```

brxken128:
> > You can use `#[specta(optional)]` or `#[serde(optional)]` not too sure which one is the best approach, to not need to define the attribute as null in TS.
> 
> Could `cfg_attr` be used for os specific usage?
> 
> For example:
> ```rs
> #[cfg_attr(target_os='android', specta(optional))]
> #[cfg_attr(target_os='ios', serde(optional))]
> ```

There's no need as a sound implementation should not need differing type signatures - it's not Android's specific fault, we just never considered the option of predefined locations being optional (as the code was originally aimed at the desktop, where we can get these locations, however mobile is a little more locked down).

brxken128 (approved):
Windows CI is just broken atm (thanks openssl). Either way, this works great on MacOS - I'm unable to actually test mobile but nothing points towards it breaking so that's a plus haha

PR #1680, opened 2023-10-26
fix: traffic light placement when non-fullscreen

brxken128:
This fixes the padding, animation duration and traffic light positioning in regards to the window being fullscreen/non-fullscreen. It turns out our positioning "method" was also the cause of our original issue - we render an invisible menu bar to pad the traffic lights, but the issue is that this menu bar remained in place even after the app was made fullscreen (via the green traffic light). More info can be found in #1582.

The traffic light placement was something I entirely overlooked as I was too busy focused on fixing the bar, borders and then padding to notice they had inadvertently been misaligned - the cause of the issue was also the reason for their alignment, but luckily the challenges that arose from that PR alone caused this one to be quite a simple fix.

There _is_ some minor visual artifacting and shifting when moving between fullscreen/not fullscreen, I played around with the timings in order to minimize the impact but there's only so much a sliding animation can really do to alleviate the (pretty noticeable) shift.

There was also the issue of the `tauri-drag-region` being locked to a 300ms duration, essentially inhibiting and limiting the other durations defined (this was also fixed).

### DIFF
--- a/apps/desktop/crates/macos/src-swift/window.swift
+++ b/apps/desktop/crates/macos/src-swift/window.swift
@@ -46,13 +46,13 @@ public func blurWindowBackground(window: NSWindow) {
 
 @_cdecl("set_titlebar_style")
 public func setTitlebarStyle(window: NSWindow, fullScreen: Bool) {
-  // this results in far less visual artifacts if we just manage it ourselves (tha native taskbar re-appears when fullscreening/un-fullscreening)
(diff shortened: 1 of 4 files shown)


ameer2468:
As a note, tested and works great. I asked Jake if we could hide the traffic lights during the resizing process to avoid strange movement of the traffic lights when do.

As also Slack does this, and I thought it looked great.

jamiepine (approved):

PR #1690, opened 2023-10-28
Landing page metadata update

brxken128:
Changes:

- Moves the vast majority of transparent images to separate directories (so they're not indexed by search engines)
- Updates all references to those images
- Adds and updates `robots.tsx`, `sitemap.tsx` and `manifest.tsx` for dynamically-generating the related [metadata files](https://nextjs.org/docs/app/api-reference/file-conventions/metadata)
- Removes the old, statically-generated counterparts from `public/`


apps/landing/src/components/AppEmbed.tsx
HeavenVolkoff, line 1: I think this component is not used anywhere in the current landing page, can't we just drop it?

HeavenVolkoff (changes requested):

apps/landing/src/components/AppEmbed.tsx
brxken128: There's a few things here and there that are duplicated or not actually used at all, the footer _component_ being the first I remember - I left the file around but commented as this was more of a subtle refactor and not a major cleanup (and it may even come back in the future? not a clue there though).

Either way, I didn't want to fully remove some code and have that cause a "where's this component??" moment for anyone a few months down the line haha

HeavenVolkoff: I mean, It is in the git history. IMO, if there is a need for this code again in the future, we can just dig it out. I not really a fan of large amounts of comment out code, but your call, just think if you are going to leave the file, and it is not used anywhere, no need to fix it...

apps/landing/src/app/manifest.tsx
```diff
@@ -0,0 +1,21 @@
...
+		start_url: '/',
+		icons: [
+			{
+				src: '/images/logo-192x192.png',
+				sizes: '192x192',
+				type: 'image/png'
+			},
+			{
+				src: '/images/logo-512x512.png',
+				sizes: 'any',
```
utkubakir, line 16: 512x512

apps/landing/src/components/AppEmbed.tsx
utkubakir: yeah agree with @HeavenVolkoff , just remove it

HeavenVolkoff (approved):

PR #1694, opened 2023-10-28
Remove unused dependencies from server

### LINKED ISSUE #1693: Unused dependencies in sd-server crate
### Describe the bug

The `tower-http` and `ctrlc` dependencies in `sd-server` are unused. This slows down build times unnecessarily.

### Reproduction

1. Build the `sd-server` crate.
2. Wait.

### Expected behavior

The only dependencies in `Cargo.toml` and `Cargo.lock` should be ones that contribute to Spacedrive.

### Platform and versions

```Shell
Output of `pnpm --version && cargo --version && rustc --version`:

8.10.0
cargo 1.73.0 (9c4383fb5 2023-08-26)
rustc 1.73.0 (cc66ad468 2023-10-03)
```

OS: Endeavour OS (rolling release)
```


### Stack trace

_No response_

### Additional context

_No response_

julianbraha:
This removes the unnecessary `ctrlc` and `tower-http` dependencies from sd-server, which speeds up build times. 

Closes #1693

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7006,14 +7006,12 @@ name = "sd-server"
 version = "0.1.0"
 dependencies = [
  "axum",
- "ctrlc",
  "http",
  "include_dir",
  "mime_guess",
  "rspc",
  "sd-core",
  "tokio",
- "tower-http",
  "tracing 0.2.0",
 ]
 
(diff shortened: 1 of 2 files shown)


julianbraha:
Seems like the action is failing due to a warning from clippy about an uneeded borrow.

I can push a commit to remove the `&`, but it seems unrelated to this PR.

HeavenVolkoff:
> Seems like the action is failing due to a warning from clippy about an unneeded borrow.

I don't think this is the case, the error is about a missing dependency on Windows required to compile `openssl-sys`. I will solve this in another PR

julianbraha:
> I don't think this is the case, the error is about a missing dependency on Windows required to compile `openssl-sys`. I will solve this in another PR

Oh, you're right. Seems that warning is in the successful clippy runs, and not specific to windows.

oscartbeaumont (approved):

PR #1702, opened 2023-10-30
[ENG-1375] Use selective instance enumeration for mdns

oscartbeaumont:
Changes:
 - [x] Support IPv6 with mDNS
 - [x] Switch to properly using [mDNS selective instance enumeration](https://datatracker.ietf.org/doc/html/rfc6763#section-7.1)
 - [ ] ~~Service metadata is getting mixed up between subservices~~ - Seems this is expected behavior in the spec but I think we can do better than the current solution regardless

These fixes are being tracked upstream [here](https://github.com/keepsimple1/mdns-sd/issues/145) and [here](https://github.com/keepsimple1/mdns-sd/pull/146).

Blocked on #1620 and a fix for the metadata issue.


oscartbeaumont:
Most of the changes in this PR has already been merged into main so gonna close this and finish up the final changes on #1970

PR #1708, opened 2023-10-30
[ENG-1193] Added icons to dialogs

ameer2468:
**Added icons to 3 dialogs**: Delete location, add location, and delete directory/file.

- Corrected Accordion component export
- Added 10 light icons 
- New Icon component in **UI director**y because the Icon component we have in interface is not accessible from **SD/UI**

image


brxken128 (approved):
Nice work!

PR #1710, opened 2023-10-30
[ENG-946] Added zooming functionality for Preview

### LINKED ISSUE #1165: [ENG-946] Zooming functionality for Quick Preview
<sub>From [SyncLinear.com](https://synclinear.com) | [ENG-946](https://linear.app/spacedriveapp/issue/ENG-946/zooming-functionality-for-quick-preview)</sub>

Tilo-K:
## This adds zoom to images in the quick preview.

https://github.com/spacedriveapp/spacedrive/assets/17810881/85027563-dd20-43f1-a7f0-bef97f00a46f

I'm not 100% sure if this is the kind of zoom, which was intended here, but that was my interpretation.

Closes #1165

### DIFF
--- a/interface/app/$libraryId/Explorer/FilePath/Thumb.tsx
+++ b/interface/app/$libraryId/Explorer/FilePath/Thumb.tsx
@@ -351,7 +351,8 @@ const Thumbnail = memo(
 		...props
 	}: ThumbnailProps) => {
 		const ref = useRef<HTMLImageElement>(null);
-
+		const [zoomed, setZoomed] = useState(false);
+		const [mousePos, setMousePos] = useState({ x: 0, y: 0 });
 		const size = useSize(ref);
 
 		const { style: blackBarsStyle } = useBlackBars(size, blackBarsSize);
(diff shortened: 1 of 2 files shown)


jamiepine:
I would prefer if this be not activated by clicking the image, maybe a button in the top right or keybind. Double click also a possibility, but not with the magnification cursor.

utkubakir:
> I would prefer if this be not activated by clicking the image, maybe a button in the top right or keybind. Double click also a possibility, but not with the magnification cursor.

Maybe scrollbar and click-hold to move around?

Tilo-K:
I changed the single click to a double click and removed the cursor. Also, I added an Icon in the top right which effectively does the same as a double click.


https://github.com/spacedriveapp/spacedrive/assets/17810881/c4c07576-efd6-4d81-92f3-eb43bb49aa79

Is this enough or should I explore the option of using scrollbars, I'm not sure If they would feel good to use, but that's up to you.

utkubakir:
> I changed the single click to a double click and removed the cursor. Also, I added an Icon in the top right which effectively does the same as a double click.
> 
>  2023-11-02.13-10-45.mp4 
> Is this enough or should I explore the option of using scrollbars, I'm not sure If they would feel good to use, but that's up to you.

You are right we should reserve scroll action for moving around the zoomed image like up and down and horizontal scroll view for right and left. I believe that would align with how trackpads work

PineappleRind:
This is how Preview/QuickLook do it:

use `Cmd``+` / `Cmd``-` to zoom in/out, along with these buttons -
image

to move around the zoomed image, there are scrollbars on the relevant axes

interface/app/$libraryId/Explorer/FilePath/Thumb.tsx
```diff
@@ -351,11 +354,16 @@ const Thumbnail = memo(
...
-
+		const [zoomed, setZoomed] = useState(false);
+		const [mousePos, setMousePos] = useState({ x: 0, y: 0 });
 		const size = useSize(ref);
 
 		const { style: blackBarsStyle } = useBlackBars(size, blackBarsSize);
 
+		useEffect(() => {
+			setZoomed(!!props.zoomed);
+		}, [props.zoomed]);
```
utkubakir, line 365: Can we get rid of this somehow? You can use `quickPreviewStore` if you need to pass things through components.

utkubakir (changes requested):

utkubakir:
> This is how Preview/QuickLook do it:
> 
> use `Cmd` `+` / `Cmd` `-` to zoom in/out, along with these buttons - image
> 
> to move around the zoomed image, there are scrollbars on the relevant axes

I think it should be more finder, zooming in or out several times to the center is great

utkubakir:
@Tilo-K are you gonna continue with this pr?

Issue #1721, opened 2023-11-02
Error logged when reading metadata on on-demand files


Carterpersall:
### Describe the bug

- When Spacedrive attempts to read the metadata from an on-demand file, an error is printed to the log
  - On-Demand files allows OneDrive to download files as they are needed by programs
  - Error only occurs if OneDrive is not running, otherwise the file will be downloaded (which is also not good)
  - Error is logged both during the indexing process, and every time the directory containing the file is opened
  - On-Demand files are a built-in feature of NTFS
Example:
```
2023-11-01T19:29:02.533749Z ERROR sd_core::object::file_identifier: Failed to extract file metadata: FileIOError {
    path: "C:\\Users\\xxxxx\\OneDrive\\Document.docx",
    source: Os {
        code: 362,
        kind: Uncategorized,
        message: "The cloud file provider is not running.",
    },
    maybe_context: None,
}
```

### Reproduction

1. Close OneDrive
2. Add directory containing on-demand files to Spacedrive (Opening an already indexed location works too)
3. One of the following occurs:
  - If OneDrive is open, the file is downloaded
  - If OneDrive is closed, an error is printed to the log

### Expected behavior

- Each file should be checked for the `FILE_ATTRIBUTE_RECALL_ON_DATA_ACCESS` attribute. And if it has the attribute, it should be handled in a way that doesn't trigger a download or the error.

### Platform and versions

```Shell
`pnpm --version`: `8.10.0`
`cargo --version`: `cargo 1.73.0 (9c4383fb5 2023-08-26)`
`rustc --version`: `rustc 1.73.0 (cc66ad468 2023-10-03)`
`Get-ComputerInfo | Select-Object OsName,OsVersion,CsModel | Format-List`:

OsName    : Microsoft Windows 11 Enterprise
OsVersion : 10.0.22621
CsModel   : Blade 14 - RZ09-0370
```
```


### Stack trace

_No response_

### Additional context

_No response_

Carterpersall:
Seems that the error occurs when attempting to generate a hash for the file, which obviously won't work if the file is "offline". A possible solution would be to skip hashing the file. Maybe instead, a hash could be created using the file's metadata? That could be used to detect whether an offline file has been changed, even if it remains "offline".

akskan160103:
Hey @Carterpersall  ,

[...]

PR #1722, opened 2023-11-02
Clarify error that occurs when getting the hash of an on-demand file

Carterpersall:
- Clarifies error that occurs when attempting to read the metadata from an on-demand file
  - Involves issue #1721 but does not solve it

Current Error:
```
2023-11-01T19:29:02.533749Z ERROR sd_core::object::file_identifier: Failed to extract file metadata: FileIOError {
    path: "C:\\Users\\xxxxx\\OneDrive\\Document.docx",
    source: Os {
        code: 362,
        kind: Uncategorized,
        message: "The cloud file provider is not running.",
    },
    maybe_context: None,
}
```
New Error:
```
2023-11-01T15:58:27.994181Z ERROR sd_core::object::file_identifier: Failed to extract metadata from on-demand file: FileIOError {
    path: "C:\\Users\\xxxxx\\OneDrive\\Document.docx",
    source: Os {
        code: 362,
        kind: Uncategorized,
        message: "The cloud file provider is not running.",
    },
    maybe_context: None,
}
```

### DIFF
--- a/core/src/object/file_identifier/mod.rs
+++ b/core/src/object/file_identifier/mod.rs
@@ -124,7 +124,17 @@ async fn identifier_job_step(
 							(metadata, file_path),
 						)
 					})
-					.map_err(|e| error!("Failed to extract file metadata: {e:#?}"))
+					.map_err(|e| {
+						if e.source
+							.raw_os_error()
+							.map(|code| code == 362)
+							.unwrap_or(false)
+						{
+							error!("Attempted to extract metadata from on-demand file: {e:#?}");
+						} else {
+							error!("Failed to extract file metadata: {e:#?}")
+						}
+					})
 					.ok()
 			}),
 	)
--- a/core/src/object/file_identifier/mod.rs
+++ b/core/src/object/file_identifier/mod.rs
@@ -130,7 +130,7 @@ async fn identifier_job_step(
 							.map(|code| code == 362)
 							.unwrap_or(false)
 						{
-							error!("Attempted to extract metadata from on-demand file: {e:#?}");
+							error!("Failed to extract metadata from on-demand file: {e:#?}");
 						} else {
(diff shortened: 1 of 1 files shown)


core/src/object/file_identifier/mod.rs
```diff
@@ -124,7 +124,17 @@ async fn identifier_job_step(
...
+					.map_err(|e| {
+						if e.source
+							.raw_os_error()
+							.map(|code| code == 362)
+							.unwrap_or(false)
+						{
+							error!("Failed to extract metadata from on-demand file: {e:#?}");
+						} else {
+							error!("Failed to extract file metadata: {e:#?}")
+						}
```
fogodev, line 136: As this is a specific windows error, can you put this part in a conditional compilation section. Something like 

```rust
#[cfg(target_os = "windows")]
{
    if e.source
        .raw_os_error()
        .map_or(false, |code| code == 362)
    {
        error!("Failed to extract metadata from on-demand file: {e:#?}");
    } else {
        error!("Failed to extract file metadata: {e:#?}");
    }
}

#[cfg(not(target_os = "windows"))]
{
    error!("Failed to extract file metadata: {e:#?}");
}
```

The `map` and `unwrap_or` can also be simplified on a `map_or`

fogodev (changes requested):
Sorry for the delay reviewing this one, went totally under my radar.

core/src/object/file_identifier/mod.rs
Carterpersall: >can you put this part in a conditional compilation section
Sure, I'm not sure why I didn't do this in the first place.

>The `map` and `unwrap_or` can also be simplified on a `map_or`
I have no idea why I didn't know `map_or` was a thing, that's much better than what I had

fogodev (approved):

PR #1728, opened 2023-11-04
Remove < 21 node restriction

fogodev:
Removing the restriction as it isn't needed anymore due to https://github.com/dart-lang/sdk/issues/53784 being solved


HeavenVolkoff (approved):
LGTM

PR #1749, opened 2023-11-06
Fix search bar

utkubakir:
Something was wrong with style caching so the search bar was white on the first-load interestingly, this pr fixes that issue


brxken128 (approved):
Pretty odd and good catch!

PR #1750, opened 2023-11-06
[ENG-1406] Fix search bar take 2

utkubakir:
https://github.com/vercel/next.js/issues/13092 =_= this should do it


brxken128 (approved):
Can confirm this works, for some reason it worked fine on my MacBook beforehand but after testing on my phone a minute ago I can see the issue - this does resolve it though, nice job! 🙏

PR #1751, opened 2023-11-06
[ENG-1360] Not always unlink objects from file_paths on update

fogodev:
This PR introduces some changes trying to preserve an object from being unlinked on file_path updates. Also some watcher fixes and improvements.


jamiepine (approved):
This is important.

PR #1761, opened 2023-11-09
Update tokio's version

fogodev:
Today's Tokio release include this fix (https://github.com/tokio-rs/tokio/issues/6133), which can potentially explain some mysterious jobs hanging that some users reported to us at the first alpha launch.


HeavenVolkoff (approved):
LGTM

PR #1766, opened 2023-11-10
[ENG-1359] Revert "Fix pdf thumb crashing app (#1460)"

HeavenVolkoff:
This reverts the changes made in PR #1460. It ended up being an unnecessary set of changes that did not fix the underlying issue and made PDF thumbnail performance way worse. The correct fix for the mentioned issue was implemented in https://github.com/bblanchon/pdfium-binaries/pull/138. Also updated `pdfium-render`.


utkubakir (approved):

brxken128:
Was a tricky one to solve but glad we got there in the end! 🎉

PR #1771, opened 2023-11-14
Fix closing tag window

### LINKED ISSUE #1769: "String must contain..." when closing tag window
### Describe the bug

When I open the "New Tag" Window and click on "Close" without any cahnges made, I get the warning "String must contain at least 1 character(s)". Clicking on "Close" a second time actually closes the window.

### Reproduction

_No response_

### Expected behavior

I would expect to for the window to be closed without a warning.

### Platform and versions

```Shell
Spacedrive for macOS v0.1.2 - 83538f2
```

### Stack trace

_No response_

### Additional context

_No response_

utkubakir:
Fixes #1769

### DIFF
--- a/interface/app/$libraryId/settings/library/tags/CreateDialog.tsx
+++ b/interface/app/$libraryId/settings/library/tags/CreateDialog.tsx
@@ -53,8 +53,7 @@ export default (
 
 	const form = useZodForm({
 		schema: schema,
-		defaultValues: { color: '#A717D9' },
-		mode: 'onSubmit'
+		defaultValues: { color: '#A717D9' }
 	});
 
 	const createTag = useLibraryMutation('tags.create');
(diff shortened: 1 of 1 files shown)


interface/app/$libraryId/settings/library/tags/CreateDialog.tsx
```diff
@@ -54,7 +54,7 @@ export default (
 	const form = useZodForm({
 		schema: schema,
 		defaultValues: { color: '#A717D9' },
-		mode: 'onBlur'
+		mode: 'onSubmit'
```
niikeec, line 57: We don't really have to use mode here as it's "onSubmit" by default

niikeec (changes requested):

niikeec (approved):

PR #1802, opened 2023-11-21
[ENG-433] removeTab on middle click

Brendonovich:


utkubakir (approved):
ez

Brendonovich:
i don't think so? there shouldn't be anything to prevent

PR #1808, opened 2023-11-23
Thumbnailer channel with capacity 0 bug

fogodev:
This PR introduce a check to skip batches with size 0 erroneously sent to be processed


HeavenVolkoff (approved):
LGTM

Issue #1828, opened 2023-11-25
Unable to add a space when renaming an item


kassah:
### Describe the bug

When renaming a folder or photo when entering the new name, if you hit space bar to add a space, the preview pops up canceling the rename operation.

### Reproduction

1. Go to a folder with a subfolder under it.
2. Right click the subfolder and select "Rename"
3. Use arrow keys to get to the middle of the name
4. Press spacebar to add a space in it.
5. Observe that preview popped and canceled the rename

Alternatively for a photo
1. Go to a folder with a subfolder under it with a photo within it.
2. Go into the subfolder
3. Right click the photo and select "Rename"
4. Use arrow keys to get to the middle of the name
5. Press spacebar to add a space in it.
6. Observe that preview popped and canceled the rename

### Expected behavior

If you're using the behavior of macOS Finder as an example, then a spacebar open of preview would only work on a item selection, but not when the context menu or rename are active.

### Platform and versions

```Shell
Spacedrive DMG (ARM) install Version 0.1.3 (20231123.222549) on MacOS Sonoma 14.1.1 (on M1 Pro)

System versions (not used for compilation):
pnpm --version && cargo --version && rustc --version
7.0.1
cargo 1.60.0 (d1fd9fe2c 2022-03-01)
rustc 1.60.0 (7737e0b5c 2022-04-04)
```

### Stack trace

```Shell
No stack trace
```

### Additional context

Attempted in List & Grid view. Gallery view does not have the rename item in the context menu.

PR #1836, opened 2023-11-26
[ENG-1466] dont open quick preview while renaming

ameer2468:
Shortcut was defined in the wrong place, we don't want to open quick preview if the user is renaming an object.


utkubakir:
Fixes #1828

PR #1842, opened 2023-11-29
Cloud connection

Brendonovich:
Introduces the ability to link a library to our cloud and the join it from another node.
This is all still feature flagged since not all the data we need is being uploaded, but it's a start for working on sync message exchanging.

### DIFF
--- a/packages/client/src/hooks/useClientContext.tsx
+++ b/packages/client/src/hooks/useClientContext.tsx
@@ -10,38 +10,38 @@ const libraryCacheLocalStorageKey = 'sd-library-list';
 export const useCachedLibraries = () =>
 	useBridgeQuery(['library.list'], {
 		keepPreviousData: true,
-		initialData: () => {
-			const cachedData = localStorage.getItem(libraryCacheLocalStorageKey);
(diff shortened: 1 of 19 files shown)


packages/client/src/hooks/useClientContext.tsx
```diff
@@ -10,38 +10,38 @@ const libraryCacheLocalStorageKey = 'sd-library-list';
...
-		}
-	}
+	// if (cachedData) {
+	// 	// If we fail to load cached data, it's fine
+	// 	try {
+	// 		return JSON.parse(cachedData) as LibraryConfigWrapped[];
+	// 	} catch (e) {
+	// 		console.error("Error loading cached 'sd-library-list' data", e);
+	// 	}
+	// }
```
utkubakir, line 40: Is this intentional?

Brendonovich: kind of - i was getting some weird behaviour with `@sd/web` and wanted to try disabling library caching, turns out it works great. might be a change worth saving for another PR tho

oscartbeaumont (approved):

PR #1856, opened 2023-12-05
Cloud Sync Basics

Brendonovich:
Adds actors for sending, receiving, and ingesting sync messages. None of this is guaranteed to work, but the pieces are in place.


apps/desktop/src-tauri/src/main.rs
```diff
@@ -174,7 +174,8 @@ async fn main() -> tauri::Result<()> {
 			Node::new(
 				data_dir,
 				sd_core::Env {
-					api_url: "https://app.spacedrive.com".to_string(),
+					// api_url: "https://app.spacedrive.com".to_string(),
+					api_url: "http://localhost:3000".to_string(),
```
utkubakir, line 178: why api url is localhost?

Brendonovich: my b

utkubakir (dismissed):
not sure about rust but the rest looks good to me

oscartbeaumont (approved):

PR #1859, opened 2023-12-05
[ENG-1490] List view: Quick preview image slider performance

ameer2468:
After debugging, this un-necessary code was causing re-renders. Should have been removed from before.


brxken128 (approved):
performance optimisations are always a massve W

PR #1871, opened 2023-12-07
[ENG-1493] Fix navigation buttons with @remix-run/router patch

Brendonovich:
Fixes our in-app navigation buttons by [patching](https://github.com/remix-run/react-router/pull/11100) `@remix-run/router` so that we can use `replace` in `redirect`.


oscartbeaumont (approved):

Brendonovich:
hold up the patch is way too big

oscartbeaumont (approved):

PR #1888, opened 2023-12-13
Basic actor manager

Brendonovich:
Adds a per-library actor storage, where actors can be declared by name + implementation and then started + stopped at will.


brxken128 (approved):
Code looks good, app builds and works fine ~~however I'm not _too_ sure how to actually test the functionality lol~~

Tested the functionality, it starts actors successfully. If they panic the app locks up, and eventually closes (or gets killed by the OS, I'm not too sure) but it's hidden behind the debug menu and only experimental so I assume it's fine? Probably good to go imo 🙏🏻

PR #1890, opened 2023-12-13
[ENG-1503] Close search bar when navigating into folder

Brendonovich:
A couple of things are happening here: Search and filters are being stored in search params so they save across navigations, and the entire state of the location explorer is being reset when navigating to a different folder by keying it on `path`.

### DIFF
--- a/interface/app/$libraryId/Explorer/View/ViewItem.tsx
+++ b/interface/app/$libraryId/Explorer/View/ViewItem.tsx
@@ -1,5 +1,9 @@
 import { useCallback, type HTMLAttributes, type PropsWithChildren } from 'react';
-import { createSearchParams, useNavigate } from 'react-router-dom';
+import {
+	createSearchParams,
+	useNavigate,
+	useSearchParams as useRawSearchParams
+} from 'react-router-dom';
 import {
 	isPath,
 	useLibraryContext,
@@ -24,6 +28,7 @@ export const useViewItemDoubleClick = () => {
 	const explorer = useExplorerContext();
 	const { library } = useLibraryContext();
 	const { openFilePaths, openEphemeralFiles } = usePlatform();
+	const [_, setSearchParams] = useRawSearchParams();
 
 	const updateAccessTime = useLibraryMutation('files.updateAccessTime');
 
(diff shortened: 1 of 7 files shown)


interface/app/$libraryId/tag/$id.tsx
```diff
@@ -82,14 +82,18 @@ export function Component() {
...
 			</SearchContextProvider>
-			<Explorer
-				emptyNotice={
-					<EmptyNotice
-						icon={<Icon name="Tags" size={128} />}
-						message="No items assigned to this tag."
-					/>
-				}
-			/>
+			<MemoisedExplorer />
```
niikeec, line 85: Memoing explorer doesn't do much here, since we use the explorer context in the component which causes a rerender each time the context value changes which is for each rerender.

interface/app/$libraryId/location/$id.tsx
```diff
@@ -230,14 +174,7 @@ const LocationExplorer = ({ location }: { location: Location; path?: string }) =
...
 			) : !preferences.isLoading ? (
-				<Explorer
-					emptyNotice={
-						<EmptyNotice
-							icon={<Icon name="FolderNoSpace" size={128} />}
-							message="No files found here"
-						/>
-					}
-				/>
+				<MemoisedExplorer />
```
niikeec, line 177: Same as for tag explorer

interface/app/$libraryId/Explorer/View/ViewItem.tsx
```diff
@@ -110,11 +115,16 @@ export const useViewItemDoubleClick = () => {
...
-							path: `${item.materialized_path}${item.name}/`
-						}).toString()
+					setSearchParams((p) => {
+						const newParams = new URLSearchParams();
+
+						newParams.set('path', `${item.materialized_path}${item.name}/`);
+						const take = p.get('take');
+						if (take !== null) newParams.set('take', take);
+
+						console.log([...newParams]);
```
niikeec, line 125: console.log()

niikeec (changes requested):

interface/app/$libraryId/tag/$id.tsx
Brendonovich: I do it because search state is hoisted, there's no reason the explorer should rerender when search state updates. It'll update on its own when the items change

Brendonovich: yeah good point actually

niikeec (approved):

PR #1897, opened 2023-12-18
[ENG-1502] I18n

utkubakir:


Brendonovich:
The `.json` locale files are going to need to be moved somewhere else - probably `@sd/interface` - since `@sd/web` needs to serve them too.

Brendonovich (approved):

PR #1893, opened 2023-12-14
[ENG-1508] Launch Spacedrop

oscartbeaumont:
Launching Spacedrop
 - [x] Remove feature flags
 - [x] Add Spacedrop to menu bar
 - [x] Working drag and drag position - #1899
 - [x] Drag and drop from Finder to menu bar icon
 - [x] Temporarily fix mdns errors
 - [x] Animation to guide the user towards dropping the file on the Spacedrop icon
 - [x] Autoclose menubar popover when using DND
 - [x] Click on the device to start upload (`openFilePickerDialog` isn't supported on web cause it would give us the file not the path on the device).

PR #1899, opened 2023-12-19
Tauri File Drop Events

oscartbeaumont:
We move to a [Tauri fork](https://github.com/spacedriveapp/tauri/tree/main2) and [Tao fork](https://github.com/spacedriveapp/tao/tree/plz-work) to get expose the internal `cursor_position` method. This cursor position is in the same "plane" as the window position which is super important for subtracting them and I could not find a mouse position library that could maintain that requirement.

This gives us an allusion of having the `position` in the `FileDropEvent` but also allows us to real-time track `FileDropEvent:: Hovered`'s `position` which not even Tauri v2 supports at the moment.

We originally changed Gtk versions under the hood but I could not get CI working for this so I managed to workout a way to avoid it. This shouldn't break anything but the Tao versions are wack so there is a small chance of an issue with it coming in but things seem fine on macOS.

This PR also forks `tauri-plugin-window-state`. I literally made zero modifications to its source code but for some reason, the app hangs on startup if we use the crates.io release.

<del>
Move to a fork of Tauri. The fork is based on Tauri v1.0.5 but uses Wry version ~~0.26~~ 0.27.

This was and never will be a recommended combination from the Tauri team but I got it working.

I tried upgrading to Tauri v2-rc.x originally but I think it will take a super significant amount of work so I don't think it's the path forward at the moment as they have made a lot of breaking changes.

This needs cross-platform testing and if it passes on all OS's we should be good to go:
 - [x] macOS
 - [ ] Windows
 - [ ] Linux (non-Wayland)
 - [ ] ~~Linux (Wayland)~~ - Will not work due to `cursor_position` [not being supported](https://docs.rs/tao/latest/tao/event_loop/struct.EventLoopWindowTarget.html#method.cursor_position)
</del>


oscartbeaumont:
This PR has the coordinate the origin (`(0, 0)`) super off the webview. That is fixed by some extra Tao patching in #1893 .

oscartbeaumont:
This has been moved into #1893

PR #1916, opened 2024-01-05
Remove `useExplorerStore`

oscartbeaumont:
Right now we have a `useExplorerState` hook which subscribes to the whole Valtio store. This PR removes that hook and introduces `useSelector`.

When `useSelector` is used with a Valtio store the callback function will run for each change to the store but only if the result changes will the component subscribing actually be rerendered.

We also use the npm package `fast-equals` to help us easily maintain stable references and ensure we only actually `setState` when the value changes.

This PR also sets the `displayName`  on many memoised components to help with debugging in React devtools.

There is a couple of usages of Valtio which are not using `useSnapshot` when they should but I copied what we previously had incase we were relying on it’s “incorrect” behaviour.

Note: We previously had issues with cross-package Valtio hence the previous design but i'm not seeing the same issues on my end so hopefully whatever the original problem was has been fixed in Valtio.

### DIFF
--- a/interface/app/$libraryId/Explorer/index.tsx
+++ b/interface/app/$libraryId/Explorer/index.tsx
@@ -49,7 +49,6 @@ export default function Explorer(props: PropsWithChildren<Props>) {
 			console.error('Error in RSPC subscription new thumbnail', err);
 		},
 		onData: (thumbKey) => {
-			console.log('HERE', thumbKey);
 			explorerStore.addNewThumbnail(thumbKey);
 		}
 	});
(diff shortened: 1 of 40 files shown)


interface/app/$libraryId/Explorer/index.tsx
```diff
@@ -45,6 +49,7 @@ export default function Explorer(props: PropsWithChildren<Props>) {
 			console.error('Error in RSPC subscription new thumbnail', err);
 		},
 		onData: (thumbKey) => {
+			console.log('HERE', thumbKey);
```
brxken128, line 52: Do we still need this?

oscartbeaumont: Whoops

brxken128 (approved):
Have tested the app, works great - additionally the changes look good to me and I don't see any glaring issues. Nice job! One step closer to getting things more optimised 🎉

ameer2468 (approved):

Brendonovich:
i'm not sure `useSelector` does anything since `useSnapshot` already tracks what values are accessed but it doesn't hurt

oscartbeaumont:
It does allow the computation prior to update but yeah for most cases you are correct.

PR #1919, opened 2024-01-08
SolidJS Build Infrastructure

oscartbeaumont:
This PR configures Vite to allow React and SolidJS JSX transforms within a single package.

This is done through the use of the `.solid.tsx` file extension to dictate a file should be treated as SolidJS.

This PR also configures ESLint to detect common SolidJS problems such as destructuring. This should help us prevent bugs making it through, especially during the migration when we will be context switching between frameworks a lot.

HMR works on all files bar the entrypoint which is exactly the same behaviour as React, although will likely be a bit annoying during the migration.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -81,3 +81,4 @@ spacedrive
 .cargo/config
 .cargo/config.toml
 .github/scripts/deps
+.vite-inspect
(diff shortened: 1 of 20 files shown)


Brendonovich (dismissed):
brilliant

Brendonovich (approved):

PR #1926, opened 2024-01-10
SolidJS explorer - Part 1

oscartbeaumont:
WIP


jamiepine:
I don't think we should actually PR any Solid stuff, it needs to stay 100% experimental for now until we can properly compare, so just keep this on a branch if you can.

PR #1930, opened 2024-01-11
Use Offset pagination

Brendonovich:
As requested by @jamiepine 
Duplicates the path and object infinite queries into versions that use offset pagination instead of cursor pagination.


Brendonovich:
closing in favour of graphite pr

PR #1932, opened 2024-01-12
Fix QuickPreview video playback

oscartbeaumont:
Change the URL getter to use consistent hashing.

This means the request for a resource will always be routed to the same backend webserver.

We use multiple webserves to ensure we aren't hitting HTTP connection limits when loading all the images in the explorer.


utkubakir (approved):
lgtm

PR #1933, opened 2024-01-12
[ENG-1535] - Remove title

ameer2468:
This resolves the title issue, confirmed and tested on a locally built version of SD.

The Tauri API clearly specifies that hiddenTitle would hide the title on MacOS, but this isn't working. Setting the title to an empty string does it.


brxken128:
I've spoken to @ameer2468, and I'll close this in favour of my new branch which fixes it :)

PR #1945, opened 2024-01-16
[ENG-1540] Remove `localStorage` if no libraries are created

brxken128:
This will 1) reset the onboarding progress on each app load, so that prior information isn't retained and 2) reduce the amount of app crashes/blank pages that happen due to invalid `localStorage` configurations.

For web, I think this will have to be handled in the TS side of things, but for desktop we can do this. I'm not sure if the TS approach will be suitable for desktop also, we can probably do some testing in due course (this works great though).

I've tested this on Linux so far and it works, and I don't see why the other platforms wouldn't work either, but I will be testing shortly just to validate.

### DIFF
--- a/apps/desktop/src-tauri/src/main.rs
+++ b/apps/desktop/src-tauri/src/main.rs
@@ -182,7 +182,7 @@ async fn main() -> tauri::Result<()> {
 	let (_guard, result) = match Node::init_logger(&data_dir) {
 		Ok(guard) => (
 			Some(guard),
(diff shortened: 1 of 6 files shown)


brxken128:
> We should probably delete the cache directory for Spacedrive when triggering this cleanup behavior. Because, at least on Linux, it also contains WebKit files

That's a good idea - I'll add it! I wasn't too sure how fragmented it was on each platform.

ameer2468 (approved):

ameer2468 (approved):

PR #1960, opened 2024-01-20
Use PathBuf instead of String for paths

### LINKED ISSUE #1959: Strings are used for paths instead of PathBuf
### Describe the bug

`PathBuf` is intended to be a flexible path type, and is similar to `String`.

### Reproduction

_No response_

### Expected behavior

_No response_

### Platform and versions

```Shell
8.14.1
cargo 1.75.0 (1d8b05cdd 2023-11-20)
rustc 1.75.0 (82e1608df 2023-12-21)
```

### Stack trace

_No response_

### Additional context

_No response_

julianbraha:
There are two instances in `core/src/api/backups.rs` where `String` is used as a path type, when `PathBuf` could be used instead. This PR makes the replacement.

Closes #1959

### DIFF
--- a/core/src/api/backups.rs
+++ b/core/src/api/backups.rs
@@ -121,29 +121,25 @@ pub(crate) fn mount() -> AlphaRouter<Ctx> {
 				)
 		})
 		.procedure("restore", {
-			R
-				// TODO: Paths as strings is bad but here we want the flexibility of the frontend allowing any path
-				.mutation(|node, path: String| async move {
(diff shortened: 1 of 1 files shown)


brxken128:
Going to give this a test quickly - it looks great, and I'm not too sure what the code comment about flexibility was referring to honestly.

brxken128 (approved):
This works and I'm happy with it - I'm still a little unsure about the string usage in the first place but if that flexibility is important we can always revert this. Thank you for the PR!

PR #1972, opened 2024-01-22
Fix app from not starting due to missing `solid-refresh` and bump version to `0.2.0`

brxken128:
On launch (at least on MacOS), the app will produce an error regarding Solid, Vite and `solid-refresh`. Navigating to `interface`, and running `pnpm i solid-refresh` fixes the issue.

I have tested this on a fresh clone, and it had the error, but this resolves the issue.

### DIFF
--- a/apps/desktop/src-tauri/Cargo.toml
+++ b/apps/desktop/src-tauri/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "sd-desktop"
-version = "0.1.4"
+version = "0.2.0"
 description = "The universal file manager."
 authors = ["Spacedrive Technology Inc <support@spacedrive.com>"]
 default-run = "sd-desktop"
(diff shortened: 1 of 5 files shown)


ameer2468 (dismissed):

ameer2468 (approved):

apps/desktop/src-tauri/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "sd-desktop"
-version = "0.1.4"
+version = "0.2.0"
```
oscartbeaumont, line 3: Was this intentional?

oscartbeaumont: Wait no you say it in the title. Just had read the description.

brxken128: Yeah, we chose to just include the bump in here and to `0.2.0` as it was about to be merged anyway

Brendonovich: sd-core also needs a version bump

Brendonovich (approved):

PR #2000, opened 2024-01-25
Rust 1.75, Tauri 1.5.4 and major dependency update

brxken128:
This extends upon @julianbraha's work and previous PR.

This:
- Resolves `RUSTSEC-2024-0006` and `RUSTSEC-2024-0003`, caused by outdated versions of `shlex` and `h2`, respectively (thank you @julianbraha!)
- Updates Tauri to `1.5.4`
- Updates our Tauri git patch and rebases it onto `1.5.4`
- Updates the vast, vast majority of our Rust dependencies
- Updates the project's default Rust version to `1.75`

The main dependencies I couldn't update were `hyper`, `axum` and `ort`. I got 99% of the way there with `hyper`/`axum` (I still have the stash for when it becomes possible - there were a LOT of HTTP `Body`s that needed to be changed and optimised), but `rspc` is using an older version of `axum` (via `httpz`) so the endpoint wouldn't translate over within `sd-server` correctly. I'd be more than happy to upgrade `axum` and `hyper` once `rspc` supports the newer version.

I have tested this on MacOS and it builds successfully, but of course this will require some very thorough testing across all 3 platforms, and across Linux' AppImage _and_ `.deb` builds.

### DIFF
--- a/.github/actions/setup-rust/action.yaml
+++ b/.github/actions/setup-rust/action.yaml
@@ -16,7 +16,7 @@ runs:
       uses: dtolnay/rust-toolchain@stable
       with:
         target: ${{ inputs.target }}
-        toolchain: '1.73'
+        toolchain: '1.75'
         components: clippy, rustfmt
 
     - name: Cache Rust Dependencies
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -74,7 +74,7 @@ To run the landing page:
 
 If you encounter any issues, ensure that you are using the following versions of Rust, Node and Pnpm:
 
-- Rust version: **1.73.0**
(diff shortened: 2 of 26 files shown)


brxken128:
I'm closing this in favour of a re-based branch that's more up-to-date.

brxken128:
> @brxken128 Can you upgrade axum and hyper to the latest version

They're specifically ones I can't upgrade as that'd require refactoring/plain rewriting a good amount amount of code (that I'm not familiar with at all), as well as maybe doing the same for a few of our dependencies.

I'd love to, as they offer some great performance and security enhancements, but it's likely not worth the time at the minute.

They'll be upgraded eventually, just not yet.

PR #2020, opened 2024-01-30
[MOB-52] Tabs animation

ameer2468:
This PR adds tab animations using https://rive.app/

Have also corrected some css classes for compilation purposes.

https://github.com/spacedriveapp/spacedrive/assets/33054370/3b82355c-8716-4873-bedc-57c73181499e


apps/mobile/src/navigation/TabNavigator.tsx
```diff
@@ -34,72 +106,62 @@ export default function TabNavigator() {
...
+			{TabScreens.map((screen, index) => (
+				<Tab.Screen
+					key={screen.name + index}
+					name={screen.name}
+					component={screen.component}
+					options={{
+						tabBarLabel: screen.label,
+						tabBarLabelStyle: screen.labelStyle,
+						tabBarIcon: () => (
+							<TouchableOpacity activeOpacity={1}>{screen.icon}</TouchableOpacity>
```
utkubakir, line 118: `<Pressable>` is usually better

apps/mobile/src/navigation/TabNavigator.tsx
```diff
@@ -34,72 +106,62 @@ export default function TabNavigator() {
...
+						)
+					}}
+					listeners={() => ({
+						tabPress: () => {
+							setActiveIndex(index);
+						}
+					})}
```
utkubakir, line 125: There is a way to get the focused state of a tab, see [here](https://reactnavigation.org/docs/tab-based-navigation/). It can save us from a unnecessary state and the listener.

utkubakir (changes requested):

utkubakir (approved):

PR #2072, opened 2024-02-08
Added Belarusian localization

artsiom-voitas:
* Added Belarusian localization
* Added entries in settings.json and general.tsx


utkubakir (approved):
выглядае добра, дзякуй!

PR #2093, opened 2024-02-14
[ENG-1321] Webview reload keybind change and make MacOS reload explorer menu item functional

brxken128:
This PR allows users to initiate a quick rescan of a location via the MacOS menu item, and it changes the reload webview keybind to be CMD/CTRL+SHIFT+R (so it's less complex than before).


interface/hooks/useKeybindEventHandler.ts
```diff
@@ -16,6 +18,9 @@ export const useKeybindEventHandler = (libraryId?: string) => {
 				case 'open_settings':
 					libraryId && navigate(`/${libraryId}/settings/client/general`);
 					break;
+				case 'reload_explorer':
+					rescan();
```
niikeec, line 22: Don't think this will run a rescan as it won't have the location id.

brxken128: It does run a rescan, and afaict it doesn't run twice. This just means the menu item activates, as well as being able to activate it *via* the menu bar item.

I'll test to see if it does run twice, and disable the in-app one if that's the case (as this one is more suited towards MacOS).

ameer2468 (approved):

PR #2096, opened 2024-02-14
New Android Build Script

Rocky43007:
When building for Android, many errors pop up when using Rust's cargo build with Gradle Daemons. Errors ranging from building but not uploading the latest `.so` file to JNI Merging issues. Therefore, I moved the build script away from `org.mozilla.rust-android-gradle:plugin:0.9.3` into a custom `.sh` build script like the iOS `build_rust.sh` script. 

The following changes are happening to `setup.sh` due to the introduction of `cargo ndk` into the build so that we can build for the android build targets. 

I've tested it on my android location-watcher testing branch, so please let me know if any issues occur on your own machine. However, from my testing, I have found this new script to be more consistent due to it working directly from the File System instead of through Gradle.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,6 +1,8 @@
 # built product/cache
 target/
 dist/
+apps/mobile/android/app/build
+apps/mobile/modules/sd-core/android/build
 
 # macOS/iOS product/cache
 .build/
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -86,6 +86,8 @@ Make sure to read the [guidelines](https://spacedrive.com/docs/developers/prereq
 
 To run the mobile app:
 
+- Install Java JDK <= 17 for Android
+  - Java 21 is not compatible: https://github.com/react-native-async-storage/async-storage/issues/1057#issuecomment-1925963956
 - Install [Android Studio](https://developer.android.com/studio) for Android and [Xcode](https://apps.apple.com/au/app/xcode/id497799835) for iOS development.
 - Run `./scripts/setup.sh mobile`
   - This will set up most of the dependencies required to build the mobile app.
--- a/apps/mobile/modules/sd-core/android/build.gradle
+++ b/apps/mobile/modules/sd-core/android/build.gradle
@@ -82,11 +82,6 @@ android {
       withSourcesJar()
     }
   }
-	sourceSets {
-		main {
-			jniLibs.srcDirs = ['../../../../../target/release']
-		}
-	}
 }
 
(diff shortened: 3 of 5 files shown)


Rocky43007:
@HeavenVolkoff Hey, Utku mentioned that you were a a shell scripting wizard. I was wondering if you knew a way to get the `cargo ndk` builds to run in parallel? I tried but when killing the script, it wouldn't kill the builds running the the background, and I couldn't get the 4 outputs to show up in case one had an error for some reason.

HeavenVolkoff:
Hey @Rocky43007 😊

I can take a look at it. Just to confirm, you want this loop to run in parallel, right? And you want the stdout and stderr of all the command to be outputted at the same time?

https://github.com/spacedriveapp/spacedrive/blob/715f47309d686b5a5d4f466853130609c57cd398/apps/mobile/modules/sd-core/android/build.sh#L44-L48

Rocky43007:
Hey! 

Yes, because once you add more than 2 targets, the build takes almost 10 minutes (especially for x86_64 taking the longest). For the outputs, yes, just for the edge case if the code crashes for one of the builds but not the others.

oscartbeaumont:
Correct me if I'm missing something but Cargo has a file system lock on the build directory so while you can spawn multiple "cargo build"'s they will still block each other.

I can't find any references online if the different architecture being built for means they don't block each other so I can't say for sure if they will/won't block but just checking it's been considered.

We could maybe set each of them to it's own target directory if this is a problem.

Rocky43007:
I just checked, and you're right. We can only do one build at a time due to the file lock. This was just something I was looking to get as a nice to have. 

References: 
https://users.rust-lang.org/t/having-multi-build-together/32469
https://users.rust-lang.org/t/compile-to-mutliple-targets-at-once/21802

Rocky43007:
One thing that needs final checking on this script is if the CI works with said script. I have no clue how to test that.

utkubakir:
> One thing that needs final checking on this script is if the CI works with said script. I have no clue how to test that.

We can ignore Android CI. I am sure it will work fine though

utkubakir (dismissed):
lgtm

HeavenVolkoff:
> @HeavenVolkoff Can you take a quick look?

Taking a look rn

HeavenVolkoff (changes requested):
This is just to block the PR while I fix some things

HeavenVolkoff (approved):
Tested on my Pixel 7 Pro, app built and launched fine, nice work @Rocky43007

PR #2103, opened 2024-02-19
[MOB-58] Settings routes new design & more

ameer2468:
**This PR**: updates the design of every settings route, except the ones that are classified as TODO.

- Also included location path data **(as before design updates)** in location records. 
- New '+' button for locations and tags screen.
- List/Grid view for Tags via props **(in another PR we can add a view toggle)**.

**Important note**:

Some, or a few of these setting pages will be re-visited in the future to make them even better.

**Images:**

frame


apps/mobile/src/components/browse/BrowseTags.tsx
```diff
@@ -1,55 +1,136 @@
...
 	const modalRef = useRef<ModalRef>(null);
+
+	const renderTagView = () => (
+		<View
+			style={twStyle(
+				`h-auto ${
+					viewStyle === 'grid' ? 'w-[90px]' : 'w-full'
+				} flex-col justify-center gap-2.5 rounded-md border border-app-line/50 bg-app-box/50 p-2`,
+				tagStyle
+			)}
```
utkubakir, line 44: no need for `${}` here

ameer2468: mb

apps/mobile/src/components/browse/BrowseTags.tsx
```diff
@@ -67,7 +148,7 @@ const BrowseTags = () => {
 
 	return (
 		<View style={tw`gap-5`}>
-			<View style={tw`w-full flex-row items-center justify-between px-7`}>
+			<View style={tw`flex-row items-center justify-between w-full px-7`}>
```
utkubakir, line 151: Not sure what do you use for formatting these but it seems to break eslint

apps/mobile/src/components/primitive/Button.tsx
```diff
@@ -30,6 +30,7 @@ const button = cva(['items-center justify-center rounded-md border shadow-sm'],
 });
 
 type ButtonProps = VariantProps<typeof button> & PressableProps;
+export type ButtonVariants = VariantProps<typeof button>['variant'];
```
utkubakir, line 33: `ButtonProps['variant']` would give you the same result

apps/mobile/src/components/settings/SettingsToggle.tsx
```diff
@@ -0,0 +1,65 @@
+import { useState } from 'react';
+import { Control, Controller } from 'react-hook-form';
+import { Switch, Text, View } from 'react-native';
+import { tw } from '~/lib/tailwind';
+
+type Props =
+	| {
+			title: string;
+			description?: string;
+			onEnabledChange?: (enabled: boolean) => void;
```
utkubakir, line 10: what does this do exactly? maybe add some comments

apps/mobile/src/components/settings/SettingsToggle.tsx
```diff
@@ -0,0 +1,65 @@
...
+			control: Control<any>;
+			name: string;
+	  }
+	| {
+			title: string;
+			description?: string;
+			onEnabledChange?: (enabled: boolean) => void;
+			control?: never;
+			name?: never;
+	  };
```
utkubakir, line 20: Why not make control and name optional instead of complicating the type?

ameer2468: > Why not make control and name optional instead of complicating the type?

This is not a good idea, you need to pass in both so the controller would work correctly. Passing either one isn't an option.

utkubakir (approved):

PR #2107, opened 2024-02-20
Update github actions due to nodejs 16 deprecation

HeavenVolkoff:
- Replace archived actions-rs/clippy-check with maintained fork actions-rs-plus/clippy-check
 - Replace redhat-actions/push-to-registry with updated fork Eusebiotrigo/push-to-registry
 - Point redhat-actions/buildah-build and softprops/action-gh-release to current master to fix Node.js deprecation
 - Made some change to the iOS build_rust script and Mobile CI to build the app and lib for the same arch as the host machine when using iphone_simulator
 - Made some changes to make cache-factory faster and avoid failing so much
 - Add trigger to run cache-factory on pull requests when there are changes to itself

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -100,6 +100,8 @@ libp2p-core = { git = "https://github.com/spacedriveapp/rust-libp2p.git", rev = 
 libp2p-swarm = { git = "https://github.com/spacedriveapp/rust-libp2p.git", rev = "a005656df7e82059a0eb2e333ebada4731d23f8c" }
 libp2p-stream = { git = "https://github.com/spacedriveapp/rust-libp2p.git", rev = "a005656df7e82059a0eb2e333ebada4731d23f8c" }
 
+[profile.dev]
+
 # Set the settings for build scripts and proc-macros.
 [profile.dev.build-override]
 opt-level = 3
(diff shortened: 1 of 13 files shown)


Cargo.toml
```diff
@@ -94,6 +94,8 @@ rspc = { git = "https://github.com/spacedriveapp/rspc.git", rev = "f3347e2e8bfe3
 tauri = { git = "https://github.com/spacedriveapp/tauri.git", rev = "8409af71a83d631ff9d1cd876c441a57511a1cbd" }
 tao = { git = "https://github.com/spacedriveapp/tao", rev = "7880adbc090402c44fbcf006669458fa82623403" }
 
+[profile.dev]
```
brxken128, line 97: I don't think this does anything as it's not overriding/setting any keys?

brxken128 (commented):
Looks good to go aside from the one comment regarding the empty `[profile.dev]` :D

Cargo.toml
HeavenVolkoff: Yeah, this is being used by the Cache Factory action:
https://github.com/spacedriveapp/spacedrive/blob/64d5ce54322b1cb519e2aa8112002e5fb66ce8c8/.github/workflows/cache-factory.yaml#L84-L102

I thought about just appending the whole section in the action, but I decided it would be better to leave an empty section in the config file to avoid anyone in the feature wanting to add it and the script overwriting it or having an error due to duplicate sections.

I can add a comment to explain why the empty section

brxken128 (approved):
Looks all good to me!

PR #2159, opened 2024-03-04
Mob: minor tweaks and fixes

ameer2468:
**This PR**:

- Scrollview of locations screen, had to add a prop to remove the scroll view to deal with a error message.
- No longer need to disable view all locations button due to design updates.
- Added a eslint rule to stop enforcing shorthands for mobile specifically to prevent un-supported classes.


utkubakir (approved):
FYI, mobile should use scrollview whenever possible for the best support for all screen sizes + keyboards etc etc

PR #2160, opened 2024-03-04
i18n some polishes

HeavySnowJakarta:
This i18n seems to be made by ai.

I made some polishment so it looks better now.

btw you have not considered a l10n platform like crowdin?


HeavySnowJakarta:
note it's still far from perfect.

PR #2174, opened 2024-03-07
add japanese

wany-oh:
I tried to add Japanese localization and keys


utkubakir (approved):

PR #2220, opened 2024-03-19
Remove `IdentityOrRemoteIdentity`

oscartbeaumont:
It will cause a security issue at some point so lets just remove it before that happens.

and Jake wants `identity` in the Spacedrive keychain/vault thing anyway which is fair so this will make that an easy migration.


brxken128 (approved):

brxken128:
The unwraps should be fine and we can always clear them up a little later on imo

PR #2225, opened 2024-03-20
[MOB-70] Update Folder/Files when renamed on iOS

Rocky43007:
If a folder or file has been renamed externally on iOS, the core has no idea how to handle this event, throwing errors and disappearing the folder/file. 

Looks like it's due to iOS sending a Create Event with the new Folder instead of sending it as a Rename Event.


core/src/location/manager/watcher/ios.rs
```diff
@@ -303,90 +311,32 @@ impl IosEventHandler<'_> {
...
+				&path,
+				fs::metadata(&key)
+					.await
+					.map_err(|e| FileIOError::from((&key, e)))?,
+				self.library,
+			)
+			.await?;
+
+			// Remove the path from the rename event queue
+			self.rename_event_queue.remove(&key.clone());
```
fogodev, line 335: Shouldn't here be just `self.rename_event_queue.remove(key);` as `key` is already a reference?

core/src/location/manager/watcher/ios.rs
```diff
@@ -303,90 +311,32 @@ impl IosEventHandler<'_> {
...
-					.await?;
-				} else {
-					trace!("No match for old path yet: {}", path.display());
-					// We didn't find a new path for this old path, so we store ir for later
-					self.old_paths_map.insert(inode, (Instant::now(), path));
-				}
-			}
-			Err(e) => return Err(FileIOError::from((path, e)).into()),
+		if let Some((key, _)) = self.rename_event_queue.iter().nth(0) {
+			let new_path_name = Path::new(key).file_name().unwrap();
```
fogodev, line 320: `key` is a `&PathBuf` here, so `key.file_name()` should do, if I'm not mistaken.

core/src/location/manager/watcher/ios.rs
```diff
@@ -303,90 +311,32 @@ impl IosEventHandler<'_> {
...
+			let new_path_name_string = Some(new_path_name.to_str().unwrap().to_string());
```
fogodev, line 321: Also try to avoid these `unwrap`s if possible, yon can put a `NonUtf8PathError` variant on `LocationManagerError` and return this error here. I don't know if paths on iOS as guaranteed to always be valid UTF-8. If they are, then you can put it written on a `expect(...)` message.

fogodev (changes requested):

core/src/location/manager/watcher/ios.rs
Rocky43007: No, because with `key`, we get the error:

```
cannot borrow `self.rename_event_queue` as mutable because it is also borrowed as immutable
mutable borrow occurs here
ios.rs(319, 27): immutable borrow occurs here
ios.rs(335, 28): immutable borrow later used by call
```

And with `&key`, it's:
```
the trait bound `std::path::PathBuf: std::borrow::Borrow<&std::path::PathBuf>` is not satisfied
the trait `std::borrow::Borrow<std::path::Path>` is implemented for `std::path::PathBuf`
for that trait implementation, expected `std::path::Path`, found `&std::path::PathBuf`
ios.rs(335, 28): required by a bound introduced by this call
map.rs(1157, 12): required by a bound in `std::collections::HashMap::<K, V, S>::remove`
```

core/src/location/manager/watcher/ios.rs
Rocky43007: Huh, that does work.

core/src/location/manager/watcher/ios.rs
Rocky43007: It looks like iOS paths are always UTF-8, so I'll add `expect()` instead of `unwrap()`.

core/src/location/manager/watcher/ios.rs
fogodev: I think that [this](https://doc.rust-lang.org/std/path/struct.PathBuf.html#method.as_path) will do the trick perhaps?

fogodev: Never mind lol, it seems we can't use a reference from inside a hashmap to remove the element in the hashmap

Rocky43007:
Working on a fix to multiple rename events being set at the same time because otherwise, it's broken and crashes.

PR #2232, opened 2024-03-22
Alpha 0.2.6

utkubakir:


HeavenVolkoff:
We may need to increase timeouts for the Cypress test run on CI... Currently, running the tests locally just to make sure this isn't something else that broke

HeavenVolkoff:
> We may need to increase timeouts for the Cypress test run on CI... Currently, running the tests locally just to make sure this isn't something else that broke

Yep, local test passed just fine. I doubled the cypress default timeout when running on CI

Rocky43007 (approved):
CI change worked as well. :D

PR #2241, opened 2024-03-24
[ENG-1693] Prevent default webview search

ameer2468:
^

Keep in mind the search input is not rendered at all times **(Ephemeral locations)**. Hence the approach.


brxken128 (approved):
Looks good!

For anyone wondering, this resolves the following issue:
Screenshot 2024-03-25 at 15 25 28

PR #2268, opened 2024-04-01
[ENG-1704] Location fixed width sizing

ameer2468:
**This PR**: sets a fixed width for location size for better visual consistency. 

**Also:**

- Removes `Size:` text - it should be obvious that it is the size. It's like putting **Status:** next to 'Connected'
- Added a tooltip when the size is truncated.
- Reorganized the layout.

Screenshot 2024-04-01 at 3 50 30 PM


brxken128 (approved):
Looks clean, and I agree with the "it should be obvious that it's the size [of the location]" sentiment. Nice job man!

PR #2273, opened 2024-04-02
[ENG-1705] Locales for sizing units

ameer2468:
^

Other languages remain as is, according to Google.


utkubakir (changes requested):
These keys are not descriptive imo, could become a problem in the future. Let's lowercase it and add size before it. (so GB = "size_gb")

brxken128 (approved):
This works for now!

I'd 100% consider using what @Rocky43007 suggested in the thread later on however, as that seems to cover all locales (and it handles whether or not the unit comes before or after the value, as well as translating the numbers themselves in some cases).

ameer2468:
> These keys are not descriptive imo, could become a problem in the future. Let's lowercase it and add size before it. (so GB = "size_gb")

We cannot do any lowercasing due to the value being returned from `byteSize` - it returns kB for kilobytes, meaning we cannot do KB for example, unless we write a simple switch function to handle this. This doesn't seem worth it imo just to handle casing.

But i'll add the size_ and we can do `{t(`size_${byteSize(location.size_in_bytes).unit}`)}`

utkubakir:
> > These keys are not descriptive imo, could become a problem in the future. Let's lowercase it and add size before it. (so GB = "size_gb")
> 
> We cannot do any lowercasing due to the value being returned from `byteSize` - it returns kB for kilobytes, meaning we cannot do KB for example, unless we write a simple switch function to handle this. This doesn't seem worth it imo just to handle casing.
> 


Or you use toLowercase function :)

PR #2274, opened 2024-04-02
[MOB-80] Drawer design improvement & more

ameer2468:

- Drawer design improvements.
- Menu/drawer button added and only shows on root screens.
- Feedback placeholder button added.
- View all button for locations only shows when there's more than 3.
- For tags view all button shows when there's more than 4.

**Note**: Job manager button/design will be given an update. Not comfortable with how it is right now.

**Images**:

Frame 4


utkubakir:
Can we add online/offline indicator to the location and make the icon slightly smaller?

ameer2468:
> Can we add online/offline indicator to the location and make the icon slightly smaller?

Frame 4

utkubakir (approved):

PR #2275, opened 2024-04-02
Fix pnpm prep fails to download native-deps on some specific network env

HeavenVolkoff:
- Adjust fetch call to enable `autoSelectFamily` and `allowH2` so `pnpm prep` can function correctly on some specific network environments (Ex: corporate networks with proxies)
 - ESLint/Prettier auto format


utkubakir:
Let's add locales to prettierignore

utkubakir (approved):

PR #2280, opened 2024-04-04
Missing thumbnails

fogodev:


fogodev:
Doesn't seem as a backend problem actually, closing this one

PR #2282, opened 2024-04-04
Disable Labeler and all mentions of Labeling

Rocky43007:
For now, only. We'll re enable it once it works properly.


utkubakir:
Let's disable the job as well

Rocky43007:
> Let's disable the job as well

It should be because we just disable the `ai` feature completely?

utkubakir:
> > Let's disable the job as well
> 
> It should be because we just disable the `ai` feature completely?

Have you tested it?

Rocky43007:
> > > Let's disable the job as well
> > 
> > 
> > It should be because we just disable the `ai` feature completely?
> 
> Have you tested it?

Just went ahead and disabled it fully. Tested and working when adding a new folder with files that was previously unindexed.

utkubakir:
would love to get @fogodev @HeavenVolkoff review on this

fogodev:
For the backend I don't think that you should comment anything at all. You only need to disable the feature in the Cargo.toml files.

core/src/api/jobs.rs
```diff
@@ -248,15 +248,17 @@ pub(crate) fn mount() -> AlphaRouter<Ctx> {
...
-						sub_path: Some(path),
-						regenerate_thumbnails: regenerate,
-						regenerate_labels: false,
-					})
-					.spawn(&node, &library)
-					.await
-					.map_err(Into::into)
+					Ok(("Currently disabled".to_string(), 0))
+
+					// Job::new(OldMediaProcessorJobInit {
```
fogodev, line 253: If you comment this job, then no thumbnails will be generated lol, also no media data extraction

brxken128:
> For the backend I don't think that you should comment anything at all. You only need to disable the feature in the Cargo.toml files. 

Yeah that seems like the easier option than commenting everything for sure.

fogodev (changes requested):

PR #2283, opened 2024-04-04
OpenDAL - Ephemeral Locations

oscartbeaumont:
Implement OpenDAL for the ephemeral indexer.

For now it will only supporting their FS backend but Google Drive, S3, etc will be easy in a follow up PR once we deal with authenticating to them.

Changes:
 - [x] Break ephemeral indexer out into `sd-indexer` (crates better for compile time and separation of concerns)
 - [x] Upgrade the indexer to be fully streaming
     - [x] Basic OpenDAL integration
     - [x] Indexer rules
     - [x] Thumbnailer
     - [x] Link with locations
     - [x] Work out sorting (done on the frontend now)
     - [x] Fix volume file path handling -> OpenDAL is *very* specific and it's currently different than it would expect.
     - [x] ~~Symlinks/Macintosh HD~~ - Legacy code doesn't handle these so we good for now.
     - [x] Doubled up file extensions
 - [x] Testing
     - [x] Is sorting correct? For what I can tell
     - [x] ~~Does location show as location?~~ - Broken in existing app so left broken
     - [x] Thumbnailer
     - [x] ~~Indexer rules~~ - Partially working, but broken in same way as existing app
 - [x] Clippy

I have created Linear issues for any of the bugs that I found, but if they are broken in the current app I have not fixed them (can be done in follow up PR's).


oscartbeaumont:
Neat although that was only the full indexer right, not the ephemeral indexer?

I had a quick loop and the changes didn't seem to overlap.

fogodev:
> Neat although that was only the full indexer right, not the ephemeral indexer?
> 
> I had a quick loop and the changes didn't seem to obviously overlap.

Yeah, this first PR will be just for the full indexer job. Another coming PR will be for the shallow one. I haven't thought much about the ephemeral one yet though

Brendonovich (approved):

PR #2288, opened 2024-04-05
Disable Labeler

Rocky43007:
For now, only. We'll re enable it once it works properly.


HeavenVolkoff (approved):
LGTM, just waiting CI to finish before merging

PR #2308, opened 2024-04-11
Minor changes

utkubakir:
- moved things around
image
- enable sort by/direction setting for list view
- more i18n


brxken128 (approved):
Can't comment much on the translations obviously (the Russian looked close enough) but the more i18n the better! Nice job man :D

PR #2314, opened 2024-04-11
Basic HTTP auth for sd-server

oscartbeaumont:
Good enough for now so I can stop bringing this up in the context of our docker image.
Screenshot 2024-04-11 at 2 08 08 PM

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,6 +11,7 @@ env:
   CARGO_NET_RETRY: 10
   RUST_BACKTRACE: short
   RUSTUP_MAX_RETRIES: 10
+  SD_AUTH: disabled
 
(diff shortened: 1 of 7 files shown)


utkubakir (changes requested):
Can we add little guide on docs about it? also CI's

oscartbeaumont:
Typescript - A TS failure merged into main  (caused by another PR)
Mobile CI - Missing element (caused by another PR)
Cypress - Pushed a fix, didn't know it relied on sd-server.

oscartbeaumont:
In terms of docs I can do but it's also autodocumenting. Auth is required in production and it will guide you through configuring it with that screenshot above.

utkubakir (approved):
LGTM

PR #2333, opened 2024-04-15
[ENG-1695] Global setting for explorer views

utkubakir:
image

also did:

- some translations
- removed tag assign mode as it's not implemented


ameer2468 (approved):

Brendonovich:
do we want settings like this to sync? location and tag explorer settings do and it's why the preferences system exists, but we can move everything to be similar to this if we don't care about them syncing

utkubakir:
> do we want settings like this to sync? location and tag explorer settings do and it's why the preferences system exists, but we can move everything to be similar to this if we don't care about them syncing

moved this discussion to Slack :)

PR #2334, opened 2024-04-16
[ENG-1629] Write new file identifier with the task system

fogodev:
New File Identifier, job and shallow, with its tasks for extracting file metadata and create or link `objects` for `file_paths`.

Also, some minor tweaks here and there.


HeavenVolkoff (approved):
LGTM

PR #2337, opened 2024-04-16
I18n polish (zh-cn)

HeavySnowJakarta:
did some review on the zh-cn i18n file.

- reviewed 1st-100th lines of the file (more than 450 lines totally). It's really a thing to translate into Chinese as you have to pay attention to many aspects to let it feel not weird. Not doing traditional Chinese because I'm not proficient in it.

- also did some comments as they may be discussed


utkubakir:
Conflicts are fun

utkubakir:
@HeavySnowJakarta

HeavySnowJakarta:
> @HeavySnowJakarta

let me have a try. i have no experience resolving conflicts and i think use github online to do it directly is slow but i'll try my best.

HeavySnowJakarta:
I noticed since the last commit some new strings have been added and some strings have been deleted. I just accepted new strings but deleted strings were remained. I can't promise there will be no problem.

utkubakir (approved):
I am sure everything will be fine :D Thanks!

PR #2338, opened 2024-04-16
Open Trash from the application

Rocky43007:
Creates basically a sym-link from the app to the OS's Trash.


core/src/api/search/mod.rs
```diff
@@ -490,6 +490,64 @@ pub fn mount() -> AlphaRouter<Ctx> {
...
+						})?
+						.wait()
+						.map_err(|err| {
+							rspc::Error::new(ErrorCode::InternalServerError, err.to_string())
+						})?;
+
+						Ok(())
+				}
+			})
+		})
```
utkubakir, line 550: Might be possible to do it all in desktop (tauri) and save us a rspc mutation - not exactly sure though @Brendonovich @oscartbeaumont

utkubakir (changes requested):

utkubakir:
You need to setup your prettier :) can be merged if/when mobile CI passess (or builds)

utkubakir (approved):

Issue #2039, opened 2024-02-01
AppImage doesn't work on systems with Glibc < 2.36


zeet2020:
### Describe the bug

on launching the spacedrive.appimage getting empty window with title
![spacedrive](https://github.com/spacedriveapp/spacedrive/assets/1552266/dd12d09a-6317-40c4-8b3e-c138e0e5c506)
on executing on cmdline i see following erroe
```2024-02-01T18:53:11.128607Z  INFO sd_desktop::tauri_plugins: apps/desktop/src-tauri/src/tauri_plugins.rs:80: Internal server listening on: http://127.0.0.1:41467 http://127.0.0.1:42435 http://127.0.0.1:45323 http://127.0.0.1:45139
Gtk-Message: 00:23:12.008: Failed to load module "canberra-gtk-module"
Gtk-Message: 00:23:12.020: Failed to load module "canberra-gtk-module```

i check canberra-gtk-module are already install on my pop-os machine 

### Reproduction

* install pop-os on vm
* download appimage
* try to open the appimage

### Expected behavior

* spacedrive should have worked 

### Platform and versions

```Shell
LSB modules are available.
Distributor ID:	Pop
Description:	Pop!_OS 22.04 LTS
Release:	22.04
Codename:	jammy
```

HeavenVolkoff:
This seems to be a similar issue to #1993. Are you using X11? Can you reproduce the same issue with the [deb release](https://github.com/spacedriveapp/spacedrive/releases/download/0.2.2/Spacedrive-linux-x86_64.deb)?

TheGB0077:
@HeavenVolkoff I found the same issue with the AppImage on today's release, but (seemingly) the issue cannot be reproduced with the deb version.

RAVENz46:
I realized that `Exec=usr/bin/spacedrive` in com.spacedrive.desktop.
But I think it should be `Exec=AppRun` or atleast `Exec=/usr/bin/spacedrive`.

AutumnalAntlers:
I've had (specific distro aside) exactly the same experience. I don't know the first thing about the tech stack, but blindly tried swapping my system libraries and glibc/linker, into the appimage since they worked for the deb, which is I confirmed is the same binary. No dice.

Loving the vision in the meantime though, thanks to everyone who's making this a reality.

[...]

PR #2345, opened 2024-04-17
Replace Appimage with deb

utkubakir:
AppImage doesn't work with Ubuntu < 22.04 based distros, Arch Linux etc.

Related: https://github.com/spacedriveapp/spacedrive/issues/2039

### DIFF
--- a/.github/actions/setup-pnpm/action.yml
+++ b/.github/actions/setup-pnpm/action.yml
@@ -11,7 +11,7 @@ runs:
     - name: Install pnpm
       uses: pnpm/action-setup@v3
       with:
-        version: 9.x.x
+        version: 9.0.2
 
     - name: Install Node.js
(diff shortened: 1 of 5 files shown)


HeavenVolkoff:
I agree, AppImage is too broken, and I am not sure if it is fixable. As soon as I finish my current tasks, I will start working on adding support for Arch (and derivatives) natively and Flatpak, which should cover all major distributions:
deb -> Ubuntu, Debian, Mint, PopOS
pkg -> Arch, Manjaro
flatpak -> Fedora and SteamOS3 natively, plus any other distro that has the fedora package available for it: https://www.flatpak.org/setup/

HeavenVolkoff (dismissed):

HeavenVolkoff (approved):

PR #2366, opened 2024-04-19
[ENG-1745] Mouse wheel resize

ameer2468:
**This PR**: adds support for mouse wheel to resize icons in list view, and item size in grid view. Only for Windows OS for a native experience.


interface/app/$libraryId/Explorer/OptionsPanel/ListView/IconSize.tsx
```diff
@@ -26,6 +26,7 @@ export const IconSize = () => {
 			<Slider
 				step={1}
 				max={sizes.length - 1}
+				value={[defaultValue]}
```
utkubakir, line 29: do we need this here?, why default value is not enough?

interface/app/$libraryId/Explorer/OptionsPanel/index.tsx
```diff
@@ -108,6 +108,7 @@ export default () => {
 							onValueChange={(value) => {
 								explorer.settingsStore.gridItemSize = value[0] || 100;
 							}}
+							value={[settings.gridItemSize]}
```
utkubakir, line 111: same as other

interface/app/$libraryId/Explorer/OptionsPanel/ListView/IconSize.tsx
ameer2468: Options panel input doesnt update otherwise.

utkubakir: i guess you can remove `defaultValue` then, having both makes it confusing

utkubakir: defaultValue will be shown when value is undefined, in this case it will never be

interface/hooks/useMouseItemResize.ts
```diff
@@ -0,0 +1,65 @@
...
+				) + deltaYModifier;
+
+			const minSize = isList ? 0 : 60;
+			const maxSize = isList ? 2 : 200;
+			const clampedSize = Math.max(minSize, Math.min(maxSize, newSize));
+
+			if (isList) {
+				explorer.settingsStore.listViewIconSize = String(
+					clampedSize
+				) as keyof typeof LIST_VIEW_ICON_SIZES;
```
niikeec, line 34: can we improve this so that modifying LIST_VIEW_ICON_SIZES (adding, removing sizes) won't require us to make adjustments here

utkubakir (approved):

PR #2383, opened 2024-04-24
[MOB-90] Visual adjustments

ameer2468:
**This PR**:

- runs eslint fix.
- Visual adjustments made previously i.e make search input abit smaller, smaller size of icon of tag etc..
- Includes other previous changes that were made too.

### DIFF
--- a/apps/mobile/src/components/locations/GridLocation.tsx
+++ b/apps/mobile/src/components/locations/GridLocation.tsx
@@ -28,7 +28,7 @@ const GridLocation: React.FC<GridLocationProps> = ({ location, modalRef }: GridL
 							)}
 						/>
 					</View>
-					<Pressable hitSlop={24} onPress={() => modalRef.current?.present()}>
+					<Pressable hitSlop={12} onPress={() => modalRef.current?.present()}>
 						<DotsThreeOutlineVertical
 							weight="fill"
 							size={20}
--- a/apps/mobile/src/components/locations/GridLocation.tsx
+++ b/apps/mobile/src/components/locations/GridLocation.tsx
@@ -28,7 +28,7 @@ const GridLocation: React.FC<GridLocationProps> = ({ location, modalRef }: GridL
 							)}
 						/>
 					</View>
-					<Pressable hitSlop={12} onPress={() => modalRef.current?.present()}>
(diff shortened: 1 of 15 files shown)


apps/mobile/src/screens/settings/Settings.tsx
```diff
@@ -141,24 +141,22 @@ export default function SettingsScreen({ navigation }: SettingsStackScreenProps<
...
+							onPress={() => navigation.navigate(item.navigateTo as any)}
+							rounded={item.rounded}
+						/>
+					))}
+				</View>
+			))}
+			<FooterComponent />
 		</ScreenContainer>
 	);
 }
```
utkubakir, line 162: any reason to why we are not using section list?

apps/mobile/src/components/locations/GridLocation.tsx
```diff
@@ -28,7 +28,7 @@ const GridLocation: React.FC<GridLocationProps> = ({ location, modalRef }: GridL
 							)}
 						/>
 					</View>
-					<Pressable hitSlop={24} onPress={() => modalRef.current?.present()}>
+					<Pressable hitSlop={12} onPress={() => modalRef.current?.present()}>
```
utkubakir, line 31: i would remove hitslop here, 20px is good size for tapping

apps/mobile/src/components/tags/GridTag.tsx
```diff
@@ -20,7 +20,7 @@ const GridTag = ({ tag, modalRef }: GridTagProps) => {
 						backgroundColor: tag.color!
 					})}
 				/>
-				<Pressable hitSlop={24} onPress={() => modalRef.current?.present()}>
+				<Pressable hitSlop={12} onPress={() => modalRef.current?.present()}>
```
utkubakir, line 23: i would remove hitslop here, 20px is good size for tapping

apps/mobile/src/screens/settings/Settings.tsx
utkubakir: yeah section list is the optimized and cleaner way to do it. No need to do it here but why change it if we don't need to? (like before)

utkubakir (approved):

PR #2408, opened 2024-04-27
[MOB-89] Separate headers

ameer2468:
**This PR**: Provides and creates a few different headers to accomodate and fullfill different purposes.

- The styling has been duplicated on purpose to allow for design iterations as needed, instead of wrapping all different headers with the same style.
- Instead of manually setting "title" for different routes, now many screens simply use the `route` object for cleaner code, however in some cases like in SettingsStack, title can be provided manually since the names of the `route` themselves are not desired.

I think this is good for now - as it offers the level of flexibility we need to iterate the design of each header + support better understanding of our headers.

### DIFF
--- a/apps/mobile/src/components/header/SearchHeader.tsx
+++ b/apps/mobile/src/components/header/SearchHeader.tsx
@@ -1,23 +1,28 @@
 import { DrawerNavigationHelpers } from '@react-navigation/drawer/lib/typescript/src/types';
 import { RouteProp, useNavigation } from '@react-navigation/native';
-import { ArrowLeft, List, MagnifyingGlass } from 'phosphor-react-native';
+import { ArrowLeft } from 'phosphor-react-native';
 import { Platform, Pressable, Text, View } from 'react-native';
 import { useSafeAreaInsets } from 'react-native-safe-area-context';
 import { tw, twStyle } from '~/lib/tailwind';
+import Search from '../search/Search';
+
+
+const searchPlaceholder = {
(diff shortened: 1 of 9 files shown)


apps/mobile/src/components/header/SearchHeader.tsx
```diff
@@ -0,0 +1,53 @@
...
+		<View
+			style={twStyle('relative h-auto w-full border-b border-app-cardborder bg-app-header', {
+				paddingTop: headerHeight + (isAndroid ? 15 : 0)
+			})}
+		>
+			<View style={tw`mx-auto h-auto w-full justify-center px-5 pb-3`}>
+				<View style={tw`w-full flex-row items-center justify-between`}>
+					<View style={tw`flex-row items-center gap-3`}>
+					<Pressable
+								hitSlop={24}
```
utkubakir, line 41: Let's not use hitSlop unless it's 100% necessary

utkubakir: eh I guess we can remove them all later, ignoring this for now

utkubakir (approved):
lgtm

apps/mobile/src/components/header/SearchHeader.tsx
ameer2468: > eh I guess we can remove them all later, ignoring this for now

I added them originally due to Jamie saying buttons were hard to click, and he suggested to increase the radius of pressability (hitSlop).

PR #2409, opened 2024-04-27
Fixed Belarusian language entry name

artsiom-voitas:
Fixed Belarusian language entry name from "by" to "be", which is correct key in i18n docs, and fixed one translation value.


interface/app/$libraryId/settings/client/general.tsx
```diff
@@ -28,7 +28,7 @@ const LANGUAGE_OPTIONS = [
 	{ value: 'fr', label: 'Français' },
 	{ value: 'tr', label: 'Türkçe' },
 	{ value: 'nl', label: 'Nederlands' },
-	{ value: 'by', label: 'Беларуская' },
+	{ value: 'be', label: 'Беларуская' },
```
utkubakir, line 31: Need more information on this. Also is this tested on Desktop? Folder name is "by", changing a language might not work here.

image

artsiom-voitas: In day.js [docs](https://cdn.jsdelivr.net/npm/dayjs@1/locale.json), they use it like `"key":"be","name":"Belarusian"`, same goes with [date-fns](https://github.com/date-fns/date-fns/tree/main/src/locale/be)
I messed up with folder name, while commit, will fix it asap

artsiom-voitas: image
I didn't mess up, we just didn't notice, the folder name was changed with the commit. And it works properly!

artsiom-voitas: image
I didn't mess up, we just didn't notice, the folder name was changed with the commit. And it works properly!

utkubakir: cool, just to confirm you tested this desktop?

artsiom-voitas: > cool, just to confirm you tested this desktop?

Yeah, it works
And just to confirm from your end, is project.inlang/settings.json gone for good? There is no other config file, which need to be used to 100% right configuration of i18n?

utkubakir (approved):

Issue #2406, opened 2024-04-26
Formatting dates with i18n


artsiom-voitas:
### Describe the bug

Currently, Spacedrive formats dates, for example on (spacedrive\interface\app$libraryId\Explorer\Inspector\index.tsx):
`<MetaData icon={Eraser} label={t('modified')} value={formatDate(dateModified)} />`
using JavaScript capabilities, which makes it impossible to have date mapping in different languages. 
It is worth considering using the i18n functions provided in the [documentation](https://www.i18next.com/translation-function/formatting#datetime)

### Reproduction

1. Check date in Inspector

### Expected behavior

Date formatting matches app language.

### Platform and versions

```Shell
spacedrive - v0.2.13
```

utkubakir:
We should tackle this https://github.com/spacedriveapp/spacedrive/discussions/2389 with this issue also

artsiom-voitas:
> We should tackle this https://github.com/spacedriveapp/spacedrive/discussions/2389 with this issue also

Do you want to add Date Format selector in the settings? Somewhere under Language settings? If yes, I will continue work on this matter tomorrow, if you want to format according to language preferences, I can push it now

utkubakir:
> Do you want to add Date Format selector in the settings? Somewhere under Language settings? If yes, I will continue work on this matter tomorrow, if you want to format according to language preferences, I can push it now

Yes, I think we should default to the local time settings and user should also be able to change it

artsiom-voitas:
> > Do you want to add Date Format selector in the settings? Somewhere under Language settings? If yes, I will continue work on this matter tomorrow, if you want to format according to language preferences, I can push it now
> 
> Yes, I think we should default to the local time settings and user should also be able to change it

Is it good idea to make a changes to local time from chosen language? For example, We add two version of English:
1. English (US): which will match key en_US and bring us MM/DD/Year and 12 hours formatting
2. English (UK): which will math key en_UK and bring us DD/MM/Year and 24 hours formatting
Also, we will have a feature in settings to change preferable date formatting style, from the list
![image](https://github.com/spacedriveapp/spacedrive/assets/85301789/6514491a-e5e6-4fff-a36b-1d2f52192f45)
Does It sound correct?

utkubakir:
yup, options like "System", "DD/MM/YYYY", "M/D/Y" would be nice. And for extra points, we could display how it would look under the time display setting so it's easily understandable.

PR #2414, opened 2024-04-29
Added date formatting according to chosen language in the settings

artsiom-voitas:
This pull requests addresses issue [#2406](https://github.com/spacedriveapp/spacedrive/issues/2406)
- added an option in settings to choose preferred way for displaying date
- added i18n entries and translations for format date settings
- added English (UK) language to distinguish Date Format on English between US variant and European


artsiom-voitas:
> Code looks good, will test it soon

Please check if the language of the displayed date format always changes when the language is changed. I have a situation where with 3-4 language switches in a row, the date displayed in the settings is not always updated to the language of the date format in the selector (I think I've made a mistake somewhere, and it doesn't always trigger a component reload, but I don't see where). It would be great to fix this issue before adding the function to the main branch. Thank you for your help

utkubakir:
> Please check if the language of the displayed date format always changes when the language is changed. I have a situation where with 3-4 language switches in a row, the date displayed in the settings is not always updated to the language of the date format in the selector (I think I've made a mistake somewhere, and it doesn't always trigger a component reload, but I don't see where). It would be great to fix this issue before adding the feature to the main branch. Thank you for your help

Looks like a race condition to me, setting the language format after i18n seems to be more consistent

utkubakir (approved):

interface/app/$libraryId/Explorer/util.ts
```diff
@@ -33,3 +34,58 @@ export function getItemId(index: number, items: ExplorerItem[]) {
...
+	it: () => import('dayjs/locale/it.js'),
+	ja: () => import('dayjs/locale/ja.js')
+};
+
+export function loadDayjsLocale(language: string) {
+	dayjsLocales[language]().then(() => {
+		language = language.replace('_', '-');
+		dayjs.locale(language);
+	});
+}
```
utkubakir, line 59: can this be optimized? bundling all locales doesn't sound smart

utkubakir (approved):

utkubakir:
Thank you, lgtm @artsiom-voitas

PR #2423, opened 2024-04-29
[ENG-1634 / ENG-1636] Integrate Media Data Extractor and Thumbnail Actor with New Task System

fogodev:
This PR introduces the brand new Media Processor job and shallow job, with its Media Data Extractor and Thumbnailer tasks.


HeavenVolkoff (commented):
LGTM, most of this is preparation work for the new index, so it is not affecting prod releases

HeavenVolkoff (approved):

PR #2412, opened 2024-04-29
Fix potential github action smells

ceddy4395:
Hey! 🙂
I want to contribute the following changes to your workflow:

 - Avoid executing  scheduled workflows on forks
 - Use names for run steps
 - Define permissions for workflows with external actions
 - Steps should only perform a single command
 - Use commit hash instead of tags for action versions
 
 These changes are part of a research Study at TU Delft looking at GitHub Action Smells. [Find out more](https://ceddy4395.github.io/research/gha-smells.html)

Closes #2411

PR #2429, opened 2024-05-01
Auto format/lint and revert some changes made to CI

HeavenVolkoff:
- Auto lint and format
 
 - Revert some mistake made to CI in #2412 and revert locking actions versions to commits hashes
	> While I agree with the intent of the mentioned PR (making Github CI actions less vulnerable to sudden changes), I am not sure hard-coding their commits SHAs is the best solution, especially considering the constant updates some of these actions receive to continue working properly under GitHub CI, which would require a non ideal amount of work from us to keep up to date if we had to keep updating commits hashes. I would like to revisit this whenever we have time to implement some sort of system that automatically manages dependencies version update such as renovatebot.com. Also the PR only changed a couple of CI actions to use commit hashes, and I think we should keep all CIs consistent

### DIFF
--- a/.github/workflows/cache-factory.yaml
+++ b/.github/workflows/cache-factory.yaml
@@ -54,11 +54,11 @@ jobs:
     runs-on: ${{ matrix.settings.host }}
     if: github.repository == 'spacedriveapp/spacedrive'
     permissions: {}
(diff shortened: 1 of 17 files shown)


utkubakir (approved):
/lgtm

PR #2433, opened 2024-05-01
[ENG-1690] Upgrade to Tauri 2.0

oscartbeaumont:
Issues:
 - [x] Window did not emit `app_ready` event fast enough. Showing window... - Bug in Tauri Specta w/ non-`Result` return types.
 - [x] Tauri drag region broken - Permission issue
 - [x] Upgrade menu to new system
 - [x] ~~Drag and drop for Spacedrop working again~~


HeavenVolkoff:
Tried to get this working on Linux. Fixed some problems that appeared during compilation. Code now compiles but it raises a panic immediately after launch, need to investigate further:
image

HeavenVolkoff:
> Can you reproduce the issue by creating a new Tauri project (`sh <(curl https://create.tauri.app/sh) --beta`) and enabling the same Cargo features?

Just tested and a new Tauri project with the same features as we have works fine.

Given the error message I got on Linux, it appears that we might be mishandling something during the core setup related to custom URI schema configuration. GTK is against running UI calls in any thread other than the main one. Now, considering the custom uri-schema feature on Tauri operates under the hood by configuring the native webview’s custom URI schema, and GTK's native webview is webkit2gtk, which consequently brings GTK assumptions to most calls for the webkit functions. It’s likely that we’re invoking a Tauri function for configuring the URI schema within a thread other than the main one, which could be triggering the panic.

HeavenVolkoff:
After some debugging I found out the problem stems from the `linux-ipc-protocol` feature. Specifically this code from `tauri-wry`:

https://github.com/tauri-apps/wry/blob/d7a89ec96d51f913835f062f481f346ba0c35e83/src/webkitgtk/web_context.rs#L367-L387

The call to `URISchemeResponse::new` causes a panic when it is executed from any thread other than main, which sometimes happens (don't know why) when tauri is doing IPC with the `linux-ipc-protocol` feature enabled.

For now the workaround is to disable `linux-ipc-protocol`, which has a slight impact on performance.

apps/desktop/src-tauri/gen/schemas/desktop-schema.json
HeavenVolkoff, line 1: Do we really need to commit these `gen/schemas` files? They are kinda large (2.5k lines each) and seem to be auto-generated. If not required it would be better to ignore them.

HeavenVolkoff (changes requested):

apps/desktop/src-tauri/gen/schemas/desktop-schema.json
HeavenVolkoff: Found an answer here, these files should be ignored:
https://github.com/tauri-apps/tauri/discussions/9051

oscartbeaumont:
Nice find.

Just to confirm a new project with `linux-ipc-protocol` works right?

I'm happy to merge this when you are.

HeavenVolkoff:
> Just to confirm a new project with `linux-ipc-protocol` works right?

Not sure, probably not. When I tested the new project earlier I didn't knew what was the cause for the panic, so I though the crash occurred during the app setup, but from what I found out debugging spacedrive later was that the crashes only happen if you actively uses the IPC connection to talk to tauri, as that is the moment the following code runs:

https://github.com/tauri-apps/wry/blob/d7a89ec96d51f913835f062f481f346ba0c35e83/src/webkitgtk/web_context.rs#L367-L387

The crash is directly tied to tauri trying to instantiate a `URISchemeResponse` (a native webkit2gtk object) which will fail if the thread running this closure is not the main one. I don't know why this closure sometimes runs outside the main thread sometimes though...

HeavenVolkoff:
> I'm happy to merge this when you are.

I think it can be merged now, Just running a release to see if the build system is working fine and to also do a final test on all OSes
https://github.com/spacedriveapp/spacedrive/actions/runs/8949548958

PR #2438, opened 2024-05-02
Don't error on missing search filters + add default search target

Brendonovich:
Stops search from crashing on favourites + recents
image


HeavenVolkoff (approved):
LGTM

PR #2439, opened 2024-05-02
[i18n]: add Arabic langauge

zaaakher:
This PR translates the app to Arabic and adds it as an option for users to switch to.

> [!NOTE]
> Arabic is a RTL language so we can perhaps (in the future) conditionally change the `dir` property of a parent html tag depending if a language is RTL or not (like Arabic, Persian, Urdu, and Hebrew). I haven't looked at the styling of the UI yet but If we're using direction based styling like `margin-right` then the `dir=rtl` will break the UI. A more universal way of doing that styling would be to replace `margin-right` with `margin-inline-start` so it will look consistently coherent regardless of `dir`


ameer2468:
Great contribution! thank you for the PR. As a native arabic speaker myself, the translations look great.

As for the layout - this is something we can address in the future.

ameer2468 (approved):

PR #2442, opened 2024-05-02
Added more translation keys to the app

artsiom-voitas:
* added more translation keys to the app
* translated all those keys to all languages available
* returned choice of shorten dates


interface/app/$libraryId/search/util.tsx
```diff
@@ -1,26 +1,27 @@
...
+		from: i18n.t('from'),
+		to: i18n.t('to')
 	},
 	trueOrFalse: {
-		true: 'is',
-		false: 'is not'
+		true: i18n.t('true'),
+		false: i18n.t('false')
 	}
 } as const;
```
utkubakir, line 26: first and last keys should be "is" and "is_not" instead

interface/app/$libraryId/Explorer/store.ts
```diff
@@ -150,24 +151,24 @@ export function isCut(item: ExplorerItem, cutCopyState: CutCopyState) {
...
 export const nonIndexedPathOrderingSchema = z.union([
-	z.literal('name').describe('Name'),
-	z.literal('sizeInBytes').describe('Size'),
-	z.literal('dateCreated').describe('Date Created'),
-	z.literal('dateModified').describe('Date Modified')
+	z.literal('name').describe(i18n.t('name')),
+	z.literal('sizeInBytes').describe(i18n.t('size')),
+	z.literal('dateCreated').describe(i18n.t('date_created')),
+	z.literal('dateModified').describe(i18n.t('date_modified'))
 ]);
```
utkubakir, line 174: does this change anything in the app? I don't think describe is used anywhere

utkubakir (approved):
/lgtm

PR #2453, opened 2024-05-03
Added even more i18n translation keys

artsiom-voitas:
* Added more i18n keys
* Synced those keys with every language available
* Translated those keys on Belarusian and Russian
* Added type safe function for translation of `ObjectKindEnum` values
* Removed pluralize function, added plural keys for required words for each language available
* Hided filters description for boolean filters


interface/app/$libraryId/Explorer/util.ts
```diff
@@ -139,3 +140,48 @@ export function generateLocaleDateFormats(language: string) {
 		return DATE_FORMATS;
 	}
 }
+
+const kinds: Record<string, string> = {
```
utkubakir, line 144: `<ObjectKindEnum,string>`

interface/app/$libraryId/search/AppliedFilters.tsx
```diff
@@ -131,6 +135,8 @@ export function FilterArg({ arg, onDelete }: { arg: SearchFilterArgs; onDelete?:
 }
 
 function pluralize(word?: string) {
-	if (word?.endsWith('s')) return word;
-	return `${word}s`;
+	if ((i18n.resolvedLanguage || i18n.language) === 'en') {
+		if (word?.endsWith('s')) return word;
+		return `${word}s`;
+	} else return word;
 }
```
utkubakir, line 142: there is a way to do this with i18n, you pass the length and it returns you the correct word

interface/app/$libraryId/settings/client/account.tsx
```diff
@@ -113,7 +114,7 @@ function HostedLocationsPlayground() {
 								className="grow"
 								value={locationName}
 								onInput={(e) => setLocationName(e.currentTarget.value)}
-								placeholder="My sick location"
+								placeholder={t('my_sick_location')}
```
utkubakir, line 117: yeah I wouldn't translate anything on the playground pages :D

interface/app/$libraryId/Explorer/util.ts
artsiom-voitas: > `<ObjectKindEnum,string>`

I'm not sure about that, I'm getting the error
image
With that function, we always use the already converted `ObjectKindEnum` value to a string

interface/app/$libraryId/search/AppliedFilters.tsx
artsiom-voitas: > there is a way to do this with i18n, you pass the length and it returns you the correct word

Trying to figure out how to handle this, because function in this component takes already translated value from the parent component, I have problems to find out how to pass original value on English, so I made this hack for now

utkubakir: https://www.i18next.com/translation-function/plurals is the link but yeah it does sound messy if the function takes the translated value.

artsiom-voitas: > https://www.i18next.com/translation-function/plurals is the link but yeah it does sound messy if the function takes the translated value.

image
Yeap, really messy

artsiom-voitas: > https://www.i18next.com/translation-function/plurals is the link but yeah it does sound messy if the function takes the translated value.

I will finish it tomorrow, I figure it out!

utkubakir (approved):
lgtm

PR #2459, opened 2024-05-07
Extend indexer rules with `.gitignore` when available

matheus-consoli:
Extend the file indexer to account for ignored files within a git repository.

---

# Example

Using this `.gitignore`:

```md
partial/*.ignoreme
secrets
```

and this file structure:

```sh
myfiles
├── file
├── partial
│  ├── file.ignoreme
│  └── readme.md
└── secrets
```

the files listed in the `.gitignore` are skipped by the indexer and does not appear in the frontend:

image
image

Closes #(issue)


0xBA5E64:
I'm personally not entirely sure how I feel about this; Having the indexer ***skip over*** certain files feels like it defeats any chance of Spacedrive being used as a true file-explorer replacement it's supposedly aims to be. A developer building an application being unable to even see their `dist/` folder due to it being part of the project's `.gitignore` for instance seems frustrating at best.

I could understand perhaps disabling higher-level metadata extraction or something similar for files falling under `.gitignore`, but fully excluding them from the index should probably be, at the very least, an ***option*** the user should have to set themselves upon creating the Location, as opposed to the default.

matheus-consoli:
I agree with you, @0xBA5E64 - `.gitignore` (and other possible sources of rules) should be an opt-out option for users.

I'm still quite new to this codebase, but I think the current indexer cannot express this configuration for now.

I would love to be pointed wrong though, and it's probably something that should be added to the roadmap

fogodev:
> I'm personally not entirely sure how I feel about this; Having the indexer _**skip over**_ certain files feels like it defeats any chance of Spacedrive being used as a true file-explorer replacement it's supposedly aims to be. A developer building an application being unable to even see their `dist/` folder or local dotenv files due to it being part of the project's `.gitignore` for instance seems frustrating at best.
> 
> `.gitignore` isn't _necessarily_ used to hide files, after all but rather just to keep them out of git's versioning, which may be done for a variety of reasons, including specifically for local-only files like dotenv's which very much might want to be accessed.
> 
> I could understand perhaps disabling higher-level metadata extraction or something similar for files falling under `.gitignore`, but fully excluding them from the index should probably be, at the very least, an _**option**_ the user should have to set themselves upon creating the Location, as opposed to the default.

The indexer rules aren't only for showing files in the explorer, for indexed locations they define what should or shouldn't be stored in db and syncronized between multiple devices. I agree that maybe excluding these files from the explorer may not be great, but usually files ignored by .gitignore, really must not be stored in db synchronized between multiple devices. Imagine having to store and sync a `node_modules` directory for example, LMAO.

fogodev:
A possible option, I've discussed internally this week, is about merging results from database and from file system when showing indexed locations in the explorer. This way we will index everything for storing and sync purposes according to indexer rules, but show everything contained in the file system, maybe with a small icon in files not in db for example.

HeavenVolkoff:
> merging results from database and from file system when showing indexed locations in the explorer. This way we will index everything for storing and sync purposes according to indexer rules, but show everything contained in the file system, maybe with a small icon in files not in db for example.

IMHO I think this is the best approach. The only question is if we handle this in the frontend or the backend. Technically we already have all the api call necessary for handling this entirely in the front, just need to call `search.ephemeralPaths` for the current path in the indexed location, and merge the results with those from the indexer. However I wonder if handling this in the backend wouldn't be better. It could check when the application is requesting a local indexed location and internally use the logic for the ephemeral location and merge the results and just send then to the front, considering we already handle Path and NonIndexedPath in a generic way it should work ootb.

core/crates/indexer-rules/src/seed.rs
```diff
@@ -12,8 +19,150 @@ pub enum SeederError {
...
+				// - rejecting all `path/to/docs/*.md` (including `path/to/docs/readme.md`)
+				//   this rule approves every file inside the git repo, except for the `docs/*.md` files
+				// - accepting `path/to/readme.md`
+				//   this REJECTS every file except for `docs/readme.md` (that is already rejected by other rule...)
+				//
+				// # Once fixed:
+				// line.remove(0); // pop the !
+				// let full = base_dir.join(line);
+				// let file = full.into_os_string().into_string().unwrap();
+				// RulePerKind::new_accept_files_by_globs_str([file]).unwrap()
```
HeavenVolkoff, line 96: While this is not implemented we should probably ignore include rules using `*` for paths that have an exclude rule. Because as it implemented rn, repositories that have a .gitignore such as:
```.gitignore
*
!src
```
which are fairly common, will be completely ignored by the indexer and nothing will show up for the user in the front-end, which is not the intended behavior.

I know the logic for this can become quite complex due to `.gitignore` hierarchical overrides, but I think only ignoring the current `.gitignore` star rules when a `!` rule is present should be enough for 90% of cases

HeavenVolkoff (changes requested):

core/crates/indexer-rules/src/seed.rs
matheus-consoli: good catch, thanks!
I made it ignore every rule `S` containing a `*` if `S` also matches a `!rule` or if the `!rule` matches `S` (the symmetry imports).

The solution is not particularly performant, as it now has to allocate and bookkeep more things and build the `Glob` object more times, but I don't think it will be too impactful, since gitignore files don't tend to be considerably long

HeavenVolkoff: Yeah, that is fine, as it is just a temporary solution also. Good job

0xBA5E64:
A file being listed in `.gitignore` only means for certain that the user doesn't want it as part of their git repo, it doesn't _necessarily_ mean it's disposable, let alone that a user doesn't want it synced across systems.

> The indexer rules aren't only for showing files in the explorer, for indexed locations they define what should or shouldn't be stored in db and syncronized between multiple devices.

Indeed, but I urge you to look at this in the reverse as well: The indexer rules don't just define what's to be synchronized between multiple devices, but also, what's being indexed and showed in the explorer.

> Imagine having to store and sync a node_modules directory for example

For things like web-development, `node_modules` is indeed a great example of a folder you'd likely not want to _sync_ in-between machines, but that might not be the case for all other files included in a `.gitignore` file, like [logs](https://github.com/github/gitignore/blob/a41112a9ab9c90d3cdbbe17f065bebeb622b3dc7/Node.gitignore#L1-L8), [`*.tgz` builds](https://github.com/github/gitignore/blob/a41112a9ab9c90d3cdbbe17f065bebeb622b3dc7/Node.gitignore#L70), [dotenv files](https://github.com/github/gitignore/blob/a41112a9ab9c90d3cdbbe17f065bebeb622b3dc7/Node.gitignore#L75-L80), local sqlite databases etc.

What about, say, game-devs, that occasionally leave out all their game assets out of version management via `.gitignore` due to size constraints or simple lack of understanding of Git LFS? Or, sensitive info like dotenv-files users might not want synced to a public repo, but users _might_ still want synced across your private devices, or at the very least, expect will be syncing when they are syncing entire locations through a tool like Spacedrive as opposed to simply `git clone`ing them on all their machines?

Excluding files from the indexer based on `.gitignore` might seem intuitive, but I can imagine a lot of people, including myself, find it quite off-putting if "syncing" a _**directory**_ through Spacedrive didn't actually sync the folder, as-is, but instead made opinionated decisions behind-the-scenes without my explicit knowing or intervention this way.

To be clear I do certainly find the value in something like this, I just feel that something like avoiding `node_modules` from syncing should be part of the sync-logic, as opposed to the indexing logic. And either way, be an optional feature listed in either the Location settings;
Screenshot 2024-05-10 at 10 49 51
-or future "Sync settings".

HeavenVolkoff:
> Indeed, but I urge you to look at this in the reverse as well: The indexer rules don't just define what's to be synchronized between multiple devices, but also, what's being indexed and showed in the explorer.

Yeah that is the intent, the sync system is being develop as an extension of the indexer. Under the hood the sync is mostly a way for sharing your indexed database between Spacedrive clients + the ability to request remote file data on-the-fly through a p2p connection. So, the indexer rules will influence what is being synced. About what is being displayed on the explorer we are discussing how to improve this, the idea is to integrate our indexed locations with the logic for the ephemeral file explorer, so that a local machine can seamlessly display all files for a location independently if they are actually being indexed or not. That way the indexed locations would work more like a normal file explorer with the added benefit of almost instantaneous search and sync for the indexed files. We would also like to implement something similar for the sync system, such as a client could remotely request an ephemeral view of a synced location to allow the user to see all the files on the remote Spacedrive, as long as the source is online.

> To be clear I do certainly find the value in something like this, I just feel that something like avoiding `node_modules` from syncing should be part of the sync-logic, as opposed to the indexing logic. And either way, be an optional feature listed in either the Location settings or future "Sync settings".

We need to find a balance so that the indexer, by default, can be performant enough while retaining most of its usability. Having good defaults for files being ignored at first is an important part of this, and ignoring files listed in `.gitignore` seems like a good default so far. However we are very much in favor of every bit of the indexer settings to be configurable by the user, so this rule (to ignore files in the `.gitignore` ) will be something toggleable by the user in the near future as well as any other indexer rule we add. About allowing specific files to be synced even when they are being ignored by some general indexer rule, we will need to implement a prioritization system to fully support the `.gitignore` syntax, with that in place it would be possible for the user to add a rule to force adding files that are being ignored by a more general rule, we could then expand the ContextMenu in the Explorer to allow the user to select non-indexed items and enable indexing, and consequently sync, for them by transparently creating such index rule. That way if a rule, such as `.gitignore`, removes some files that the user needs to be indexed/synced he could easily add them back.

HeavenVolkoff (commented):
Nice work. However some files ignored by the root `.gitignore` which are present in sub-directories are still being indexed. I think this can be fixed if we adapt the implementation of the following functions from `gix-ignore`:
[gix_ignore::search::Search::pattern_matching_relative_path](https://docs.rs/gix-ignore/latest/src/gix_ignore/search.rs.html#169-180)
[gix_ignore::search::pattern_matching_relative_path](https://docs.rs/gix-ignore/latest/src/gix_ignore/search.rs.html#107-138)

However I would be fine merging this as is and leaving the changes recommended above as an improvement for later if they are not something straight-forward to implement.

HeavenVolkoff (approved):
LGTM Great work @matheus-consoli

PR #2462, opened 2024-05-08
[ENG-293] Tag assign mode

### LINKED ISSUE #2350: [ENG-293] Tag assign mode
This feature makes assigning tags extremely simple and will be a prominent feature in the marketing material.

The tag button in the TopBar should hide the path view and instead show a line up of numbered tags, selecting some files and pressing that number will assign a tag to those files.

Another approach to this could be to have a floating tag picker UI that works in a similar fashion as above.

<sub>From [SyncLinear.com](https://synclinear.com) | [ENG-293](https://linear.app/spacedriveapp/issue/ENG-293/tag-assign-mode)</sub>

iLynxcat:
This PR introduces **tag assign mode** to Spacedrive.

## Changes:

- Add new **tag bar** to explorer which shows all tags
  - Click a tag and then press a key 1-9 to assign that tag a hotkey
  - With tag bar open, select files and press a hotkey assigned to a tag to assign that tag to the selected files
  - Click a tag with an assigned hotkey and press <kbd>Escape</kbd> to unassign its hotkey
- Adds <kbd>Cmd/Ctrl</kbd>+<kbd>Alt</kbd>+<kbd>T</kbd> shortcut for opening/closing tag bar
- Includes localization changes to support new tag bar UI

Closes #2350


interface/app/$libraryId/Explorer/ExplorerTagBar.tsx
```diff
@@ -0,0 +1,296 @@
...
+
+interface TagItemProps {
+	tag: Tag;
+	assignKey?: string;
+	isAwaitingKeyPress: boolean;
+	onKeyPress: (e: KeyboardEvent) => void;
+	onClick: () => void;
+}
+
+const TagItem = ({
```
myung03, line 242: this should be refactored to another file since this file is already huge

interface/app/$libraryId/Explorer/ExplorerTagBar.tsx
```diff
@@ -0,0 +1,296 @@
...
+	);
+
+	return (
+		<div
+			className={clsx(
+				'flex flex-col-reverse items-start justify-center border-t border-t-app-line bg-app/90 px-3.5 py-1 text-ink-dull backdrop-blur-lg ',
+				`h-[64px]`
+			)}
+		>
+			<em className="text-sm tracking-wide">{t('tags_bulk_instructions')}</em>
```
myung03, line 174: the italics look a bit strange, maybe just normal text is fine. But on that note I think this should be a tool tip instead, since it makes the tag bar unnecessarily large. Maybe putting a tooltip on the right hand side of the bar, or somewhere else that makes sense.

interface/locales/en/common.json
```diff
@@ -665,6 +669,11 @@
   "tag_one": "Tag",
   "tag_other": "Tags",
   "tags": "Tags",
+  "tags_bulk_assigned": "Assigned tag \"{{tag_name}}\" to {{file_count}} $t(file, { \"count\": {{file_count}} }).",
+  "tags_bulk_failed_with_tag": "Could not assign tag \"{{tag_name}}\" to {{file_count}} $t(file, { \"count\": {{file_count}} }): {{error_message}}",
+  "tags_bulk_failed_without_tag": "Could not tag {{file_count}} $t(file, { \"count\": {{file_count}} }): {{error_message}}",
+  "tags_bulk_instructions": "Select one or more files and press a key to assign the corresponding tag.",
```
myung03, line 675: this wording is rather misleading because it only allows you to assign number keys 1-9 to tags, so it might be beneficial to specify that

myung03: I didn't test with more than 9 tags though so I definitely may be wrong here!

myung03 (changes requested):
really cool functionality! just some styling issues and visual updates. lmk if you have any questions about any of my suggestions/if you disagree with any of them, would love to learn more about your thought process or explain any confusion :))

interface/locales/en/common.json
utkubakir: curious what does this part translates to: `$t(file, { \"count\": {{file_count}} })`. Looks weird to have it on i18n

iLynxcat: changes to the proper singular/plural form of "file"/"files" based on the file_count number

interface/app/$libraryId/Explorer/ExplorerTagBar.tsx
iLynxcat: I just think a bar of tags appearing mysteriously at the bottom with no upfront text is going to be very confusing for users who aren't familiar with the feature. I'd be open to adding an option to _hide_ the hint text or other ideas.

interface/app/$libraryId/Explorer/TopBarOptions.tsx
```diff
@@ -118,11 +119,28 @@ export const useExplorerTopBarOptions = () => {
 			showAtResolution: 'xl:flex'
 		},
 		{
-			toolTipLabel: t('key_manager'),
```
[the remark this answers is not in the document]
iLynxcat: I have no clue what happened here! Might have been a merge issue

interface/app/$libraryId/Explorer/ExplorerTagBar.tsx
ameer2468: I think the text for now is fine - altho I do vision in another iteration we can improve otherwise.

My last commit has made visual improvements that I think were needed.

Screenshot 2024-06-17 at 11 18 07 AM

interface/app/$libraryId/Explorer/ExplorerPathBar.tsx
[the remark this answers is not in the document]
ameer2468: I do like this, but i'm against this due to native experiences not performing animations for these features.

Meaning when you toggle path bar, it just appears/disappears, and it should be the same or consistent with tag bar too.

We can always challenge native experiences and do things our own way, atleast for this. Let's give it some thought in a separate PR.

interface/app/$libraryId/Explorer/ExplorerTagBar.tsx
iLynxcat: I worry about visual balance here, having centred content in the tag bar and left-aligned content in the path bar looks odd. Curious what @jamiepine's vision on this is, but we can iterate while waiting on his response

interface/app/$libraryId/Explorer/ExplorerPathBar.tsx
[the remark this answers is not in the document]
iLynxcat: Worth noting that I do plan on reworking the style of this bottom bar area to allow for quick animations, but as it stands it would be janky and not work too well given the main explorer list depends on fixed heights to offset content — this will not only mess with animations (which would be fine to work around with small changes) but also cause a large accessibility issue if users wish to use much larger text size within the UI. It's not something that is considered well in the current UI but will be important toward the release stage

interface/app/$libraryId/Explorer/ExplorerTagBar.tsx
iLynxcat: Agreed, and I will move this off to another file — I should note that I want to do some restructuring of this area of the codebase as the Explorer directory is a total mess without any organization

interface/locales/en/common.json
iLynxcat: @myung03 I agree with you but I can't think of a way that looks/feels good to communicate that explicitly right now. I think this is something that causes very little friction one single time to a user who may not realize it. Once they try any key other than 1-9 and it fails, then they try a key 1-9 and it succeeds, it'll be easily understood.

I do agree we chould communicate it better when we revise the help text design overall. Maybe an info icon popup appended to the help text?

interface/app/$libraryId/Explorer/ExplorerTagBar.tsx
ameer2468: > I worry about visual balance here, having centred content in the tag bar and left-aligned content in the path bar looks odd. Curious what @jamiepine's vision on this is, but we can iterate while waiting on his response

Valid point - i'm gonna play with it and see what better can be done.

ameer2468: I've experimented with the layout a bit - i see potential in this. So i'm gonna leave it and see what other ideas can be shared.

Screenshot 2024-06-17 at 4 14 45 PM

interface/locales/en/common.json
myung03: Totally agree - not sure if this is sufficient, but I would probably just change it to "Select one or more files, and press a number key to assign to the corresponding tag." 

0 not being an option makes a marginal difference and I think just adding that extra bit of wording makes it a lot clearer (instead of having the user have to press a random key and trial + error until they realize that only number keys work). This is a bit of a personal preference thing though so lmk if you disagree :))

interface/app/$libraryId/Explorer/ExplorerTagBar.tsx
myung03: i like this layout, but i can see it becoming an issue once the user has 8 or more tags as it will overlap with the text. can we have some kind of side scroller logic for tags if there are more than X amount?

myung03: also I know we said the text was fine, but would it be possible to make it a dismissible component (similar to the "Meet Grid View" thing when you open up a folder). I feel like this would be the best of both worlds so the tags dont appear with no text but once the user closes the text there will be more space for the tags.

ameer2468 (approved):

PR #2465, opened 2024-05-09
Improved Japanese translation

wany-oh:
I have translated the newly added items and corrected the mistranslations caused by the existing machine translation.


utkubakir (approved):
Thank you!

PR #2470, opened 2024-05-10
Revert untranslated keys added to locales

HeavenVolkoff:


interface/locales/it/common.json
```diff
@@ -587,9 +582,7 @@
   "system": "Sistema",
   "tag": "Tag",
   "tag_one": "Tag",
-  "tag_many": "Tags",
   "tag_other": "Tags",
-  "tags": "Tags",
```
HeavenVolkoff, line 592: Should we include these @utkubakir? They weren't originally in this translation file, but they have the same values as the others already included...

HeavenVolkoff: Or should we remove the other non translated keys as well

utkubakir: yeah this is a bit weird, If I remember correctly, you only need tag_one and tag_other if the count value is not changing the pluralization. I say remove it it's a duplicate.

https://www.i18next.com/translation-function/plurals

artsiom-voitas:
Please, check General channel in Discord, before merging! 
"tag_" keys should stay

HeavenVolkoff:
> Please, check General channel in Discord, before merging! "tag_" keys should stay

Ok, will restore the tag keys, thanks for the warning @artsiom-voitas

ameer2468 (approved):

PR #2472, opened 2024-05-10
Fix macOS bugs

HeavenVolkoff:
- [X] Fix macOS resize bug by replacing custom macOS blur effect with Tauri's V2 native impl
 - [X] Fix open with functionality
 - [X] Fix app breaking on dev when reloading the webview


fogodev (approved):
Tested here, it seems to be working fine.

PR #2473, opened 2024-05-10
Update image crate

fogodev:
Bump image crate version

Also had to patch pdfium-render to bump its image crate version too, will revert once my bump PR on their repo is merged.


HeavenVolkoff (approved):
LGTM

PR #2476, opened 2024-05-11
[ENG-1775] Scan location using new jobs

fogodev:
Integrating the new jobs system to scan locations, with indexer, file identifier and media processor jobs, (and also their shallow versions), also solving a bunch of bugs that appeared due to real usage of the job system and the task system. 

Aside from that, I also did some maintenance in the location manager and location watcher, and normalized logging using more `tracing` features through all core and its inner subcrates.


jamiepine (approved):
After thorough testing everything seems fine, lets merge!

PR #2477, opened 2024-05-12
[ENG-1778] Total library size statistic

jamiepine:
This PR adds a total for the size of the library, the combination of all Location's size value. It also fixes the naming scheme of the statistics table.


jamiepine:
not sure what happened with i18n here, but I added two keys that require translations.

utkubakir:
> not sure what happened with i18n here, but I added two keys that require translations.

looks like it got sorted by the extension which is ok

myung03 (approved):

myung03 (approved):

ameer2468 (approved):

packages/client/src/lib/humanizeSize.ts
```diff
@@ -126,11 +126,10 @@ export const humanizeSize = (
 		unit.from === 0n
 			? Number(bytes)
 			: Number((bytes * BigInt(precisionFactor)) / unit.from) / precisionFactor;
-	const plural = use_plural && value !== 1 ? 's' : '';
 
 	return {
-		unit: (is_bit ? BYTE_TO_BIT[unit.short as keyof typeof BYTE_TO_BIT] : unit.short) + plural,
-		long: (is_bit ? BYTE_TO_BIT[unit.long as keyof typeof BYTE_TO_BIT] : unit.long) + plural,
```
HeavenVolkoff, line 133: Why remove the plural logic @jamiepine? It could be disabled when calling the function: `humanizeSize(size, { use_plural: false })`

PR #2480, opened 2024-05-13
i18n Keys for Size values and partial Job Manager translation

artsiom-voitas:
* added keys for size translation
* added couple of keys for Job Manager, which isn't hard to get with string methods of TS from `packages` directory.
* added keys for indexer rules in the settings
* added keys for metadata in Inspector
* minor translation fixes
* minor update of `StatCard` style


interface/app/$libraryId/Layout/Sidebar/JobManager/Job.tsx
```diff
@@ -51,19 +53,19 @@ function Job({ job, className, isChild, progress }: JobProps) {
...
-							'The job completed with errors. Please see the error log below for more information. If you need help, please contact support and provide this error.',
+						title: t('error'),
+						description: t('job_error_description'),
 						children: JobError
 					});
 				}
 			}
 		]);
 	}
+	console.log(job.status);
```
ameer2468, line 68: Let's remove this console log

interface/app/$libraryId/settings/library/locations/IndexerRuleEditor/index.tsx
```diff
@@ -115,6 +115,8 @@ export default function IndexerRuleEditor<T extends IndexerRuleIdFieldType>({
 								editable
 									? (e) => {
 											e.stopPropagation();
+											console.log(rule);
```
ameer2468, line 118: This one too

interface/app/$libraryId/overview/FileKindStats.tsx
ameer2468, line 1: Don't we also need formatNumber here? because I see you removed it.

utkubakir (approved):

PR #2483, opened 2024-05-14
Faster debug builds

matheus-consoli:
Config the dev profile for faster debug builds. That's a simple change, but I found it to be quite good for dev workflow.

A clean `pnpm tauri dev` went from 5m to 3m20s (better, but I'm certain we can improve), and it's even better when during small iterations of changes-and-check, or change-and-test (when the incremental changes start doing its thing).

The downside of these changes is that we lose the debug symbols (**only in debug builds, this does not affect release builds**)


HeavenVolkoff (approved):
LGTM

PR #2484, opened 2024-05-14
Added Peer screen

jamiepine:
Screenshot 2024-05-13 at 9 50 36 PM

Spacedrop dropzone doesn't actually work yet


interface/util/hardware.ts
myung03, line 1: just a thought - is there are way for us to rename the icon png files or another dynamic method to fully avoid the switch statements (i.e Rename MiniSilverBox.png to just MacMini.png). I think the switch statement would be difficult to maintain as the number of device types increase.

myung03 (approved):

PR #2494, opened 2024-05-17
[MOB-95] Tags support & more

ameer2468:

- Various UI fixes, in filters page for scrolling locations and tags, and in locations and tags pages.
- Add/remove tags from objects
- Tags show in the explorer of Grid/list views
- Clicking on search from a location or a tag will add the appropriate search filter
- List view visual improvements
- Overview, library size added.
- Ericson gave me Rust code to make tags data is sent to FE correctly.
- Tags explorer is now working - however with useObjects query we get orphan objects **(Ericson thinks so)**, so I switched to Path for now, and it works fine.

**Issues**: 

Spoke to @fogodev about this - when you initially add a location on mobile or desktop, and then you try to tag a Folder/dir specifically, the update will not show immediately unless you refresh that location. There's an issue on the BE.

Frame 1


utkubakir (dismissed):
love it

utkubakir (approved):

PR #2498, opened 2024-05-20
Remove all recursion_limit

matheus-consoli:
These clauses don't seem necessary anymore


matheus-consoli:
I'm closing this PR as the iOS build needs the recursion limit increased (it's related to the Prisma machinery)
At least, now we know that the other platforms don't require this change.

PR #2500, opened 2024-05-21
Package & distribute frontend bundle when doing releases for Linux

HeavenVolkoff:
This is a required step for future work regarding new Linux releases types

Closes: ENG-1783


matheus-consoli (approved):
LGTM

PR #2506, opened 2024-05-24
[MOB-98] Rename, delete, & more

ameer2468:

- Several visual adjustments, making some elements smaller as they were too big.
- Browse screen location and tag sections redesigned
- Error handling for some actions **(will do a few more in another PR)**
- Rename and delete support
- Toast messages redesign

**Notes**:

There's currently issues with the Watcher which I believe Arnab is working on, some updates are not being reflected correctly.

Frame 1

### DIFF
--- a/apps/mobile/babel.config.js
+++ b/apps/mobile/babel.config.js
@@ -23,6 +23,7 @@ module.exports = function (api) {
 					}
 				}
 			]
-		]
+		],
+		overrides: [{ test: /\.solid.tsx$/, presets: ['solid'] }]
 	};
 };
--- a/apps/mobile/babel.config.js
+++ b/apps/mobile/babel.config.js
@@ -23,7 +23,6 @@ module.exports = function (api) {
 					}
 				}
 			]
-		],
-		overrides: [{ test: /\.solid.tsx$/, presets: ['solid'] }]
+		]
 	};
 };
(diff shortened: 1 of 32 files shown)


utkubakir:
is CI job relevant here?

ameer2468:
> is CI job relevant here?

I just fixed the test after running it locally. Also removed a SolidJS reference + finally fixed the annoying warning message that would see on initial render in onboarding and in the app.

utkubakir:
> > is CI job relevant here?
> 
> I just fixed the test after running it locally. Also finally fixed the annoying warning message that you would see on initial render in onboarding and in the app.

How? What was it lol

ameer2468:
> > > is CI job relevant here?
> 
> > 
> 
> > I just fixed the test after running it locally. Also finally fixed the annoying warning message that you would see on initial render in onboarding and in the app.
> 
> 
> 
> How? What was it lol
> 
> 
> 
> I am not a fan of these multiple useEffects to be honest, those functions are supposed to only run once on the app start but that might not be the case here

That's what the warning was about. You cannot run that code unless it's in a effect. The reason is it should only be run when buildInfo data is present. It's to prevent it from running during the rendering phase.

In that case, there's no harm in using effects, as that's what it should have been in the start.

utkubakir (approved):

PR #2507, opened 2024-05-25
added feedback for delete

myung03:
Added feedback for when user deletes a location

### DIFF
--- a/interface/locales/en/common.json
+++ b/interface/locales/en/common.json
@@ -1,5 +1,4 @@
 {
-  "Location added successfully": "Location added successfully.",
   "about": "About",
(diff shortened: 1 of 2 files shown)


interface/locales/en/common.json
```diff
@@ -1,4 +1,5 @@
 {
+  "Location added successfully": "Location added successfully.",
```
utkubakir, line 2: missed this

interface/locales/en/common.json
```diff
@@ -388,7 +389,9 @@
   "local_locations": "Local Locations",
   "local_node": "Local Node",
   "location": "Location",
+  "location_added_successfully": "Location added successfully. ",
```
utkubakir, line 392: extra space

utkubakir (changes requested):

interface/locales/en/common.json
myung03: having issues with adding the toast to location actually - trying to add the toast.success in the try part of the try catch but not working? any idea why

utkubakir: not sure where. happy to move this discussion to slack with more information on where in the code

utkubakir (approved):

PR #2513, opened 2024-05-28
Default Delete Action Function

Rocky43007:
Allows users to chose which delete function they would like to have default for CMD/CTRL + Delete/Backspace:

- Show Prompt
- Move to Trash
- Delete Instantly


Rocky43007:
Now being implemented by #2833

PR #2517, opened 2024-05-29
Disable AppNap for macOS and fix app version in deb

HeavenVolkoff:
- Disable app nap on app launch on macOS, as this was negatively affecting file indexer when the app window was in the background
 - Remove version from tauri.conf.json so tauri fallback to the version defined in Cargo.toml
 - Probably fix updater asking to update when you already have (same problem as above)


fogodev (approved):
Great, this damn App Nap was F*ing my new parallel job system

PR #2524, opened 2024-05-30
Serve thumbnails over P2P

oscartbeaumont:
This PR builds on top of #2523 to allow thumbnails to be served via P2P from the node where the file exists.

Although we will sync preview media, having the ability to serve it over P2P means we can suffice before the first sync and we can also support scenarios were only a portion of the preview media is sunk.

### DIFF
--- a/core/src/custom_uri/mod.rs
+++ b/core/src/custom_uri/mod.rs
@@ -1,7 +1,9 @@
 use crate::{
 	api::{utils::InvalidateOperationEvent, CoreEvent},
 	library::Library,
-	object::media::old_thumbnail::WEBP_EXTENSION,
+	object::media::old_thumbnail::{
+		get_ephemeral_thumb_key, get_indexed_thumb_key, WEBP_EXTENSION,
+	},
 	p2p::operations::{self, request_file},
 	util::InfallibleResponse,
 	Node,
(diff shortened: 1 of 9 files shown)


core/src/custom_uri/mod.rs
```diff
@@ -222,13 +222,56 @@ async fn get_or_init_lru_entry(
...
+			"/thumbnail/:lib_id/:cas_id",
 			get(
 				|State(state): State<LocalState>,
-				 extract::Path(path): extract::Path<String>,
+				 extract::Path((library_id, cas_id)): extract::Path<(String, String)>,
 				 request: Request<Body>| async move {
+					let (path, home_node) = if library_id == "ephemeral" {
+						(get_ephemeral_thumb_key(&cas_id), None)
+					} else {
+						let library_id = Uuid::from_str(&library_id).map_err(bad_request)?;
```
oscartbeaumont, line 233: By running these database queries within the file serving code, this PR significantly slows the loading of all preview media so we might want to do this out of band.

interface/app/$libraryId/Explorer/FilePath/Thumb.tsx
```diff
@@ -67,9 +67,9 @@ export const FileThumb = forwardRef<HTMLImageElement, ThumbProps>((props, ref) =
 
 		if (thumbType === 'thumbnail')
 			if (
-				loadState.thumbnail !== 'error' &&
-				itemData.hasLocalThumbnail &&
-				itemData.thumbnailKey.length > 0
+				loadState.thumbnail !== 'error' // &&
```
oscartbeaumont, line 70: Without this check disabled the node will just fallback to the placeholder instead of making a request over P2P to the other node for the file.

Obviously disabling it like i've done here is not a permanent solution.

interface/app/$libraryId/Explorer/FilePath/Thumb.tsx
```diff
@@ -87,8 +87,8 @@ export const FileThumb = forwardRef<HTMLImageElement, ThumbProps>((props, ref) =
 				break;
 
 			case 'thumbnail':
-				if (itemData.thumbnailKey.length > 0)
-					return platform.getThumbnailUrlByThumbKey(itemData.thumbnailKey);
+				if (itemData.casId)
+					return platform.getThumbnailUrlByThumbKey(library.uuid, itemData.casId);
```
oscartbeaumont, line 91: With the current implementation this can be called as:
 - `platform.getThumbnailUrlByThumbKey("ephemeral", itemData.casId)`
 - `platform.getThumbnailUrlByThumbKey(library.uuid, itemData.casId)`

As the codebase has no calls using the `"ephemeral"` variant I am certain this PR breaks thumbnails for the ethemeral explorer.

oscartbeaumont (commented):
As I finish at Spacedrive tomorrow i'm not going to be able to drive this PR to completion but i've solved the P2P part making it easy for someone else to pick up if they can fix the thumbnail handling parts.

jamiepine:
lets get this picked up @ameer2468

PR #2535, opened 2024-06-01
Cleanup: remove unnecessary #[allow(dead_code)] and #[allow(unused)] attributes

julianbraha:
I found 4 instances of `#[allow(unused)]` and 6 instances of `#[allow(dead_code)]` across 5 source files, where the code was not actually dead or unused. This PR removes the usage of these unnecessary attributes.


matheus-consoli (approved):
nice, thank you!

PR #2538, opened 2024-06-04
[ENG-1159] Report progress when copying files

matheus-consoli:
This fixes the reporting of progress when copying files.

The initial idea was to split the file into chunks and, at every couple of chunks transmitted, we would emit an event reporting the progress of the current copy.

But that proved a poor approach and didn't pay for its unnecessary complexity.

The current implementation (WIP), uses a less intrusive metric: spawn jobs for `tokio::fs::copy` the file; and another job that periodically checks the destiny file's metadata and estimates the progress - this is less precise but more efficient.


packages/client/src/utils/jobs/useJobInfo.tsx
```diff
@@ -258,11 +258,14 @@ export function useJobInfo(job: Report, realtimeUpdate: JobProgressEvent | null)
...
-				name: `${isQueued ? 'Copy' : isRunning ? 'Copying' : 'Copied'} ${
-					isRunning ? completedTaskCount + 1 : completedTaskCount
-				} ${isRunning ? `of ${job.task_count}` : ``} ${plural(job.task_count, 'file')}`,
-				textItems: [[{ text: job.status }]]
+				name: isQueued
+					? `Copy ${taskCount} ${plural(taskCount, 'file')}`
+					: isRunning
+						? `Copying ${completedTaskCount}% of ${realtimeUpdate?.info} ${plural(taskCount, 'file')} (${humanizeSize(parseInt(realtimeUpdate?.message || '0'))})`
+						: `Copied ${taskCount} ${plural(taskCount, 'file')}`,
+				textItems: [[{ text: realtimeUpdate?.phase }], [{ text: job.status }]]
```
myung03, line 266: Instead of using the word "Copy" I think it would make more sense to say "Duplicate", since thats the button that you press to start the command.

myung03 (commented):
mostly backend changes so just took a look at your frontend code, looks great!

packages/client/src/utils/jobs/useJobInfo.tsx
matheus-consoli: hmm, I don't have a strong opinion on this, but I just wanted to point out that there are two buttons, one called "Duplicate" the file in the same directory (effectively, a shortcut for copy-and-paste), and one called "Copy" for copying the files so the user can paste it elsewhere - so it sounds a bit ambiguous? idk
image

I updated the message, thanks

HeavenVolkoff (approved):
Tested, great work @matheus-consoli

PR #2552, opened 2024-06-14
[MOB-94] Onboarding haptics

ameer2468:
**This PR**: Simples adds haptics when tapping key buttons for each screen.


Rocky43007 (approved):
Looks good!

PR #2553, opened 2024-06-17
Streamline conversion from Key to GenericArray

julianbraha:
This constructs the `GenericArray` directly, instead of using an iterator. 

And resolves a TODO comment.


jamiepine (approved):
lgtm

PR #2555, opened 2024-06-18
[ENG-1794] Overview Rework

myung03:
This PR reworks the overview section, reorganizing current information into a more visually appealing format for users. 
- LibraryStats has been reworked into a storage bar 
- FileKindStats has been reworked into a bar graph 
- NodeGraph commented out and placed in Network for future use. This component is planned to showcase active nodes for a library. 
- minor refactor for previous library stats to clean up component, and overview index component (previously had a lot pointless dummy data commented out, removed for clarity).


core/src/api/libraries.rs
```diff
@@ -131,9 +131,16 @@ pub(crate) fn mount() -> AlphaRouter<Ctx> {
...
 				statistics: Vec<KindStatistic>,
+				total_identified_files: i32,
+				total_unidentified_files: i32,
 			}
 			R.with2(library()).query(|(_, library), _: ()| async move {
 				let mut statistics: Vec<KindStatistic> = vec![];
+
+				let total_unidentified_files = library.db.file_path().count(vec![]).exec().await?;
+
+				let total_identified_files = library.db.object().count(vec![]).exec().await?;
```
iLynxcat, line 142: tiny change, I think this whitespace isn't necessary

```suggestion
				let total_unidentified_files = library.db.file_path().count(vec![]).exec().await?;
				let total_identified_files = library.db.object().count(vec![]).exec().await?;
```

interface/app/$libraryId/network.tsx
```diff
@@ -72,3 +72,265 @@ export const Component = () => {
 		</ExplorerContextProvider>
 	);
 };
+
+// NOTE -> this is code for the node graph. The plan is to implement this in network (moved from overview page). Jamie asked me to save the code somewhere
+// so placing it here for now!
```
iLynxcat, line 77: Can we put this component in its own file?

Is there any reason it wouldn't compile if uncommented, or can we uncomment it?

interface/app/$libraryId/overview/FileKindStats.tsx
```diff
@@ -1,99 +1,201 @@
...
+import { getIcon } from '@sd/assets/util';
+import { useLibraryQuery } from '@sd/client';
+import { useNavigate } from 'react-router';
 import { useLocale } from '~/hooks';
+import { Info } from '@phosphor-icons/react';
+import { motion } from 'framer-motion';
+
+const INFO_ICON_CLASS = "inline size-3 text-ink-faint opacity-0";
+const TOTAL_FILES_CLASS = "flex items-center justify-between whitespace-nowrap text-sm font-medium text-ink-dull mt-2 px-1";
+const UNIDENTIFIED_FILES_CLASS = "relative flex items-center text-xs text-ink-faint";
```
iLynxcat, line 13: We should be a little more explicit that this is a 'list' of classes

```suggestion
const INFO_ICON_CLASSLIST = "inline size-3 text-ink-faint opacity-0";
const TOTAL_FILES_CLASSLIST = "flex items-center justify-between whitespace-nowrap text-sm font-medium text-ink-dull mt-2 px-1";
const UNIDENTIFIED_FILES_CLASSLIST = "relative flex items-center text-xs text-ink-faint";
```

interface/app/$libraryId/overview/FileKindStats.tsx
```diff
@@ -1,99 +1,201 @@
...
+  const BARCOLOR_END = '#004C99';
+
+  const formatCount = (count: number) => (count >= 1000 ? (count / 1000).toFixed(0) + 'k' : count.toString());
+
+  const handleResize = useCallback(() => {
+    if (containerRef.current) {
+      const factor = window.innerWidth > 1500 ? 0.35 : 0.4;
+      setCardWidth(window.innerWidth * factor);
+    }
+  }, []);
```
iLynxcat, line 58: I think we can ship an alpha release with this right now, but we should revise later use flexbox to automatically handle sizing as needed. Is there a particular reason that wouldn't work and we'd need to keep this in JS, or do you think we can work toward that later on? 

I specifically make this comment because we have some very unperformant views in the UI as-is and I'd like to make everything as natively styled as possible when it comes to layout.

interface/app/$libraryId/overview/FileKindStats.tsx
```diff
@@ -1,99 +1,201 @@
...
+    }
+  }, [data, isDark]);
+
+  const sortedFileKinds = [...fileKinds].sort((a, b) => b.count - a.count);
+  let maxFileCount: number;
+  if (sortedFileKinds && sortedFileKinds[0]) {
+  maxFileCount = sortedFileKinds.length > 0 ? sortedFileKinds[0].count : 0;
+  }
+
+  const getPercentage = (value: number) => `${((value / maxFileCount) * BARHEIGHT).toFixed(2)}px`;
```
iLynxcat, line 109: This should be moved out of the component and refactored to take in maxFileCount and BARHEIGHT as parameters. The function should also be given a more explicit name even if it's longer — maybe we call the function `getBarHeightPixels`, and rename `BARHEIGHT` to `STAT_BAR_MAX_HEIGHT` or similar.

interface/app/$libraryId/overview/LibraryStats.tsx
```diff
@@ -2,128 +2,167 @@ import { Info } from '@phosphor-icons/react';
...
 
 interface StatItemProps {
-	title: string;
-	bytes: bigint;
-	isLoading: boolean;
-	info?: string;
+  title: string;
+  bytes: bigint;
+  isLoading: boolean;
+  info?: string;
```
iLynxcat, line 13: Please make sure formatting with Prettier is on in your editor or double check it before each commit — these lines should start with a tab (`\t`) and not two spaces

interface/app/$libraryId/overview/StorageBar.tsx
```diff
@@ -0,0 +1,106 @@
+import React, { useState } from 'react';
+import { Tooltip } from '@sd/ui'; // Ensure you import your Tooltip component correctly
+import { humanizeSize } from '@sd/client';
+import { useIsDark} from '~/hooks';
+
+//change this to dynamically change the width of the bar (needed since each section's width is calculated independently)
+const BARWIDTH = 520
+
+// Function to lighten color
```
iLynxcat, line 9: This comment is redundant given the function name is `lightenColor`

```suggestion
```

interface/app/$libraryId/overview/StorageBar.tsx
```diff
@@ -0,0 +1,106 @@
...
+  const amt = Math.round(2.55 * percent);
+  const R = (num >> 16) + amt;
+  const G = ((num >> 8) & 0x00ff) + amt;
+  const B = (num & 0x0000ff) + amt;
+  return `#${(
+    0x1000000 +
+    (R < 255 ? (R < 1 ? 0 : R) : 255) * 0x10000 +
+    (G < 255 ? (G < 1 ? 0 : G) : 255) * 0x100 +
+    (B < 255 ? (B < 1 ? 0 : B) : 255)
+  )
```
iLynxcat, line 21: I always love to see bitwise logic 😋

iLynxcat, line 21: I always love to see bitwise logic 😋

interface/app/$libraryId/overview/index.tsx
```diff
@@ -16,6 +16,8 @@ import NewCard from './NewCard';
 import StatisticItem from './StatCard';
 
 export const Component = () => {
+
+
```
iLynxcat, line 20: Extraneous whitespace should not have been added

```suggestion
```

iLynxcat (changes requested):
Lots of small suggestions here, overall good looking as we discussed during our review session with @ameer2468 :)

**Only big issue:** Lots of lines here are not formatted per our Prettier config. Please make sure to use the Prettier vscode plugin or run `pnpm prettier --write .` before pushing changes!

interface/app/$libraryId/overview/StorageBar.tsx
myung03: imagine me getting a 4 year computer science degree to learn some mf bitwise logic 😞

interface/app/$libraryId/network.tsx
myung03: you are totally right I did not think of that - any suggestions for where to put it?

interface/app/$libraryId/overview/FileKindStats.tsx
myung03: There are a number of reasons I chose to this approach initially, but open to refactoring: 
- the gap is directly dependent on the width of the bars, as well as the number of bars. I thought it would be a better idea to manually calculate the bar width/other related css qualities and it would provide more control over the spacing and stuff, but this design decision may be poor as I just didn't really have any other good ideas of how to accomplish this 
- the number of bars is variable, and I wasn't sure if flexbox would be effective in cases where we had 20+ bars or other edge cases.

interface/app/$libraryId/overview/LibraryStats.tsx
myung03: fixed - prettier is enable though so im not sure why this doesn't work?

interface/app/$libraryId/overview/FileKindStats.tsx
myung03: I think if I abstract out maxFileCount/bar height I would have to move the entire useEffect out into the index file, which might obfuscate the code a bit (since its all related to the file). Any suggestions on how to best achieve the refactor? 
- also opting to simply remove bar height variable since its just in 1 place, so doesn't really make sense imo to use a global variable for that

interface/app/$libraryId/overview/LibraryStats.tsx
HeavenVolkoff: Also make sure to install the [EditorConfig](https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig) extension to autoconfigure vscode to respect our project spacing configurations defined in [.editorconfig](https://github.com/spacedriveapp/spacedrive/blob/main/.editorconfig)

interface/app/$libraryId/overview/FileKindStats.tsx
iLynxcat: it's doable with flexbox (or honestly maybe grid), happy to talk with you more in depth on layout options here!

interface/app/$libraryId/overview/LibraryStats.tsx
iLynxcat: > fixed - prettier is enable though so im not sure why this doesn't work?

you may have to manually configure it to autoformat on save. I can help you with checking your vscode config, gimme a slack message at some point if you want

myung03: yes remind me about this tmr please!

interface/app/$libraryId/overview/FileKindStats.tsx
```diff
@@ -1,99 +1,241 @@
...
-					{items !== undefined && (
-						<p className="text-xs text-ink-faint">
-							{t('item_with_count', { count: items })}
-						</p>
-					)}
+				<div className={TOTAL_FILES_CLASSLIST}>
+					<Tooltip className="flex items-center" label={t('bar_graph_info')}>
+						<div className="flex items-center gap-2">
+							<span
+								className={`${isDark ? 'text-white' : 'text-black'} text-xl font-black`}
```
ameer2468, line 147: We use **clsx** for conditional styling, please use clsx instead.

interface/app/$libraryId/overview/StorageBar.tsx
```diff
@@ -0,0 +1,112 @@
...
+import { humanizeSize } from '@sd/client';
+import { Tooltip } from '@sd/ui'; // Ensure you import your Tooltip component correctly
+
+import { useIsDark } from '~/hooks';
+
+//change this to dynamically change the width of the bar (needed since each section's width is calculated independently)
+const BARWIDTH = 520;
+
+const lightenColor = (color: string, percent: number) => {
+	const num = parseInt(color.replace('#', ''), 16);
```
ameer2468, line 11: While this is a solution that works - why not instead use tailwind color classes? blue-400, blue-500 etc...

I think all this code to support a lighter color is un-necessary.

interface/app/$libraryId/overview/FileKindStats.tsx
iLynxcat: got this in my local working copy, will be included in a commit soon :)

myung03: fixed by Haden!

interface/app/$libraryId/overview/StorageBar.tsx
myung03: I wanted it to be dynamic so in the future we wouldn't have to change the hover colour everytime a colour needed to be changed. This was more impactful at first (i.e when light and dark mode had different colours) but less obvious now. I talked to Haden and he said it should be fine for now, but we can definitely refactor in the future. I've left a note in the code for future change (if thats okay)?

ameer2468: > I wanted it to be dynamic so in the future we wouldn't have to change the hover colour everytime a colour needed to be changed. This was more impactful at first (i.e when light and dark mode had different colours) but less obvious now. I talked to Haden and he said it should be fine for now, but we can definitely refactor in the future. I've left a note in the code for future change (if thats okay)?

Yep that sounds great! and I understand your reason now.

ameer2468 (approved):

PR #2558, opened 2024-06-19
Improve Chinese translation

Whisht:
I was revised some improper translation and added some translation for the missing items.

Some other items may also need amendment, but I cannot determine the preferable translation without the context (for some items, I do not know the precise interface location of them).


ameer2468 (approved):

PR #2567, opened 2024-06-26
i18n polish for zh-cn on some terms of ui

HeavySnowJakarta:
i've noticed it's better to commit as soon as possible. once too many revises are made it may causes conflicts


ameer2468 (approved):

PR #2569, opened 2024-06-26
[MOB-72] Tracing logs not working on Android

Rocky43007:
This PR fixes the issue of getting logs from the rust core on Android. Currently, when building for Android, logs will stop showing up after the `Spacedrive Online!` message.

Now, logs will go to `adb logcat` instead of the logfile. Please run `adb logcat | grep -i com.spacedrive.app` when looking at logs, in order to only see logs from the Spacedrive Rust core.

### DIFF
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -262,7 +262,7 @@ impl Node {
 					.with_file(true)
 					.with_line_number(true)
 					.with_ansi(false)
-					.with_target(false)
+					.with_target(true)
 					.with_writer(logfile)
 					.with_filter(EnvFilter::from_default_env()),
 			)
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -258,16 +258,14 @@ impl Node {
 
 		let registry = registry();
 
-		#[cfg(target_os = "android")]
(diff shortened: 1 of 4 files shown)


core/src/lib.rs
```diff
@@ -235,11 +235,13 @@ impl Node {
...
-			} else {
-				"info"
-			};
+			// let level = if cfg!(debug_assertions) {
+			// 	"debug"
+			// } else {
+			// 	"info"
+			// };
+
+			let level = "debug";
```
fogodev: This part here, can you enable it back? We don't want debug logs on prod builds

core/src/lib.rs
```diff
@@ -254,12 +256,15 @@ impl Node {
...
-		tracing_subscriber::registry()
+		let registry = registry();
+
+		let registry = registry
 			.with(
 				tracing_subscriber::fmt::layer()
 					.with_file(true)
 					.with_line_number(true)
 					.with_ansi(false)
+					.with_target(false)
```
fogodev, line 265: I think we want to show which target output each log

fogodev (changes requested):

fogodev (approved):

PR #2574, opened 2024-06-27
[ENG-1809] Make core more resilient to crashes

fogodev:
Catching panics when running tasks and jobs. Also introducing a `catch_unwind` when generating thumbnails as a quick workaround until https://github.com/image-rs/image/issues/2230 is fixed upstream.

### DIFF
--- a/core/crates/heavy-lifting/src/media_processor/helpers/thumbnailer.rs
+++ b/core/crates/heavy-lifting/src/media_processor/helpers/thumbnailer.rs
@@ -14,6 +14,7 @@ use sd_file_ext::extensions::{VideoExtension, ALL_VIDEO_EXTENSIONS};
 
 use std::{
 	ops::Deref,
+	panic,
 	path::{Path, PathBuf},
 	str::FromStr,
 	time::Duration,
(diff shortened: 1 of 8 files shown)


core/crates/heavy-lifting/src/media_processor/helpers/thumbnailer.rs
```diff
@@ -378,7 +382,19 @@ async fn generate_image_thumbnail(
 
 		move || {
 			// Handling error on receiver side
-			let _ = tx.send(inner_generate_image_thumbnail(file_path));
+
+			let _ = tx.send(
+				panic::catch_unwind(|| inner_generate_image_thumbnail(&file_path)).unwrap_or_else(
```
HeavenVolkoff, line 387: This doesn't seem to be working. I tested with the image in https://github.com/image-rs/image/issues/2230 and the app still crashes. Should we revert to image-rs 0.24?

image

HeavenVolkoff (changes requested):

core/crates/heavy-lifting/src/media_processor/helpers/thumbnailer.rs
fogodev: Damn boy, but sure, let's revert then

HeavenVolkoff: Now that we reverted image, should we keep this catch unwind?

HeavenVolkoff (changes requested):

core/crates/heavy-lifting/src/media_processor/helpers/thumbnailer.rs
fogodev: It didn't catch the bad assert, but hopefully it catches other panics, so I see no harm leaving it there, we don't want crashed nonetheless

HeavenVolkoff (approved):

PR #2577, opened 2024-07-01
Quick sizing handling fix

ameer2468:
Better than previous approach^

### DIFF
--- a/interface/app/$libraryId/overview/StatCard.tsx
+++ b/interface/app/$libraryId/overview/StatCard.tsx
@@ -1,6 +1,6 @@
-import { useEffect, useMemo, useRef, useState } from 'react';
+import { useEffect, useMemo, useState } from 'react';
 import { humanizeSize } from '@sd/client';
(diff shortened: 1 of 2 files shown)


matheus-consoli (dismissed):
looking good!

fogodev (approved):

interface/app/$libraryId/overview/StatCard.tsx
```diff
@@ -20,17 +20,16 @@ const StatCard = ({ icon, name, connectionType, ...stats }: StatCardProps) => {
 
 	const isDark = useIsDark();
 
-	// Returns the value in a simpler form and unit i.e 2.5 TB
-	const totalSpaceSingle = humanizeSize(stats.totalSpace);
+	//TODO: Improve this
```
matheus-consoli, line 23: in the future, can we add a name when adding a todo? :slightly_smiling_face: 

like so:
```
// TODO(matheus-consoli): Improve this
```

that way is easy to track the person responsible for the note

matheus-consoli (approved):

PR #2597, opened 2024-07-10
Become Rust

jamiepine:
This PR removes the .gitignore for the prisma crate, resulting in the DB codegen being commited. This will change us from being 51% TypeScript and make us officially a Rust repo (which we are at heart), the 25th most popular one too :D


jamiepine:
actually sike this is a terrible idea, imaging having to commit codegen changes every time, yuck! write more rust team!!!

PR #2598, opened 2024-07-10
[ENG-1820] Better sync settings

jamiepine:


Rocky43007:
It is now being taken over by #2633 due to the multitude of API changes in the core and backend.

PR #2627, opened 2024-07-22
Switch the log message to a prettier and easier to read format

matheus-consoli:
This changes the log message to an easier-to-read format on the terminal (logging to files was not changed)

Before:

image

After:

image


iLynxcat (approved):
❤️ this is great, even for those of us who don't hack on the Rust code as much

PR #2630, opened 2024-07-22
Change `openssl-dev` dependency to `openssh-devel`

### LINKED ISSUE #2629: OpenSSH dependency install fails on `dnf`-based Linux distros
### Check for other issues

### Issue description

When I go to use setup.sh, the script fails due to `openssl-dev` not being available in current repos, I'm using Fedora 40.

### Steps to reproduce

1. Using any linux distro using dnf, try to use the install script.
2. Y to all prompts that require, it will then fail on trying to install openssl-dev

### Platform and versions

- OS: Fedora 40
- Spacedrive: N/A

9.5.0
cargo 1.79.0 (ffa9cf99a 2024-06-03)
rustc 1.79.0 (129f3b996 2024-06-10)

</pre></details>

### Stack trace

_No response_

### Additional context

_No response_

### Code of Conduct

greendoescode:
Changes openssl-dev to openssl-devel, which doesn't currently exist on default repos.
Line 188

Closes #2629


iLynxcat:
gonna leave this one mostly to the Linux boys, but will other package managers be able to fetch `openssh-devel` just fine, or do we need a conditional check?

greendoescode:
> gonna leave this one mostly to the Linux boys, but will other package managers be able to fetch `openssh-devel` just fine, or do we need a conditional check?

I have a feeling older versions have openssl-dev and not openssl-devel, but I havent actually checked. It might possibily require a condition though

HeavenVolkoff:
From what I could find `openssl-devel` is the correct package name for fedora 39 and 40, and fedora <= 38 is EOL according to:
https://docs.fedoraproject.org/en-US/releases/eol/#_unsupported_fedora_linux_releases

Considering no one on our team uses Fedora Linux for development I think we can just merge this as is and if someone complains in the future we can add the conditional check if necessary

HeavenVolkoff (approved):

HeavenVolkoff:
thanks for the contribution @greendoescode

PR #2631, opened 2024-07-22
Remove explicit ref in checkout for type and style check CI job

iLynxcat:
This is currently breaking checkouts for PRs from forks


myung03 (approved):
lgtm!

PR #2654, opened 2024-08-08
Migrate to Tauri v2 Release Candidate

Rocky43007:
This PR migrates Spacedrive to use the new Tauri V2 Release Candidate!

There aren't many changes due to the fact that we were using the v2 beta already, and my own testing shows that everything works and the app builds with no issues.

### DIFF
--- a/apps/desktop/src-tauri/Cargo.toml
+++ b/apps/desktop/src-tauri/Cargo.toml
@@ -35,7 +35,7 @@ uuid = { workspace = true, features = ["serde"] }
 # Specific Desktop dependencies
 # WARNING: Do NOT enable default features, as that vendors dbus (see below)
 opener = { version = "0.7.1", features = ["reveal"], default-features = false }
-tauri = { version = "=2.0.0-beta.17", features = [
+tauri = { version = "2.0.0-beta", features = [
 	"macos-private-api",
 	"unstable",
 	"linux-libxdo",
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -70,36 +71,36 @@ webp = "0.2.6"                                        # Update blocked by image
 
 [workspace.dependencies.prisma-client-rust]
 git = "https://github.com/brendonovich/prisma-client-rust"
-rev = "4f9ef9d38ca732162accff72b2eb684d2f120bab"
+rev = "4f9ef9d"
```
pauliesnug, line 74: using short refs are a bit insecure, but it should be fine

HeavenVolkoff: How insecure are we talking? Do you have any links to any known vulnerabilities on using short SHA hashes with git dependencies on cargo? I don't mind reverting this, or increasing the number of characters, if necessary. Was just trying to tidy up the file and make it a bit easier to read

HeavenVolkoff: From a quick search, I found this:
https://stackoverflow.com/a/21015031

Maybe increasing to 10 chars would be better

pauliesnug: @HeavenVolkoff yeah, 10 chars is a safe bet. just mostly worried about sha collisions, intentional or unintentional. especially in larger repos. for PCR it's not a huge risk, but in the other ones modified its possible for something to go wrong. there are tools that can make a git commit start with a certain few letters, which can be dangerous

fogodev (approved):

PR #2633, opened 2024-07-23
[ENG-1828] Migrate to new Cloud Services API

Rocky43007:
We have been working on moving our cloud services to a new infrastructure and codebase due to many issues in development and testing for release. This PR amalgamates all core and frontend changes in the app to work with our new cloud services system and fixes problems we may have faced during development.

PR #2660, opened 2024-08-15
Crypto revamp

fogodev:
Revamping crypto crate, removing a ton of half-baked or overcomplicated code. This code is unused for now, it is prep-work for #2633


HeavenVolkoff (approved):
LGTM

PR #2665, opened 2024-08-16
Update to Expo SDK 51

Rocky43007:
This PR updates the mobile dependencies to Expo SDK 51, giving us access to the new React Native architecture and prepping for Expo's SDK 52. This allows us to move entirely to the new React Native arch when Expo SDK 52 releases. Plus, there is more stability! 

Due to the update to the latest React Native version, the iOS CI runner has been updated to use `macos-14` for Xcode 14.3 and now tests against iOS 17.4.


HeavenVolkoff (approved):
LGTM

PR #2666, opened 2024-08-16
Fix pdf thumbnails

HeavenVolkoff:
Use a fork of pdfium-renderer including the required fix while a new version isn't released

Related:
https://github.com/ajrcarey/pdfium-render/pull/155


Rocky43007 (approved):
Tested and works.

PR #2673, opened 2024-08-18
Added translation keys for FileKindStats, general (settings) and StarfieldEffect components

artsiom-voitas:
- Added translation keys for `FileKindStats`, `general` (settings) and `StarfieldEffect` components. 
- Translated those keys on Belarusian and Russian.
- Shortened string translation on Belarusian and Russian languages, which deforms the appearance of the Logout button
- Added counters into `getWording() ` translation variables, so we can use plural translation with the right words for specific languages
- Removed unnecessary translation keys in all languages (for example, `file_two file_many` which did nothing for almost every language, except Arabic)


interface/locales/ar/common.json
```diff
@@ -88,12 +86,14 @@
   "config": "Config",
   "configure_location": "تكوين الموقع",
   "confirm": "Confirm",
+  "Connect": "يتصل",
```
ameer2468, line 89: As a native Arabic speaker, this should be اتصل

interface/locales/ar/common.json
```diff
@@ -88,12 +86,14 @@
...
   "connect_cloud": "Connect a cloud",
   "connect_cloud_description": "Connect your cloud accounts to Spacedrive.",
   "connect_device": "Connect a device",
   "connect_device_description": "Spacedrive works best on all your devices.",
   "connect_library_to_cloud": "قم بتوصيل المكتبة بـ Spacedrive Cloud",
   "connected": "متصل",
+  "Connecting": "توصيل",
```
ameer2468, line 96: This should be: الاتصال

ameer2468 (approved):

Issue #1560, opened 2023-10-13
"Open With" list order is random


dylanbr0wn:
### Describe the bug

Opening context menu of a file and opening the Open With submenu produces an unordered list of results that changes each time it is opened. There is also a significant delay from when list is open to when it is populated or updated.

https://github.com/spacedriveapp/spacedrive/assets/40218657/8eda3954-98d9-4201-8913-0226c77b9abd

### Reproduction

1. Open Spacedrive and navigate to a file
2. Open Context menu of the file
3. Hover over the Open With option

### Expected behavior

I would expect this list to be ordered alphabetically, perhaps with my default application for this specific file type as the first option. Would also expect this list to be populated when dropdown is opened and not after. 

### Platform and versions

```Shell
pnpm --version && cargo --version && rustc --version:
8.9.0
cargo 1.73.0 (9c4383fb5 2023-08-26)
rustc 1.73.0 (cc66ad468 2023-10-03)

OS:
ProductName:		macOS
ProductVersion:		14.0
BuildVersion:		23A344
```

killua1zoldyck:
If no one is working on this, can I work on this?

HeavenVolkoff:
Sure, go ahead @killua1zoldyck, I'll assign this issue to you. I agree with what you wrote in the `expected behavior`, only point I would add is that if there is an API that returns a list of apps already sorted by preference/most used to prefer that over alphabetic order (If I remember correctly this worked correctly at some point in time, some bug must have cropped up somewhat recently)

Below is where most of the logic for this feature is located. Native, backend and front-end:
 
Linux:
https://github.com/spacedriveapp/spacedrive/blob/main/apps/desktop/crates/linux/src/app_info.rs#L75-L122

macOS:
https://github.com/spacedriveapp/spacedrive/blob/main/apps/desktop/crates/macos/src-swift/files.swift#L30-L105

Windows:
https://github.com/spacedriveapp/spacedrive/blob/main/apps/desktop/crates/windows/src/lib.rs#L55-L94

Backend:
https://github.com/spacedriveapp/spacedrive/blob/main/apps/desktop/src-tauri/src/file.rs#L110-L256

Front-end:
https://github.com/spacedriveapp/spacedrive/blob/main/interface/app/%24libraryId/Explorer/ContextMenu/OpenWith.tsx

MadisonKonig:
Is it fine to just have this be fixed on the front-end? It's just a single line of code to sort the list that is shown.

[...]

PR #2697, opened 2024-09-08
Fix "open with" list ordering

MadisonKonig:
#1560

Closes "Open With" list order is random

Fixed the bug both on the front end, and on the back end. 

For the front, it's just a simple sorting of the values by name. 

For the back, the problem arose as it's initially put into a HashSet, which doesn't keep and order. Thankfully it is turned into a Vec before propagating as a result.  I would have liked to keep it all nested in one tidy chain of function calls, but sort doesn't return anything.
<https://users.rust-lang.org/t/chaining-functions-yields-error-method-not-found-in/93252/4>

I think @iLynxcat mentioned that it's better to do it on the backend, but the front end works just fine too since it'll be a small list. But I'll give both as solutions, and let you decide which to keep :)


iLynxcat:
let's keep the JavaScript sorting now, seems that locale-sensitive compare doesn't work as we'd expect in Rust. (e.g. apps would be ordered Ferris, FeWheel, Férris [mind my terrible example], instead of Ferris, Férris, FeWheel)

MadisonKonig:
I'm a bit confused, as the JavaScript is using `localeCompare()`, but you're wanting to keep it even though it's not meeting expectations? Did you mean to type Rust?

iLynxcat:
> I'm a bit confused, as the JavaScript is using `localeCompare()`, but you're wanting to keep it even though it's not meeting expectations? Did you mean to type Rust?

didn't type super clearly there; proper Unicode-respecting comparison/sorting does not work as expected in Rust, and does in JS as I understand it

iLynxcat (approved):
Thanks for contributing this, it's been a petty annoyance for some users who'll be relieved it's fixed!

PR #2715, opened 2024-09-20
Add SQLite pragmas to optimize performace

jamiepine:
This PR optimizes SQLite performance by configuring key PRAGMA settings, including:

	•	Enabling Write-Ahead Logging (journal_mode = WAL) for improved concurrency.
	•	Setting synchronous = NORMAL to balance durability and performance.
	•	Increasing memory-mapped I/O size to 512MB for faster database access.
	•	Limiting the journal file size to 64MB to prevent excessive disk usage.
	•	Setting the cache size to 10,000 pages (~40MB) for faster reads.

These changes aim to enhance database performance, especially in environments with concurrent transactions.

### DIFF
--- a/.cspell/project_words.txt
+++ b/.cspell/project_words.txt
@@ -34,6 +34,7 @@ lütke
 marietti
 mehrzad
 Mjpeg
+Mmap
 mytton
 narkhede
 naveen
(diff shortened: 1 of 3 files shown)


mercxry (approved):

iLynxcat (approved):
nice! performance boost is always great

PR #2728, opened 2024-09-25
Close file handle after getting its index (windows)

### LINKED ISSUE #2725: Handle leak in Windows
### Check for other issues

### Issue description

When Spacedrive index new locations, its handle count increases and does not decrease again. This causes Windows to slow down.

### Steps to reproduce

1. Open Task Manager, choose to show Handles column, observe sd-desktop.exe
2. Click on Add Location, select location with lots of files
3. Observe Handles count increase

### Platform and versions

- OS: Windows 11
- Spacedrive: v0.4.2 - 0af8dbf

</pre></details>

### Stack trace

_No response_

### Additional context

High handle count in Windows will make the OS slow. It does not show up on any CPU time measurement, but the whole machine feels sluggish. When I noticed it was slow, I found Spacedrive having nearly 500.00 handles open. Killing the process fixed the issue (after a minute or so).

### Code of Conduct

matheus-consoli:
We're using the raw Windows API to read the file index ("inode"), but forgetting to close the handle after reading the metadata causing a leak of file handles.

This PR fixes the issue by manually closing the handle.

It will be better to use `MetadataExt::file_index` in the future once it gets stabilized.

Closes #2725

### DIFF
--- a/core/crates/file-path-helper/src/lib.rs
+++ b/core/crates/file-path-helper/src/lib.rs
@@ -103,11 +103,13 @@ pub fn path_is_hidden(path: impl AsRef<Path>, metadata: &Metadata) -> bool {
 }
 
 #[cfg(target_family = "windows")]
+// TODO(matheus-consoli): rewrite this function using a safe API once `MetadataExt::file_index` got stabilized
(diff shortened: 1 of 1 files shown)


matheus-consoli:
before this pr:
image


after it:
image

indexing the same location

fogodev (approved):

PR #2735, opened 2024-09-26
Better font rendering & light theme improvements

jamiepine:
Shifted font weights to overall lighter weight, I feel like the difference on macOS at least is amazing, bolder fonts for UI can look off, compared to Finder, here you can see a before and after in light theme.

This PR will include tweaked light versions of all icons, for now its just the document and it already looks better.
Screenshot 2024-09-26 at 2 05 13 PM


jamiepine:
Closing as part of v2 rewrite transition

PR #2741, opened 2024-09-29
Organize and clean up internationalization file keys

iLynxcat:
Lots of files were ordered wrong, had uppercase/inconsistent key names, overall the whole thing was a mess. We should probably put something in place to ensure these don't get messed up like this again but probably not worth the time and effort at this exact moment

Having files formatted wrong makes editing translations via i18n Ally a pain since the extension reformats everything every time you make an edit within the extension

Closes ENG-1922


Rocky43007 (approved):
LGTM

PR #2745, opened 2024-09-30
Handle thumbnail encoder error

### LINKED ISSUE #2736: webp dimension issue of some kind causing crash on start
### Check for other issues

### Issue description

When trying to start spacedrive after giving it some files to look at, it panics immediately using the webp crate ... not really sure what's going on beyond that.

### Steps to reproduce

_No response_

### Platform and versions

- OS: Gentoo
- Spacedrive: latest alpha ... 0.4.2 or something?

### Stack trace

```shell
2024-09-27T00:05:52.703833Z ERROR sd_core: panicked at /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/webp-0.2.6/src/encoder.rs:70:44:
called `Result::unwrap()` on an `Err` value: VP8_ENC_ERROR_BAD_DIMENSION, panic.file: "/home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/webp-0.2.6/src/encoder.rs:70", panic.column: 44
at core/src/lib.rs:290
```

### Additional context

_No response_

### Code of Conduct

matheus-consoli:
When generating thumbnails, we may crash if the file is invalid, e.g. has an invalid width or height.

The application shouldn't crash, but report the invalid file as a non-critical error and keep processing the remaining files.

This PR does that by not calling the [`Encoder::encode`](https://github.com/jaredforth/webp/blob/4e7616a0a62bc4a78a9c63cfd3040fdd6757f92b/src/encoder.rs#L70), which `.unwrap`s, but rather calling `Encoder::encode_simple` and handling its result.

Closes #2736

### DIFF
--- a/core/crates/heavy-lifting/src/media_processor/helpers/thumbnailer.rs
+++ b/core/crates/heavy-lifting/src/media_processor/helpers/thumbnailer.rs
@@ -363,7 +363,7 @@ fn inner_generate_image_thumbnail(
 		)
 	})?;
 
-	let thumb = encoder.encode_advanced(&*WEBP_CONFIG).map_err(|reason| {
+	let thumb = encoder.encode_advanced(&WEBP_CONFIG).map_err(|reason| {
 		thumbnailer::NonCriticalThumbnailerError::WebPEncoding(
 			file_path.clone(),
 			format!("{reason:?}"),
--- a/core/crates/heavy-lifting/src/media_processor/helpers/thumbnailer.rs
+++ b/core/crates/heavy-lifting/src/media_processor/helpers/thumbnailer.rs
@@ -32,7 +32,7 @@ use tokio::{
 };
 use tracing::{error, instrument, trace};
 use uuid::Uuid;
-use webp::Encoder;
+use webp::{Encoder, WebPConfig};
 
 // Files names constants
 pub const THUMBNAIL_CACHE_DIR_NAME: &str = "thumbnails";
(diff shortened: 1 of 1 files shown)


core/crates/heavy-lifting/src/media_processor/helpers/thumbnailer.rs
```diff
@@ -354,10 +354,19 @@ fn inner_generate_image_thumbnail(
...
 
+	let thumb = encoder
+		.encode_simple(false, TARGET_QUALITY)
+		.map_err(|reason| {
+			thumbnailer::NonCriticalThumbnailerError::WebPEncoding(
+				file_path.clone(),
+				format!("{reason:?}"),
+			)
+		})?;
+
```
HeavenVolkoff, line 365: Looking into `webp` crate source, the `encode_simple` method also has a unwrap inside it:

https://github.com/jaredforth/webp/blob/main/src/encoder.rs#L78-L88

Maybe we should user `encode_advanced` instead, and have a single `WebPConfig` instance that is created during the thumbnailer init

HeavenVolkoff (changes requested):

core/crates/heavy-lifting/src/media_processor/helpers/thumbnailer.rs
matheus-consoli: The config now is a static variable, and every call to the thumbnailer uses it.
I `.expect`ed on its creation, which is basically the same thing as `unwrap` + context, but as it only occurs once in the application's lifetime, I think it's ok.

HeavenVolkoff (approved):
LGTM

PR #2746, opened 2024-09-30
Remove dependency on `once_cell` and use `std` types

matheus-consoli:
`once_cell` was broadly used by the Rust community, but nowadays the Rust's standard library offers the same functionality. Removing a dependency is always good.


fogodev (dismissed):

matheus-consoli:
removed from the queue because of a conflict with https://github.com/spacedriveapp/spacedrive/pull/2745

fogodev (approved):

PR #2748, opened 2024-10-01
Fix Inspector notes

jamiepine:
Notes in the Inspector work now... not that anyone noticed they didn't 🥲
- Ensure note value updates when prop data changes.
- Fixed a rendering bug in MediaData caused by separate metadata functions leading to inconsistent updates. Replaced them with a single renderMetadata function for better consistency.
- Added a Valtio store in Note.tsx to cache note values across files, prioritizing cached values unless the backend value changes.


ameer2468:
Tested and works great! 

lgtm

ameer2468 (approved):

PR #2798, opened 2024-11-05
Add `node-gyp` dependency to `scripts`

iLynxcat:
This PR adds `node-gyp` as a dependency in `scripts`, fixing an issue where if `node-gyp` isn't globally installed the install command for `registry-js` (a dependency of `os-proxy-config`) would fail, forcing other install and postinstall scripts to be forcefully aborted with it.


HeavenVolkoff (approved):
LGTM

PR #2813, opened 2024-11-16
Fix issue with Node 23

fogodev:
I was having issues trying to launch spacedrive in dev mode due to updating my Node to version 23. So I tried to fix it, changing the desktop app to use ES Modules instead of CommonJS. 

I'm not too sure as TS isn't much my thing, so this PR needs a proper reviewing from some TS guys.

### DIFF
--- a/apps/desktop/.eslintrc.mjs
+++ b/apps/desktop/.eslintrc.mjs
@@ -1,4 +1,4 @@
-module.exports = {
+export default {
 	extends: [require.resolve('@sd/config/eslint/web.js')],
 	parserOptions: {
 		tsconfigRootDir: __dirname,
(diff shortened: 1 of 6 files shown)


iLynxcat (commented):
Tailwind supports `.config.ts` now, we should test to see if it works in our environment if we’re already renaming it

iLynxcat:
We have a `nvmrc`, though — we should update this to 23 if it’s supported everywhere or you should use a node version manager (like `fnm`, which I use) to automatically switch to the correct Node version https://github.com/spacedriveapp/spacedrive/blob/main/.nvmrc

jamiepine:
Closing as part of v2 rewrite transition

PR #2839, opened 2024-12-21
[ENG-1402] Drag & drop into OS

Rocky43007:
This PR adds functionality to drag and drop files from Spacedrive into the OS


Rocky43007:
It freezes the Drag Overlay UI, but otherwise, the functionality exists and works, even with multiple files.

PR #2840, opened 2024-12-22
Refine i18n zh-CN translations for better accuracy

lzt1008:
This PR aims to refine the Chinese (zh-CN) translations to get better accuracy and readability.

Some of the existing Chinese translations use expressions or terms that are not commonly used and not appropriate in the context.

| Original Text   | Previous Translation | Updated Translation | Description                                                                 |
|------------------|----------------------|----------------------|----------------------------------------------------------------------------------|
| "Apply"         | "申请"               | "应用"               | "申请" (apply) is typically used for requesting qualifications or permissions, which is inappropriate in this context. "应用" (apply) is more accurate for functionality like applying settings. |
| "Extension"     | "扩大"               | "扩展名"             | "扩大" (expand) means to increase scope, which does not align with the technical term "Extension." "扩展名" (file extension) is the correct term in this context. |
| "Hidden"        | "隐"                 | "已隐藏"             | "隐" (hidden) is incomplete and rarely used, which might confuse users. "已隐藏" (hidden) clearly indicates the state and aligns with natural Chinese usage. |
| "Location"      | "地点"               | "位置"               | "地点" (place) typically refers to a specific geographical location. "位置" (position) is a more general term that fits technical and interface-related contexts better. |


Rocky43007 (approved):

PR #2843, opened 2025-01-02
Optimize Option<Flash>::from(FlashValue) using bitwise AND

julianbraha:
Resolves this TODO in `crates/media-metadata/src/exif/flash/data.rs`:
`// TODO(brxken128): This can be heavily optimized with bitwise AND`


Rocky43007 (approved):
Looks good! Sorry for taking forever to merge this.

PR #2895, opened 2025-12-04
fix(windows): Resolve build errors and runtime scripts for Windows development

Yukeey07:
This PR addresses multiple compilation errors and runtime issues preventing the project from building and running correctly on Windows environments.

**Changes:**

- **Core (Volume/FS)**: Fixed unresolved imports for FilesystemHandler in refs.rs and ntfs.rs.

- **Core (NTFS)**: Resolved "use of moved value" errors in resolve_ntfs_path by properly cloning paths before spawn_blocking closures.

- **Core (Entry & Local Backend)**: Removed usage of the unstable file_index API (feature windows_by_handle) in favor of returning None for inodes on Windows to fix stable build errors.

- **Core (Windows Platform)**: Fixed a syntax error where total_bytes_capacity was accessed as a field instead of a method.

- **CLI**: Guarded Unix-specific permission logic (PermissionsExt::set_mode) with #[cfg(unix)] to prevent compilation failures on Windows.

- **Scripts**: Updated apps/tauri/scripts/dev-with-daemon.ts to correctly handle the .exe extension and Windows file paths when spawning the daemon.

**⚠️ Important Note**
After building, the necessary DLLs (FFmpeg, OpenSSL, etc.) are not automatically placed in the debug folder. To run the app, please execute the following PowerShell commands in the project root to copy the dependencies:

```
Copy-Item -Path "apps\.deps\lib\*.dll" -Destination "target\debug" -ErrorAction SilentlyContinue
Copy-Item -Path "apps\.deps\bin\*.dll" -Destination "target\debug" -ErrorAction SilentlyContinue
```

### DIFF
--- a/apps/cli/src/domains/update/mod.rs
+++ b/apps/cli/src/domains/update/mod.rs
@@ -169,7 +169,6 @@ async fn download_file(url: &str, expected_size: u64) -> Result<Vec<u8>> {
 
 fn replace_binary(path: &PathBuf, data: &[u8]) -> Result<()> {
 	use std::fs;
-	use std::os::unix::fs::PermissionsExt;
 
 	// Create backup
 	let backup_path = path.with_extension("bak");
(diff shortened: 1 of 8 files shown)


jamiepine:
@cursor review

PR #2944, opened 2025-12-28
fix(security): prevent path traversal in P2P file transfers and deletions

### LINKED ISSUE #2943: Security concern: P2P protocols can access files outside Libraries
### Check for other issues

- [x] I've checked for other issues, and none exist.

### Issue description

Hey!
While testing, I was doing a quick audit of the codebase and found something I think is worth discussing regarding the P2P protocol layer. 
From what I can see, the current implementation treats paired devices as fully trusted for *all* file operations on the whole file system of paired nodes, not just operations within their Library Locations.

### The problem (as I understand it)

When users add a Location in Spacedrive, they're drawing an explicit boundary: "this folder is part of my Spacedrive-managed space."
My interpretation is that this represents the trust contract. A user might trust Spacedrive to sync their photos folder, but that doesn't necessarily mean consent to access SSH keys, browser profiles, or system files.

Right now, once a device is paired, there's nothing stopping it from requesting operations outside those boundaries **on all peered nodes**:
- **Reads** of any file (> data exfiltration)
- **Writes** to arbitrary paths (> drop malware into startup folders, overwrite configs)
- **Deletes** of critical files (> ransomware-style destruction)

This turns a single compromised node into a potential pivot point for achieving Remote Code Execution on all connected Spacedrive peers (as an attacker controlling one paired device could easily execute code on every other node by dropping malicious scripts in startup directories or modifying config files).

In my opinion, the Location model exists to give users control, but this bypass seems to undermine that design principle.

### Technical details

The issue is in [file_transfer.rs](https://github.com/spacedriveapp/spacedrive/blob/544f4ef0636165801de779d383793087f2c35652/core/src/service/network/protocol/file_transfer.rs) and [file_delete.rs](https://github.com/spacedriveapp/spacedrive/blob/544f4ef0636165801de779d383793087f2c35652/core/src/service/network/protocol/file_delete.rs)
Paths from remote devices are converted to local `PathBuf` and used directly without validation:

```rust
// file_delete.rs (current code)
for path in paths {
    let local_path = path
        .as_local_path()
        .ok_or_else(|| anyhow::anyhow!("Path is not local"))?;

    // Get file size before deletion
    let size = tokio::fs::metadata(local_path).await...
    
    // Perform deletion based on mode (no boundary check!)
    let result = match mode {
        DeleteMode::Permanent => LocalDeleteStrategy.permanent_delete(local_path).await,
        // ...
    };
}
```

The same pattern exists in file transfers.

While investigating, I also came across a comment in [validate_path_access()](https://github.com/spacedriveapp/spacedrive/blob/544f4ef0636165801de779d383793087f2c35652/core/src/service/network/protocol/file_transfer.rs#L1020) that mentions adding "library-level access control" later, so it looks like this was already on the radar to some extent.
The fix I'm proposing implements that missing validation for both the file transfer and file deletion protocols.

### Proposed fix
The PR I'm submitting adds path validation to both handlers:

- Paths are canonicalized (blocking `../` tricks) and checked against registered Locations
- Requests targeting paths outside Locations are rejected and logged
- Regression tests included to prevent reintroduction

Let me know what you think!

### Steps to reproduce

_No response_

### Platform and versions

- OS: All
- Spacedrive: v2.0.0-alpha.1

`bun --version && cargo --version && rustc --version`
1.3.5
cargo 1.92.0 (344c4567c 2025-10-21)
rustc 1.92.0 (ded5c06cf 2025-12-08)

</pre></details>

### Stack trace

```shell

```

### Additional context

_No response_

### Code of Conduct

- [x] I agree to follow the Code of Conduct.

slvnlrt:
Add path validation to FileTransferProtocolHandler and FileDeleteProtocolHandler to prevent unauthorized file access during P2P operations.

- Add is_path_allowed() validation against registered Locations
- Add dynamic location lookup via CoreContext integration
- Add security regression tests for both protocols

Prevents malicious paired devices from reading/writing/deleting files outside registered library locations.

Closes #2943

### DIFF
--- a/core/src/service/network/protocol/file_transfer.rs
+++ b/core/src/service/network/protocol/file_transfer.rs
@@ -365,30 +365,29 @@ impl FileTransferProtocolHandler {
 	}
 
 	/// Get all allowed paths by combining static allowed_paths with dynamic locations.
-	/// This queries all libraries for their registered locations.
-	fn get_all_allowed_paths(&self) -> Vec<PathBuf> {
+	/// This queries all libraries for their registered locations asynchronously.
+	async fn get_all_allowed_paths(&self) -> Vec<PathBuf> {
 		let mut paths = Vec::new();
 
-		// Add statically configured allowed paths
-		{
+		// Add statically configured allowed paths (clone to avoid holding lock across await)
+		let static_paths = {
 			let allowed = self.allowed_paths.read().unwrap();
-			paths.extend(allowed.clone());
-		}
+			allowed.clone()
+		};
+
+		paths.extend(static_paths);
 
 		// Add dynamic location paths from all libraries via CoreContext
 		if let Some(ctx) = &self.core_context {
-			let library_manager_guard = ctx.library_manager.blocking_read();
+			let library_manager_guard = ctx.library_manager.read().await;
 			if let Some(library_manager) = library_manager_guard.as_ref() {
-				// Get all active libraries using tokio's block_on (safe in async context)
-				let library_list: Vec<std::sync::Arc<crate::library::Library>> =
-					tokio::runtime::Handle::current().block_on(library_manager.list());
+				// Get all active libraries
+				let library_list = library_manager.list().await;
 				for library in library_list {
 					// Get locations for this library using LocationManager
 					let location_manager =
 						crate::location::LocationManager::new((*ctx.events).clone());
-					if let Ok(locations) = tokio::runtime::Handle::current()
-						.block_on(location_manager.list_locations(&library))
-					{
+					if let Ok(locations) = location_manager.list_locations(&library).await {
 						for loc in locations {
 							paths.push(loc.path.clone());
 						}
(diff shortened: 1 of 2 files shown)


jamiepine:
Amazing PR thank you, issue was very descriptive also and I agree completely in that locations should be a hard boundary for multi-device access.

jamiepine:
@cursor review

slvnlrt:
> Amazing PR thank you, issue was very descriptive also and I agree completely in that locations should be a hard boundary for multi-device access.

Thanks! Glad it was useful :)

Cursor is a sniper lol, it caught a possible blocking call issue + a second unvalidated entry point. Both should be fixed in the latest commit!

jamiepine:
After running `cargo test -p sd-core --test file_transfer_test`, I saw this fix in action preventing the test from passing due to `/temp/received_files` not being in `allowed_paths`

While trying to fix the test I also discovered the async/await issues weren't fully resolved. The issue was that even though `blocking_read()` and `block_on()` were removed, the `RwLockReadGuard` was still being held across `.await` points in `get_all_allowed_paths()`, which makes the future not `Send`.

My commit fixes this by:
1. Updating the test to register the allowed path directly
2. Cloning the data and dropping the lock before any `.await` calls
3. Making all the path validation methods properly async throughout the call chain
4. Using `.read().await` instead of `blocking_read()` on the library manager

The `file_transfer_test` now passes successfully!

```
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 2 filtered out; finished in 63.10s
```

PR #2971, opened 2026-01-13
Enhanced Size View with Per-Tab Persistence and Dynamic Controls

jamiepine:
Major improvements to the Size View functionality with per-tab state persistence, dynamic zoom/pan controls, and enhanced visualization options.

https://github.com/user-attachments/assets/83846cbb-677d-4ce3-8134-53d7bf83cf70

### Per-Tab State Persistence
- Zoom level and pan position now persist when switching between tabs
- Each tab maintains its own independent size view transform state
- Smooth transitions between tabs without losing your position

### Smart Tab Navigation Detection
- Distinguishes between tab switches and actual folder navigation
- Only resets zoom when navigating into folders (double-clicking)
- Preserves zoom when switching tabs

### Enhanced Folder Visibility
- Folders are always included in size view, regardless of size
- Folders get a minimum bubble size (2.5% of average file size) to remain visible
- All folders are shown in addition to the largest files

### Dynamic Item Limit Control
- New slider in View Settings to control number of items shown (50 to total file count)
- Shows current selection vs total available: "100 / 863"
- Adapts max value based on current directory contents

### Improved Layout & Controls
- Floating zoom controls now account for inspector width
- Smooth transitions when toggling inspector
- Top gradient mask properly layered above size view but below preview/sidebar
- Selection state properly applies on mount

### Full Transform Persistence
- Stores complete D3 transform (zoom scale + x/y pan position)
- Accurately restores both zoom level and pan offset when returning to a tab
- No more "jumping" back to center when switching tabs

### State Management
- Added `sizeViewTransform` to `TabExplorerState` (replaces simple `sizeViewZoom`)
- Updated `ViewSettings` interface with `sizeViewItemLimit` property
- View settings for size view now properly managed through UI state

### Component Updates
- `SizeView.tsx`: Complete refactor of zoom/pan persistence logic
- `TabManagerContext.tsx`: Enhanced to store full transform object
- `ExplorerProvider`: Updated to handle transform state
- `ViewSettings.tsx`: New dynamic slider for item limit control
- `ShellLayout.tsx`: Improved z-index layering for gradients

### Bug Fixes
- Fixed selection not applying on initial mount
- Fixed view settings not updating for `showFileSize` and `sizeViewItemLimit`
- Fixed zoom controls hidden behind inspector
- Fixed stale data showing during tab switches

## Test Plan
- [x] Switch between tabs and verify zoom/pan persists
- [x] Navigate into folders and verify zoom resets
- [x] Toggle inspector and verify controls remain visible
- [x] Adjust item limit slider and verify bubble count changes
- [x] Select files in size view, switch views, return to size view - selection persists
- [x] Switch between tabs with different zoom levels

## Screenshots
_Size view with dynamic controls and folder visibility_


jamiepine:
@tembo fix this:

> ### Connection watcher leaves stale entry blocking future watchers
> **Medium Severity**
> 
> When a watched connection closes but other connections to the same node exist, the node remains in `watched_nodes` while the watcher task completes. This leaves remaining connections without a watcher, and prevents new connections from spawning watchers since the node is still in the set. When those unwatched connections eventually close, instant offline detection won't occur. The node should be removed from `watched_nodes` when the watcher completes regardless of other connections, allowing subsequent connection closures to be monitored.
> 
> `core/src/service/network/core/mod.rs#L1854-L1861`

PR #3052, opened 2026-04-13
fix(windows): replace broken icon.ico placeholder, fix Obsidian SVG gradients

slvnlrt:
## Summary

Small housekeeping PR for two minor asset issues.

### icon.ico

The current icon.ico is a 103-byte placeholder that breaks the Windows Tauri build (error RC2175, invalid icon format). Replaced it with a proper multi-size .ico generated from the existing icon.png in the same icons directory. Nothing fancy, just enough to unblock cargo tauri build on Windows. Can be swapped for the final branding icon whenever that's ready.

### Obsidian adapter SVG

The gradientTransform attributes in adapters/obsidian/icon.svg used commas as decimal separators (0,7282 instead of 0.7282), likely from a locale-dependent design tool export. Browsers can't parse that per SVG spec. Replaced with dots.

## Files changed (2)

- apps/tauri/src-tauri/icons/icon.ico -- valid .ico replacing broken placeholder
- adapters/obsidian/icon.svg -- decimal separator fix in gradient transforms


jamiepine:
oh the branch was deleted after merge, I need this fix so I'll cherry pick. thank you!!

PR #3061, opened 2026-04-18
fix library sync backfill O(N^2) hotspots on both sides of the protocol

### LINKED ISSUE #3058: Library sync backfill slows progressively on large shared tables (O(N^2) sender work + per-record receiver event emission)
## Summary

Backfilling a large shared table (observed with ~327k `content_identity` records on initial pair) starts fast (~4-5k records/batch) then slows dramatically as progress advances (~100 records/batch past 100k applied). The slowdown is progressive and grows with how much has already been synced, not with batch size — a classic O(N) or O(N^2) in the backfill loop.

Root causes are a combination of issues on both sides of the sync protocol. `content_identity` is a `shared` model so it flows through `SharedChangeRequest` / `peer_log` (not the device-owned state path). Several code paths do more work per batch than they need to, and some scale with total-rows-so-far rather than batch size.

## Reproduction

1. Index a library on device A until `content_identity` has >100k rows.
2. Pair a new device B against the same library.
3. Enable library sync.
4. Observe the `content_identity` backfill in logs. Batches start around the configured `backfill_batch_size` (default 5000) and progressively shrink to ~100/batch past ~100k applied.

### 1. Sender fetches the entire peer log on every batch request (no SQL LIMIT)

`core/src/infra/sync/peer_log.rs:215-238` — `PeerLog::get_since(since_hlc)` runs:

```sql
SELECT hlc, model_type, record_uuid, change_type, data
FROM shared_changes WHERE hlc > ? ORDER BY hlc ASC
```

There is no `LIMIT`. The caller (`core/src/service/sync/protocol_handler.rs:118-121`) materializes every remaining row into `Vec<SharedChangeEntry>` — parsing JSON in `data` per row — and only then does `.take(limit)`:

```rust
let entries = self.peer_sync.peer_log.get_since(since_hlc).await?;
let has_more = entries.len() >= limit;
let limited: Vec<_> = entries.into_iter().take(limit).collect();
```

With ~327k entries, each batch request reloads and re-parses ~320k → 315k → 310k rows on the sender. Sender work per batch is O(total_remaining), so total sender work across the backfill is roughly O(N^2 / batch_size). This is the single biggest factor in the observed slowdown.

**Fix:** push `LIMIT ?` (pass `batch_size + 1` so `has_more` can still be derived) into the SQL in both branches of `get_since`.

### 2. First-request snapshot is capped at 100k per model type

`core/src/service/sync/protocol_handler.rs:126-150` — when `since_hlc` is `None`, the sender also builds a `current_state` snapshot via `query_all_shared_models(None, 100_000, db)`. This cap means ~227k of the 327k `content_identity` records cannot come through the snapshot and must be fetched via peer log pagination — which then hits root cause #1.

**Fix:** either page the snapshot, or use only the peer log path for models that exceed the cap.

### 3. Receiver emits resource events per-record with 2 DB queries each

`core/src/service/sync/backfill.rs:737` invokes `ResourceManager::emit_batch_resource_events` after each applied batch. That call delegates to `emit_resource_events` (`core/src/domain/resource_manager.rs:267-275`) which loops per-UUID through `map_dependency_to_virtual_ids` (`core/src/domain/resource_manager.rs:174-181`):

```rust
for resource_id in &resource_ids {
    let virtual_mappings =
        map_dependency_to_virtual_ids(&self.db, resource_type, *resource_id).await?;
    ...
}
```

For the `content_identity` case, this routes into `core/src/domain/file.rs:151-169`:

```rust
\"content_identity\" => {
    let ci = content_identity::Entity::find()
        .filter(content_identity::Column::Uuid.eq(dependency_id))
        .one(db).await?.ok_or_else(...)?;

    let entries = entry::Entity::find()
        .filter(entry::Column::ContentId.eq(ci.id))
        .all(db).await?;
    ...
}
```

That is two DB round trips per applied record, purely to emit UI invalidation events. For 327k records that is ~654k extra queries during backfill. The composite partial index `idx_entries_content_volume` (added in `core/src/infra/db/migration/m20260414_000001_add_redundancy_indexes.rs`) keeps each query fast, but the sheer volume of round trips is wasted work — and during an initial backfill the entries table is empty anyway, so the fan-out query always returns zero rows.

**Fix:** skip per-record `emit_batch_resource_events` during backfill (receiver knows it is in `DeviceSyncState::Backfilling`) and emit one coarse \"content_identity invalidated\" event at the end of the model's backfill. Alternatively, make `emit_batch_resource_events` actually batch — do one bulk lookup of affected entries via `IN (...)` instead of a loop.

### 4. Sender per-record FK conversion is not batched

`core/src/infra/db/entities/content_identity.rs:264` (inside `query_for_sync`) runs:

```rust
for fk in Self::foreign_key_mappings() {
    crate::infra::sync::fk_mapper::convert_fk_to_uuid(&mut json, &fk, db).await?;
}
```

`convert_fk_to_uuid` (`core/src/infra/sync/fk_mapper.rs:65-102`) does one `mime_types` lookup per record. For a 5000-record batch that's 5000 extra round trips, despite `batch_lookup_uuids_for_local_ids` (`core/src/infra/sync/fk_mapper.rs:129-148`) already existing.

**Fix:** refactor `query_for_sync` to collect all FK ints, batch-lookup UUIDs, then attach to each JSON. Applies beyond `content_identity`.

## Suggested fix order (impact vs. effort)

1. **Add `LIMIT` to `PeerLog::get_since`** — tiny change, removes the dominant O(N^2) sender cost.
2. **Skip or coalesce `emit_batch_resource_events` during backfill** — removes ~2 × N redundant DB round trips on the receiver.
3. **Batch FK resolution in `query_for_sync`** — removes ~N round trips on the sender per batch, benefits all shared models.
4. **Page the initial-snapshot path** or drop the 100k cap in `get_current_shared_state` once the peer log path is efficient.

Happy to pick up (1) as a first PR.

## Environment

- Branch: `spacedrive-redundancy`
- Observed during initial pair + library sync, `content_identity` ≈ 327k records.

### LINKED ISSUE #3060: Entries (device-owned) sync backfill is slow: missing cursor index, per-record FK resolution, redundant per-entry closure rebuild
## Summary

Device-owned `entry` backfill (the `entries` table) is significantly slower than it should be, beyond what library size alone would explain. Unlike the shared-model issues in #3058, entries flow through the `StateRequest` → `get_device_state` → `query_for_sync` path, with a different set of hotspots. Three main causes: the sender cursor is not index-backed, the receiver deliberately processes entries one-at-a-time defeating batch FK resolution, and the closure table is rebuilt per-entry during backfill even though a bulk rebuild runs afterward.

### 1. Sender cursor has no supporting index (biggest sender-side cost)

`core/src/infra/db/entities/entry.rs:247-265` — `query_for_sync` paginates by `(indexed_at, uuid)`:

```rust
query = query
    .order_by_asc(Column::IndexedAt)
    .order_by_asc(Column::Uuid);
query = query.limit(batch_size as u64);
```

with a tie-breaker filter `(indexed_at > cursor_ts) OR (indexed_at = cursor_ts AND uuid > cursor_uuid)`.

The `indexed_at` column was added in `core/src/infra/db/migration/m20251021_000001_add_indexed_at_to_entries.rs` but **no index** was created on it. Existing entry indexes (see `core/src/infra/db/migration/m20240101_000001_initial_schema.rs:568-609`) cover `uuid`, `parent_id`, `kind`, and closure columns — none of them support this cursor.

Result: each batch request does a full-table scan ordered by `indexed_at` (or an unsorted scan + in-memory sort). Sender-side O(entries) per batch request, total O(N × batches) ≈ O(N²/batch_size).

**Fix:** add a composite index on `(indexed_at, uuid)` covering both ORDER BY and the cursor filter.

### 2. Receiver processes entries one-at-a-time, defeating batch FK resolution

`core/src/service/sync/backfill.rs:542-604` — because entries are hierarchical, the receiver explicitly opts out of batched FK resolution:

```rust
let processed_data = if model_type == \"entry\" && !fk_mappings.is_empty() {
    // Process entries individually: resolve FK → insert → resolve deps → next
    let mut succeeded = Vec::with_capacity(record_data.len());
    for data in record_data {
        let result = crate::infra::sync::batch_map_sync_json_to_local(
            vec![data.clone()],   // <-- single-element vec per call
            fk_mappings.clone(),
            &db,
        ).await?;
        ...
    }
}
```

The comment explains why (children need their just-inserted parents resolvable), but it means `batch_map_sync_json_to_local` is invoked N times with a 1-element vector. Each invocation still does one DB round trip per FK type. Entries have 5 FKs: `parent_id`, `content_id`, `metadata_id`, `volume_id`, `device_id` (see `core/src/infra/db/entities/entry.rs:99-110`). So the receiver runs up to ~5 × batch_size FK-lookup round trips per batch, where batched resolution could do ~5 total.

**Fix:** resolve non-self-referential FKs (`content_id`, `metadata_id`, `volume_id`, `device_id`) in one batch across all entries in the batch, then only resolve `parent_id` per-entry (or topo-sort the batch and resolve parent_id per depth tier). Most FK lookups are independent of insert order.

### 3. Redundant per-entry `entry_closure` rebuild during backfill

`core/src/infra/db/entities/entry.rs:567-660` — every applied entry calls `rebuild_entry_closure`, which does:

```rust
// Delete existing closure records (as descendant)
super::entry_closure::Entity::delete_many()
    .filter(super::entry_closure::Column::DescendantId.eq(entry_id))
    .exec(db).await?;

// Insert self-reference
self_closure.insert(db).await?;

// If parent exists, copy parent's ancestors
INSERT INTO entry_closure (ancestor_id, descendant_id, depth)
SELECT ancestor_id, ?, depth + 1
FROM entry_closure
WHERE descendant_id = ?
```

That is at minimum one DELETE + one INSERT + one INSERT-FROM-SELECT per entry during backfill. For a tree of average depth D, closure rows inserted per entry ≈ D+1. Over N entries, backfill writes O(N × D) closure rows.

**The kicker:** `Model::post_backfill_rebuild` (`core/src/infra/db/entities/entry.rs:167-169`) is registered on the Syncable trait and invoked by the backfill coordinator (`core/src/service/sync/backfill.rs:222-226`) via `run_post_backfill_rebuilds`. That function calls `rebuild_all_entry_closures` — a bulk rebuild of the entire closure table after backfill completes. So the per-entry closure work done during the hot loop is wasted: it's thrown away and redone in bulk.

**Fix:** detect backfill context in `Entry::apply_state_change` and skip `rebuild_entry_closure` during backfill (rely on `post_backfill_rebuild` instead). Only run the per-entry path for real-time sync ops. Could be done by threading a `&ApplyContext { is_backfill: bool }` parameter through `Syncable::apply_state_change`, or by having the backfill loop call a separate entry-specific apply path that omits the closure rebuild.

### 4. Per-entry FTS5 search-index trigger overhead (minor, but cumulative)

Migration `core/src/infra/db/migration/m20250120_000001_create_fts5_search_index.rs` installs AFTER INSERT/UPDATE/DELETE triggers on the `entries` table (filtered to `kind = 0` / files only, refined in `m20260114_000001_fix_search_index_include_directories.rs`). Each inserted entry fires an FTS5 write, per-row. Not O(N²), but adds a per-entry constant that would also benefit from deferral.

**Fix (optional):** for backfill, defer FTS5 indexing by dropping the triggers at backfill start and doing an `INSERT INTO fts ... SELECT` repopulation at the end. Same pattern as the closure rebuild.

### 5. Per-record FK conversion on sender (same as #3058 item 4)

`core/src/infra/db/entities/entry.rs:327-340` inside `query_for_sync` loops `convert_fk_to_uuid` per FK per record. Entries have 5 FKs so this is up to 5 × batch_size extra round trips per batch on the sender. The directory-path lookup at lines 275-293 is already batched — the FK conversions should follow the same pattern using `batch_lookup_uuids_for_local_ids` (`core/src/infra/sync/fk_mapper.rs:129-148`).

## Suggested fix order

1. **Add composite index on `entries(indexed_at, uuid)`** — one-migration change. Directly fixes the progressive sender-side slowdown.
2. **Skip per-entry `rebuild_entry_closure` during backfill** — large receiver-side win, leverages existing `post_backfill_rebuild` hook.
3. **Batch non-self FK resolution across the entry batch** on the receiver — moderate complexity, big constant-factor improvement.
4. **Batch FK conversion on the sender in `query_for_sync`** — mirrors the directory-path batching already in place.
5. **Defer FTS5 indexing during backfill** (optional, larger refactor).

Related: #3058 (shared models, e.g. `content_identity`).

## Environment

- Branch: `spacedrive-redundancy`
- Observed during initial pair + library sync with a large entries table.

jamiepine:
## Summary
- Fix progressive slowdown during initial library sync by removing O(N^2) patterns on sender and receiver.
- Addresses #3058 (shared models, `content_identity`) and #3060 (device-owned `entries`).
- Observed slowdown from ~5k records/batch down to ~100/batch past 100k applied is gone — batch size holds steady at the configured `backfill_batch_size`.

## Changes

**Sender**
- `PeerLog::get_since` now takes `limit: Option<usize>` and pushes `LIMIT ?` into SQL. Callers fetch `limit + 1` to derive `has_more`. Previously every `SharedChangeRequest` reloaded and parsed the entire remaining `shared_changes` log before truncating in memory.
- `Entry::query_for_sync` and `ContentIdentity::query_for_sync` batch FK→UUID conversion across the whole batch via a new `convert_fks_to_uuids_batch` helper — one DB round trip per FK type instead of per record × FK.

**Schema**
- New migration `m20260417_000001_add_entries_sync_cursor_index` adding `idx_entries_indexed_at_uuid` to back the cursor used by `Entry::query_for_sync`. Without it each batch request fell back to a full-table scan ordered by `indexed_at`.

**Receiver**
- New `infra::sync::backfill_context` module exposes a task-local `in_backfill` scope. The backfill coordinator wraps phases 2 + 3 in it; per-record hooks check `is_in_backfill()` to skip work that's redundant with the post-backfill pass.
- `Entry::apply_state_change` skips per-entry `entry_closure` rebuild during backfill. The existing `post_backfill_rebuild` pass already does a single bulk rebuild at the end.
- `ResourceManager::emit_batch_resource_events` short-circuits during backfill (avoids the per-UUID fan-out queries in `file.rs::route_from_dependency`). The backfill coordinator emits one `Event::Refresh` after the post-backfill rebuild so the frontend drops stale cached views.
- Entry FK resolution in `backfill.rs` splits mappings into self-referential (`parent_id`) and the rest. Non-self FKs resolve in one batch across the whole batch; only `parent_id` still runs per-entry so children can see just-inserted parents.

## Test plan
- [ ] Fresh pair + library sync with >100k `content_identity` records — batch throughput holds at configured `backfill_batch_size` end to end
- [ ] Fresh pair + library sync with >100k `entries` — same
- [ ] Incremental catch-up after offline period still updates `entry_closure` correctly (catch-up path is outside `in_backfill`)
- [ ] UI views populate after initial backfill completes (the coarse `Event::Refresh` fires and invalidates caches)
- [ ] Existing sync tests pass (`cargo test -p sd-core --lib sync`)

Closes #3058
Closes #3060

> [!NOTE]
> Performance fix for library sync backfill: eliminates O(N²) query patterns on both sender and receiver sides. Sender-side: implements database-level pagination in `PeerLog::get_since` and batches FK→UUID resolution. Receiver-side: introduces `backfill_context` module to skip redundant per-record work during bulk sync, defers `entry_closure` rebuilds and resource event emissions to post-backfill pass. Result: batch throughput remains constant even with 100k+ records instead of degrading to 100 records/batch.
> <sub>Written by [Tembo](https://app.tembo.io) for commit [17fdc0a](https://github.com/spacedriveapp/spacedrive/commit/17fdc0afc4e1f5977feb1f150bf8acfa9b908262).</sub>

### DIFF
--- a/core/src/domain/resource_manager.rs
+++ b/core/src/domain/resource_manager.rs
@@ -269,6 +269,19 @@ impl ResourceManager {
 		resource_type: &str,
 		resource_ids: Vec<Uuid>,
 	) -> Result<()> {
+		// During backfill we skip per-record fan-out. For models like
+		// content_identity each UUID triggers 2 DB queries via dependency
+		// routing, which dominates apply time on large libraries. The backfill
+		// coordinator emits a single coarse invalidation after the scope ends.
+		if crate::infra::sync::is_in_backfill() {
+			tracing::trace!(
+				resource_type = %resource_type,
+				count = resource_ids.len(),
+				"Skipping per-record resource event emission during backfill"
+			);
+			return Ok(());
+		}
+
 		// For now, delegate to single-resource handler
 		// In future, could optimize by batching virtual resource construction
 		self.emit_resource_events(resource_type, resource_ids).await
--- a/core/src/infra/db/entities/content_identity.rs
+++ b/core/src/infra/db/entities/content_identity.rs
@@ -246,13 +246,15 @@ impl Syncable for Model {
 
 		let results = query.all(db).await?;
 
-		let mut sync_results = Vec::new();
+		let mut sync_results: Vec<(Uuid, serde_json::Value, chrono::DateTime<chrono::Utc>)> =
+			Vec::with_capacity(results.len());
 		for content in results {
-			if content.uuid.is_none() {
-				continue;
-			}
+			let uuid = match content.uuid {
+				Some(u) => u,
+				None => continue,
+			};
 
-			let mut json = match content.to_sync_json() {
+			let json = match content.to_sync_json() {
 				Ok(j) => j,
 				Err(e) => {
 					tracing::warn!(error = %e, content_hash = %content.content_hash, "Failed to serialize content_identity for sync");
@@ -260,22 +262,39 @@ impl Syncable for Model {
 				}
 			};
 
-			// Convert FK to UUID for cross-device compatibility
-			for fk in Self::foreign_key_mappings() {
-				if let Err(e) =
-					crate::infra::sync::fk_mapper::convert_fk_to_uuid(&mut json, &fk, db).await
+			sync_results.push((uuid, json, content.last_verified_at));
+		}
+
+		// Batch FK → UUID conversion across the whole batch: one DB round trip
+		// per FK type instead of one per (record × FK).
+		let fk_mappings = Self::foreign_key_mappings();
+		if !fk_mappings.is_empty() && !sync_results.is_empty() {
+			let mut payloads: Vec<serde_json::Value> =
+				sync_results.iter().map(|(_, json, _)| json.clone()).collect();
+
+			for fk in &fk_mappings {
+				if let Err(e) = crate::infra::sync::fk_mapper::convert_fks_to_uuids_batch(
+					&mut payloads,
+					fk,
+					db,
+				)
+				.await
 				{
 					tracing::warn!(
 						error = %e,
-						uuid = %content.uuid.unwrap(),
 						fk_field = fk.local_field,
-						"Failed to convert FK to UUID, skipping content_identity"
+						"Batch FK conversion failed for content_identity"
 					);
-					continue;
+					return Err(sea_orm::DbErr::Custom(format!(
+						"ContentIdentity FK batch conversion failed: {}",
+						e
+					)));
 				}
 			}
 
-			sync_results.push((content.uuid.unwrap(), json, content.last_verified_at));
+			for ((_, json, _), resolved) in sync_results.iter_mut().zip(payloads.into_iter()) {
+				*json = resolved;
+			}
 		}
 
 		Ok(sync_results)
--- a/core/src/infra/db/entities/content_identity.rs
+++ b/core/src/infra/db/entities/content_identity.rs
@@ -271,8 +271,10 @@ impl Syncable for Model {
 		// payload.
 		let fk_mappings = Self::foreign_key_mappings();
 		if !fk_mappings.is_empty() && !sync_results.is_empty() {
-			let mut payloads: Vec<serde_json::Value> =
-				sync_results.iter().map(|(_, json, _)| json.clone()).collect();
+			let mut payloads: Vec<serde_json::Value> = sync_results
+				.iter()
+				.map(|(_, json, _)| json.clone())
+				.collect();
 			let mut failed_indices: std::collections::HashSet<usize> =
 				std::collections::HashSet::new();
 
--- a/core/src/infra/db/entities/content_identity.rs
+++ b/core/src/infra/db/entities/content_identity.rs
@@ -266,35 +266,55 @@ impl Syncable for Model {
 		}
 
 		// Batch FK → UUID conversion across the whole batch: one DB round trip
-		// per FK type instead of one per (record × FK).
+		// per FK type instead of one per (record × FK). Records that fail
+		// resolution are dropped so peers never see a sender-local int in the
+		// payload.
 		let fk_mappings = Self::foreign_key_mappings();
 		if !fk_mappings.is_empty() && !sync_results.is_empty() {
 			let mut payloads: Vec<serde_json::Value> =
 				sync_results.iter().map(|(_, json, _)| json.clone()).collect();
+			let mut failed_indices: std::collections::HashSet<usize> =
+				std::collections::HashSet::new();
 
 			for fk in &fk_mappings {
-				if let Err(e) = crate::infra::sync::fk_mapper::convert_fks_to_uuids_batch(
+				match crate::infra::sync::fk_mapper::convert_fks_to_uuids_batch(
 					&mut payloads,
 					fk,
 					db,
 				)
 				.await
 				{
-					tracing::warn!(
-						error = %e,
-						fk_field = fk.local_field,
-						"Batch FK conversion failed for content_identity"
-					);
-					return Err(sea_orm::DbErr::Custom(format!(
-						"ContentIdentity FK batch conversion failed: {}",
-						e
-					)));
+					Ok(failed) => failed_indices.extend(failed),
+					Err(e) => {
+						tracing::warn!(
+							error = %e,
+							fk_field = fk.local_field,
+							"Batch FK conversion failed for content_identity"
+						);
+						return Err(sea_orm::DbErr::Custom(format!(
+							"ContentIdentity FK batch conversion failed: {}",
+							e
+						)));
+					}
 				}
 			}
 
-			for ((_, json, _), resolved) in sync_results.iter_mut().zip(payloads.into_iter()) {
-				*json = resolved;
-			}
+			sync_results = sync_results
+				.into_iter()
+				.zip(payloads.into_iter())
+				.enumerate()
+				.filter_map(|(idx, ((uuid, _, ts), resolved))| {
+					if failed_indices.contains(&idx) {
+						tracing::warn!(
+							uuid = %uuid,
+							"Dropping content_identity with unresolved FK from sync batch"
+						);
+						None
+					} else {
+						Some((uuid, resolved, ts))
+					}
+				})
+				.collect();
 		}
 
 		Ok(sync_results)
--- a/core/src/infra/db/entities/entry.rs
+++ b/core/src/infra/db/entities/entry.rs
@@ -292,16 +292,16 @@ impl crate::infra::sync::Syncable for Model {
 				std::collections::HashMap::new()
 			};
 
-		// Convert to sync format with FK mapping
-		let mut sync_results = Vec::new();
+		// Serialize each row to JSON with its UUID and timestamp.
+		let mut staged: Vec<(Uuid, serde_json::Value, chrono::DateTime<chrono::Utc>)> =
+			Vec::with_capacity(results.len());
 
 		for entry in results {
 			let uuid = match entry.uuid {
 				Some(u) => u,
-				None => continue, // Skip entries without UUIDs
+				None => continue,
 			};
 
-			// Serialize to JSON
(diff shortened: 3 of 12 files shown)


jamiepine:
Pushed `bd1c7130f` addressing two of the four new CodeRabbit comments:

- **`fk_mapper.rs` absent-field handling** — now treats `None` as failed, matching `convert_fk_to_uuid`'s contract (only explicit JSON `null` is a legitimate null FK).
- **`entry.rs` cursor alignment** — no longer falls back to `modified_at` when `indexed_at` is `NULL`. The query filters/orders on `indexed_at` only, so the returned cursor would otherwise misrepresent the scan position. The backfill migration populated existing rows; any `NULL` here is a data bug and is now logged + skipped.

Not addressing the two "SQL-limited page shrinkage" comments in this PR:

The concern is legitimate (tail-of-batch FK failures can stall the cursor) but the fix is out of scope here:

1. The pre-existing per-record `convert_fk_to_uuid` call sites had the **same** failure mode silently — they warned and `continue`d the inner FK loop, then pushed the entry to `sync_results` anyway with the sender-local int intact. That silent FK-corruption path is what the earlier review already caught and what this PR fixes.
2. Proper resolution needs over-fetching in `query_for_sync` (or a parallel "max scanned cursor" return value), which is a bigger refactor than belongs in a perf PR.
3. In practice this triggers only when a record references a FK target that's missing from the sender's own DB — a data corruption bug, not normal operation.

Filed as a follow-up rather than blocking this PR. Open to doing the over-fetch refactor as a separate change if reviewers prefer.

